PR #4, opened 2023-04-03
Add documentation for rotation

mkorbel1:
## Description & Motivation

Adds documentation for rotation, linked from the main README.

## Related Issue(s)

N/A

## Testing

Adjusted `check_documentation.sh` to allow for relative links.  This *does* break links in the generated README inside the generated `dart doc` output, but I couldn't find a way to configure that to fix it.

## Backwards-compatibility

No

## Documentation

That's all it is


doc/rotate.md
```diff
@@ -0,0 +1,40 @@
...
+
+If you want to rotate a signal by a fixed amount, you can use the "fixed" rotation modules: [`RotateLeftFixed`](https://intel.github.io/rohd-hcl/rohd_hcl/RotateLeftFixed-class.html) and [`RotateRightFixed`](https://intel.github.io/rohd-hcl/rohd_hcl/RotateRightFixed-class.html).  These will be implemented as a simple signal swizzle (no added gates or logic).
+
+If you want to rotate by a dynamic amount controlled by another signal, you can use the non-fixed rotation modules: [`RotateLeft`](https://intel.github.io/rohd-hcl/rohd_hcl/RotateLeft-class.html) and [`RotateRight`](https://intel.github.io/rohd-hcl/rohd_hcl/RotateRight-class.html).  These are implemented as a case statement with a static swizzle per matching rotate amount.  Use the `maxAmount` argument to limit the amount of hardware generated in case you know the `rotateAmount` will never be greater than a certain value.
+
+Instantiating these synthesizable modules looks similar for all four.  An example is shown below creating a dynamic rotate left module that rotates a 16-bit signal by an 8-bit value which is guaranteed to never be larger than 10.
```
desmonddak, line 17: all four rotation _combinations_.

doc/rotate.md
```diff
@@ -0,0 +1,40 @@
...
+
+```dart
+final original = Logic(width: 16);
+final rotateAmount = Logic(width: 8);
+final mod = RotateLeft(original, rotateAmount, maxAmount: 10);
+final rotated = mod.rotated;
+```
+
+A set of convenient `extension`s are provided on top of `Logic` to more easily rotate signals with a simpler API: [`rotateLeft`](https://intel.github.io/rohd-hcl/rohd_hcl/RotateLogic/rotateLeft.html) and [`rotateRight`](https://intel.github.io/rohd-hcl/rohd_hcl/RotateLogic/rotateRight.html).  These methods construct the same underlying modules, but based on the type of `rotateAmount` (`int` vs `Logic`) will decide whether to use a "fixed" or "non-fixed" implementation.
```
desmonddak, line 26: Should you describe these first/only?  Is there value to exposing the direct APIs?
It seems to raise abstraction to simply say 'if you pass in a constant the rotator will synthesize a swizzle, rather than describe 4 variants.  Just two variants and say the code will optimize for you.  Otherwise there's a lot of API description for something that be brutally simple.

Perhaps the API description should be that simple:  we support a rotate (left/right).  It does the right thing for you.
If you provide a maximum amount, it will synthesize with fewer shifters.  If you specify a constant, it will just swizzle.  No need to describe how it is implemented (e.g. the 'extension' business)

mkorbel1: I agree that the simpler APIs should be the focus and come first.  Sometimes people may want to instantiate actual modules (for converting to SV, or as a list of `Module`s, etc.) but you're right that that is less common.  But I'll keep the other ones there, just moved to a less prominent spot.

desmonddak (approved):

Issue #3, opened 2023-04-03
Use `Logic.or()` instead of custom `OrReduction` in one-hot


mkorbel1:
### Describe the bug

There's already an `or` reduction via `Logic.or()` that we should use instead of the `OrReduction` module.

PR #5, opened 2023-04-03
use built-in or-reduction

desmonddak:
## Description & Motivation

Cleanup implementation and test of one-hot codec

## Related Issue(s)

Resolves issue #3

## Testing

Removed OrReduction class and tests.  Replaced with .or() bitlevel function.

## Backwards-compatibility

Nothing should be using the OrReduction class.

## Documentation

No update


mkorbel1 (approved):
Looks good!

PR #21, opened 2023-05-05
Sort

### LINKED ISSUE #9: Sort
### Motivation

Sorting is a useful component for the library

### Desired solution

A simple first version of sorting to include in the library

### Alternatives considered

_No response_

### Additional details

_No response_

quekyj:
## Description & Motivation

A bitonic sorting module that can sort a List of `Logic`.

## Related Issue(s)

Fix #9 

## Testing

Tested with ascending and descending order test.

## Backwards-compatibility

No

## Documentation

Added documentation to the constructor and public variable.


lib/src/bitonic_sort.dart
```diff
@@ -0,0 +1,181 @@
...
+      }
+    } else if (_inputs.length == 1) {
+      _outputsFinal.add(addOutput('sorted_0', width: _inputs[0].width));
+      _outputsFinal[0] <= _inputs[0];
+    }
+  }
+}
+
+/// Sort [inputs] to specified order.
+class BitonicSort extends Module {
```
mkorbel1, line 135: Can we create an `abstract class Sort` that this extends from so that different `Sort` classes could be swapped out more easily?

lib/src/bitonic_sort.dart
```diff
@@ -0,0 +1,181 @@
...
+
+  /// The [sorted] result.
+  List<Logic> get sorted => _outputs;
+
+  /// Constructs a [Module] to sort list of [Logic] based on [direction].
+  ///
+  /// The sorting module will recursively split inputs into a bitonic sequence
+  /// and passed to [_BitonicMerge] to perform merging process and return
+  /// the final [sorted] results.
+  BitonicSort(Logic clk, Logic reset, List<Logic> inputs, int direction,
```
mkorbel1, line 150: `int direction` is confusing, what value are you supposed to put?  It would be better to make this a `bool ascending` or perhaps an `enum SortDirection {ascending, descending}`?

lib/src/bitonic_sort.dart
```diff
@@ -0,0 +1,181 @@
...
+        _outputsFinal[i] <= mergeRes[i];
+      }
+    } else if (_inputs.length == 1) {
+      _outputsFinal.add(addOutput('sorted_0', width: _inputs[0].width));
+      _outputsFinal[0] <= _inputs[0];
+    }
+  }
+}
+
+/// Sort [inputs] to specified order.
```
mkorbel1, line 134: It would be good to document the latency of the sorter (as a function of number of items sorted, probably), and also clarify that it is pipelined rather than multi-cycle.

lib/src/bitonic_sort.dart
mkorbel1, line 1: Please write a document in the `doc/` directory about how `Sort` and `BitonicSorc` works, some examples of how to use it, etc. (look at other docs for examples).

lib/src/bitonic_sort.dart
```diff
@@ -0,0 +1,181 @@
...
+  List<Logic> get sorted => _outputs;
+
+  /// Constructs a [Module] to sort list of [Logic] based on [direction].
+  ///
+  /// The sorting module will recursively split inputs into a bitonic sequence
+  /// and passed to [_BitonicMerge] to perform merging process and return
+  /// the final [sorted] results.
+  BitonicSort(Logic clk, Logic reset, List<Logic> inputs, int direction,
+      {super.name}) {
+    clk = addInput('clk', clk);
```
mkorbel1, line 152: Add a check that all `inputs` are the same width (assuming that's something you want to do)

lib/src/bitonic_sort.dart
mkorbel1, line 150: Does this support non-power-of-2 numbers of elements?  If not, should we support it?  Or else it should be explicitly documented and also add a check to make sure the rule is followed

lib/src/bitonic_sort.dart
```diff
@@ -0,0 +1,181 @@
...
+/// Sort [inputs] to specified order.
+class BitonicSort extends Module {
+  /// The list of inputs port.
+  final List<Logic> _inputs = [];
+
+  /// The list of outputs port.
+  final List<Logic> _outputs = [];
+
+  /// The [sorted] result.
+  List<Logic> get sorted => _outputs;
```
mkorbel1, line 143: We can change this to `late final List<Logic> sorted = UnmodifiableListView(_outputs);` to lazily initialize a view that doesn't allow for modification of the constructed list.

lib/src/bitonic_sort.dart
mkorbel1, line 150: Is `toSort` better than generic `inputs`?  Especially since `inputs` masks the built-in `Module.inputs`.

lib/src/bitonic_sort.dart
```diff
@@ -0,0 +1,181 @@
...
+  /// The sorting module will recursively split inputs into a bitonic sequence
+  /// and passed to [_BitonicMerge] to perform merging process and return
+  /// the final [sorted] results.
+  BitonicSort(Logic clk, Logic reset, List<Logic> inputs, int direction,
+      {super.name}) {
+    clk = addInput('clk', clk);
+    reset = addInput('reset', reset);
+
+    for (var i = 0; i < inputs.length; i++) {
+      _inputs.add(addInput('x$i', inputs[i], width: inputs[i].width));
```
mkorbel1, line 156: Let's name the ports generated the same thing (`toSort`?) instead of `x`

lib/src/bitonic_sort.dart
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    if (_inputs.length > 1) {
+      final k = _inputs.length ~/ 2;
```
mkorbel1, line 160: Do we really need this `k` variable?  The name is not very informative, and it's only used to name the `sortLeft` and `sortRight`, but can't we just call them `sort_left` and `sort_right` without the `_$k`?

lib/src/bitonic_sort.dart
```diff
@@ -0,0 +1,181 @@
...
+      final sortLeft = BitonicSort(
+          clk, reset, _inputs.getRange(0, _inputs.length ~/ 2).toList(), 1,
```
mkorbel1, line 162: Maybe let's change the `Sort` argument to `Iterable<Logic>` instead of `List<Logic>` so that we don't have to call `toList()`?

lib/src/bitonic_sort.dart
```diff
@@ -0,0 +1,181 @@
...
+          name: 'sort_left_$k');
+
+      final sortRight = BitonicSort(clk, reset,
+          _inputs.getRange(_inputs.length ~/ 2, _inputs.length).toList(), 0,
+          name: 'sort_right_$k');
+
+      final res = sortLeft.sorted + sortRight.sorted;
+
+      final y = _BitonicMerge(clk, reset, res, direction);
```
mkorbel1, line 171: Why not make this something like
```dart
final mergeResult = _BitonicMerge(clk, reset, res, direction).sorted;
```
Then we don't have this poorly-named `y` variable and also don't have to keep calling `.sorted` below.

lib/src/bitonic_sort.dart
```diff
@@ -0,0 +1,181 @@
...
+  List<Logic> get swapped => _outputs;
+
+  /// Compare and Swap the order of [i] and [j] in [inputs] based on the
+  /// direction given.
+  ///
+  /// The position [i] and [j] will swapped if [direction] is 1 and [_inputs]
+  /// of [i] greater than [_inputs] of [j] or [direction] is 0 and [_inputs]
+  /// of [i] lower than [_inputs] of [j].
+  _CompareSwap(
+      Logic clk, Logic reset, List<Logic> inputs, int i, int j, int direction)
```
mkorbel1, line 30: Same change on `direction` here as below as an argument

lib/src/bitonic_sort.dart
```diff
@@ -0,0 +1,181 @@
...
+
+/// Compare and Swap [Logic] to the specified order.
+class _CompareSwap extends Module {
+  final List<Logic> _inputs = [];
+
+  /// The list of sorted [Logic] result.
+  final List<Logic> _outputs = [];
+
+  /// The [swapped] list of result.
+  List<Logic> get swapped => _outputs;
```
mkorbel1, line 21: Same as below, use `UnmodifiableListView`

lib/src/bitonic_sort.dart
```diff
@@ -0,0 +1,181 @@
...
+  /// direction given.
+  ///
+  /// The position [i] and [j] will swapped if [direction] is 1 and [_inputs]
+  /// of [i] greater than [_inputs] of [j] or [direction] is 0 and [_inputs]
+  /// of [i] lower than [_inputs] of [j].
+  _CompareSwap(
+      Logic clk, Logic reset, List<Logic> inputs, int i, int j, int direction)
+      : super(name: 'compare_swap_${i}_$j') {
+    clk = addInput('clk', clk);
+    reset = addInput(reset.name, reset);
```
mkorbel1, line 33: why not `'reset'` instead of `reset.name`?

lib/src/bitonic_sort.dart
mkorbel1, line 30: Same change on name of `inputs` as below

lib/src/bitonic_sort.dart
```diff
@@ -0,0 +1,181 @@
...
+
+    for (var i = 0; i < inputs.length; i++) {
+      _inputs.add(addInput('x$i', inputs[i], width: inputs[i].width));
```
mkorbel1, line 36: Same change on name of ports as below

lib/src/bitonic_sort.dart
```diff
@@ -0,0 +1,181 @@
...
+            newValA < _inputs[i],
+            newValB < _inputs[j]
+          ]),
+    ]);
+
+    _inputs[i] = newValA;
+    _inputs[j] = newValB;
+
+    for (var k = 0; k < _inputs.length; k++) {
+      _outputs.add(addOutput('y$k', width: _inputs[k].width));
```
mkorbel1, line 61: Let's name the output port something nicer than `y`, like maybe `swapped`?

lib/src/bitonic_sort.dart
```diff
@@ -0,0 +1,181 @@
...
+      _outputs[k] <= _inputs[k];
+    }
+  }
+}
+
+class _BitonicMerge extends Module {
```
mkorbel1, line 67: good to have a doc comment even on private classes

lib/src/bitonic_sort.dart
```diff
@@ -0,0 +1,181 @@
...
+  List<Logic> get sorted => _outputsFinal;
+
+  /// Merge and sort [bitonicSequence] into [direction] given.
+  ///
+  /// List of [Logic] will compare and swap [Logic] position based on the
+  /// [direction] given by [BitonicSort] to first created a bitonic sequence.
+  /// The final stage will sort the bitonic sequence into sorted order
+  /// of [direction].
+  _BitonicMerge(
+      Logic clk, Logic reset, List<Logic> bitonicSequence, int direction)
```
mkorbel1, line 87: Again `direction` type

lib/src/bitonic_sort.dart
```diff
@@ -0,0 +1,181 @@
...
+      : super(name: 'bitonic_merge') {
+    clk = addInput('clk', clk);
+    reset = addInput('reset', reset);
+
+    for (var i = 0; i < bitonicSequence.length; i++) {
+      _inputs.add(
+          addInput('x$i', bitonicSequence[i], width: bitonicSequence[i].width));
```
mkorbel1, line 94: Again name of port

lib/src/bitonic_sort.dart
```diff
@@ -0,0 +1,181 @@
...
+  List<Logic> _inputs = [];
+
+  /// A list of [Logic] that hold the results after [_CompareSwap].
+  final List<Logic> _outputs = [];
+
+  /// A list of [Logic] that hold the final outputs of List of result.
+  final List<Logic> _outputsFinal = [];
+
+  /// The [sorted] result.
+  List<Logic> get sorted => _outputsFinal;
```
mkorbel1, line 78: Again, use `UnmodifiableListView`

lib/src/bitonic_sort.dart
```diff
@@ -0,0 +1,181 @@
...
+
+      for (var i = 0; i < 0 + k; i++) {
+        final indexA = i;
+        final indexB = i + k;
+        final swap =
+            _CompareSwap(clk, reset, _inputs, indexA, indexB, direction);
+        _inputs = swap.swapped;
+      }
+
+      for (var i = 0; i < _inputs.length; i++) {
```
mkorbel1, line 108: do you need these `_outputs` at all? can we just remove them?

lib/src/bitonic_sort.dart
```diff
@@ -0,0 +1,181 @@
...
+        _outputs.add(addOutput('y$i', width: _inputs[i].width));
+        _outputs[i] <= _inputs[i];
+      }
+
+      final mergeLeft = _BitonicMerge(clk, reset,
```
mkorbel1, line 113: name these modules `merge_left` and `merge_right`?

test/bitonic_sort_test.dart
```diff
@@ -0,0 +1,113 @@
...
+      reset.inject(0);
+
+      Simulator.registerAction(100, () {
+        for (var i = 0; i < topMod.sorted.length; i++) {
+          expect(topMod.sorted[i].value.toInt(), i + 1);
+        }
+      });
+
+      Simulator.setMaxSimTime(100);
+      // WaveDumper(topMod, outputPath: 'lib/src/sort/recursive_list.vcd');
```
mkorbel1, line 63: delete commented line of code

test/bitonic_sort_test.dart
```diff
@@ -0,0 +1,113 @@
...
+          expect(topMod.sorted[i].value.toInt(), topMod.sorted.length - i);
+        }
+      });
+
+      Simulator.setMaxSimTime(100);
+      // WaveDumper(topMod, outputPath: 'lib/src/sort/recursive_list.vcd');
+
+      await Simulator.run();
+    });
+  });
```
mkorbel1, line 112: Let's add some more interesting tests than just these two cases where they are pre-sorted in the wrong direction?  Can we randomly shuffle (fixed seed) an array a few times and make sure the sort works?

mkorbel1, line 112: If we support it, add a test for non-power-of-two numbers of items

mkorbel1, line 112: Try adding negative numbers as well and see if it works?  (you might find a bug in ROHD, i think testing is not great for negative numbers in ROHD, but we should try it and file a bug if there's an issue)

mkorbel1, line 112: Add a test where we push different arrays in every cycle, and check that the output changes the expected number of cycles later (basically demonstrate that this can be used as a pipelined sorter).

test/bitonic_sort_test.dart
```diff
@@ -0,0 +1,113 @@
...
+}
```
mkorbel1, line 113: Add tests for exceptions and edge cases:
- what if an empty list is provided?
- what if 1 element is provided?
- what if elements are not all the same width?
- what if multiple elements have the same value?
- anything else?

mkorbel1 (changes requested):
Great work!  This is an awesome addition to the library.

lib/src/sort.dart
```diff
@@ -0,0 +1,213 @@
...
+  /// The list of outputs port.
+  final List<Logic> _outputs = [];
+
+  /// The [sorted] result.
+  List<Logic> get sorted => UnmodifiableListView(_outputs);
+
+  /// Constructs a [Module] to sort list of [Logic].
+  ///
+  /// The sorting module will recursively split inputs into a bitonic sequence
+  /// and passed to [_BitonicMerge] to perform merging process and return
```
mkorbel1, line 164: Probably shouldn't reference private classes in a public API.  The interesting thing is that it uses the bitonic sort algorithm.

lib/src/sort.dart
```diff
@@ -0,0 +1,213 @@
...
+      final bitonicSequence = sortLeft.sorted + sortRight.sorted;
+
+      final mergeResult = _BitonicMerge(clk, reset,
+              bitonicSequence: bitonicSequence, isAscending: isAscending)
+          .sorted;
+      for (var i = 0; i < mergeResult.length; i++) {
+        _outputs.add(addOutput('sorted_$i', width: _inputs[i].width));
+        _outputs[i] <= mergeResult[i];
+      }
+    } else {
```
mkorbel1, line 208: This will break if the number of elements is 0.  Returning an empty list for `_outputs` is fine if the input list is empty, since that's sorted and removes an edge case for users to consider.

quekyj: The number of elements can't be 0 as the inputs will validate if the inputs is power of 2.

test/bitonic_sort_test.dart
quekyj: Added test for `2^0` = 1. Single input given to the sort module.

test/bitonic_sort_test.dart
```diff
@@ -0,0 +1,518 @@
...
+        //   }
+        // });
+
+        Simulator.setMaxSimTime(100);
+
+        await Simulator.run();
+      });
+    });
+
+    group('Descending Order: ', () {
```
mkorbel1, line 328: perhaps you can parameterize tests to support both ascending and descending to avoid code duplication

mkorbel1 (changes requested):

lib/src/sort.dart
```diff
@@ -0,0 +1,250 @@
...
+    for (var k = 0; k < _inputs.length; k++) {
+      _outputs.add(addOutput('swapped$k', width: _inputs[k].width));
+      _outputs[k] <= _inputs[k];
+    }
+  }
+}
+
+class _BitonicMerge extends Module {
+  /// A list of [Logic] that hold inputs.
+  List<Logic> _inputs = [];
```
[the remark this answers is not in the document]
quekyj: ```_inputs = swap. Swapped;```

The inputs will be reassign to the swapped value later

test/bitonic_sort_test.dart
```diff
@@ -0,0 +1,483 @@
...
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+import 'package:test/test.dart';
+
+Future<void> main() async {
+  tearDown(() async {
+    await Simulator.reset();
+  });
+
+  List<int> generateNonDuplicateRand(int seed, int maxLength, int listLength) {
```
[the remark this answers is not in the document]
quekyj: I think can go with `int` first since the `List` provides `.sort()` function which sorts the generated numbers. I am not sure do we have sort function that can sort `LogicValue` to ascending order.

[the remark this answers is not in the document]
quekyj:
```dart
  type '_SmallLogicValue' is not a subtype of type 'Comparable<dynamic>' in type cast
  dart:collection                     ListBase.sort
  test/bitonic_sort_test.dart 193:19  main.<fn>.<fn>.<fn>
```

Is this an issue in ROHD? I get this error when trying to use `.sort()` on List of `LogicValue`. Probably want to disable calling `.sort()` or implement this function.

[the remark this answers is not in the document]
mkorbel1: I'm not sure if it's a bug, but that's definitely a missing feature.  We should implement it though!  Filed here: https://github.com/intel/rohd/issues/371

lib/src/sort.dart
```diff
@@ -0,0 +1,261 @@
...
+      _inputs.add(addInput('toSort$i', toSort[i], width: toSort[i].width));
+    }
+
+    final ascending = isAscending == true ? Const(1) : Const(0);
+    final newValA = Logic(width: toSort[0].width);
+    final newValB = Logic(width: toSort[0].width);
+
+    Sequential(clk, [
+      If(
+          (ascending & (_inputs[i] > _inputs[j])) |
```
mkorbel1, line 69: we can know `isAscending` at generation time, so we should just use a dart ternary operator here instead of using logic with `Const`s.

lib/src/sort.dart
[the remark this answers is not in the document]
mkorbel1: oh i see, we should just declare this list inside the constructor then, since it's a local variable rather than a property of the object

test/bitonic_sort_test.dart
```diff
@@ -0,0 +1,479 @@
...
+      final clk = SimpleClockGenerator(10).clk;
+      final reset = Logic(name: 'reset');
+
+      final toSort = <Logic>[];
+
+      expect(() async {
+        BitonicSort(clk, reset, toSort: toSort, name: 'top_level');
+      }, throwsA((dynamic e) => e is RohdHclException));
+    });
+    test('should return itself if single element is given.', () async {
```
mkorbel1, line 64: let's add some spacing (one blank line) between `test`s so it looks nicer

test/bitonic_sort_test.dart
```diff
@@ -0,0 +1,479 @@
...
+          await waitCycles(topMod.latency).then((value) {
+            for (var i = 0; i < topMod.sorted.length; i++) {
+              expect(topMod.sorted[i].value.toInt(), input[i]);
+            }
+          });
+
+          await clk.nextNegedge;
+        }
+
+        await Simulator.simulationEnded;
```
mkorbel1, line 286: similar feedback as in the multiplier, end the simulation explicitly, don't `await` the `waitCycles`, ensure the `expect` ran the right number of times

mkorbel1 (changes requested):

lib/src/sort.dart
mkorbel1: I meant something like this:
```dart
      If( ascending ? (_inputs[i] > _inputs[j]) : _inputs[i].lt(_inputs[j])
```
Note how with this approach the generated hardware doesn't even consider whether it's ascending or descending, it just gets generated one way or the other.

test/bitonic_sort_test.dart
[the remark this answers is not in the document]
quekyj: I will mark this resolved first temporary since the test is not critical.

PR #23, opened 2023-05-10
schematic generation

desmonddak:
## Description & Motivation

A first attempt at schematic generation

## Related Issue(s)

None

## Testing

Local running of github actions and loading of resultant HTML.

## Backwards-compatibility

No.

## Documentation

This will generate HTML schematics for components. These will need to be linked into the existing documentation.
The way that the static-assets are copied over may need to be changed or moved. Finally, we are using dart test to do the
generation which is overkill, we should generate the Verilog more directly.


tool/gh_actions/install_synthesis.sh
```diff
@@ -0,0 +1,36 @@
...
+	       python3 \
+	       python3-dev \
+	       python3-pip \
+	       python3-setuptools \
+	       python3-wheel \
+	       python3-tk
+
+sudo git clone https://github.com/YosysHQ/yosys.git
+cd yosys
+sudo make -j$(nproc)
```
mkorbel1, line 36: Perhaps we can use one of the pre-built yosys binaries to speed this up:
https://yosyshq.net/yosys/download.html

test/rf_test.dart
```diff
@@ -159,4 +160,30 @@ void main() {
...
+      for (var i = 0; i < numRd; i++)
+        DataPortInterface(dataWidth, addrWidth)..en.put(0)
+    ];
+
+    final rf = RegisterFile(clk, reset, wrPorts, rdPorts, numEntries: 20);
+
+    await rf.build();
+
+    final res = rf.generateSynth();
+    File('build/${rf.definitionName}.v').openWrite().write(res);
```
mkorbel1, line 187: Use `File.writeAsStringSync` so that it doesn't run asynchronously on a stream.  I ran this as an experiment and only some files got generated because of a race between the file write and the tests completing.

mkorbel1 (changes requested):

desmonddak:
I have finalized these changes in a fork, so this pull request can be dropped.

test/arbiter_test.dart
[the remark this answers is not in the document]
desmonddak: I separated this into a gen/ directory.

tool/gh_actions/install_synthesis.sh
desmonddak: This is now done, I pull down the OSS CAD suite.

PR #25, opened 2023-05-12
Component updates in README

mkorbel1:
## Description & Motivation

Add some more component ideas to the README and update links and wording for some others.

## Related Issue(s)

N/A

## Testing

N/A

## Backwards-compatibility

No

## Documentation

N/A


desmonddak (commented):
Looks good. We'll think about a different presentation of components to replace a two-level list.

PR #28, opened 2023-05-17
Carry save multiplier

### LINKED ISSUE #10: Pipelined integer multiply
### Motivation

Integer multiplication in a pipeline is a nice component for the library

### Desired solution

A simple pipelined integer multiplier

### Alternatives considered

_No response_

### Additional details

_No response_

quekyj:
## Description & Motivation

Add carry save multiplier module.

## Related Issue(s)

Fix #10 

## Testing

Added unit test in the test folder

## Backwards-compatibility

No

## Documentation

Yes


lib/src/carry_save_multiplier.dart
```diff
@@ -0,0 +1,186 @@
...
+/// A datatype class that hold result from [FullAdder].
+class FullAdderResult {
+  /// The output sum from full adder.
+  final sum = Logic(name: 'sum');
+
+  /// The output carry-out from full adder.
+  final cOut = Logic(name: 'c_out');
+}
+
+/// A [FullAdder] module that can perform simple adder. The full adder output
```
mkorbel1, line 22: Doc comment is a little confusing.  Let's describe it more specifically, something more like
```dart
/// A simple full-adder with inputs [a] and [b] to be added with a [carryIn].
```

lib/src/carry_save_multiplier.dart
```diff
@@ -0,0 +1,186 @@
...
+/// can be defined in the truth table.
+class FullAdder extends Module {
```
mkorbel1, line 24: A full adder is a good component all by itself, we should include it in a separate file?

lib/src/carry_save_multiplier.dart
```diff
@@ -0,0 +1,186 @@
...
+// Implementation of pipeline multiplier module.
+//
+// 2023 February 21
+// Author: Max Korbel <max.korbel@intel.com>
+//
+
+import 'package:rohd/rohd.dart';
+
+/// A datatype class that hold result from [FullAdder].
+class FullAdderResult {
```
mkorbel1, line 14: Let's ditch this `FullAdderResult` object and just have `sum` and `cOut` directly as outputs (`get`ters) of the `FullAdder` module.

lib/src/carry_save_multiplier.dart
```diff
@@ -0,0 +1,186 @@
...
+    b = addInput('b', b, width: b.width);
+    carryIn = addInput('carry_in', carryIn, width: carryIn.width);
+
+    final carryOut = addOutput('carry_out');
+    final sum = addOutput('sum');
+
+    final and1 = carryIn & (a ^ b);
+    final and2 = b & a;
+
+    Combinational([
```
mkorbel1, line 48: Use assigns `<=` isntead of a ` Combinational` to make things cleaner.

lib/src/carry_save_multiplier.dart
```diff
@@ -0,0 +1,186 @@
...
+      carryOut < and1 | and2,
+    ]);
+
+    fullAdderResult.sum <= output('sum');
+    fullAdderResult.cOut <= output('carry_out');
+  }
+}
+
+/// An [NBitAdder] that perform addition.
+class NBitAdder extends Module {
```
mkorbel1, line 59: This is a good component by itself, put it in another file?

mkorbel1 (changes requested):

lib/src/carry_save_multiplier.dart
```diff
@@ -4,29 +4,19 @@
...
 class FullAdder extends Module {
-  /// The variable that hold [FullAdder] results.
-  final fullAdderResult = FullAdderResult();
+  /// The result [sum] from [FullAdder].
+  Logic get sum => output('sum');
 
-  /// The results returned from [FullAdder].
-  FullAdderResult get fullAdderRes => fullAdderResult;
+  /// The result [cOut] from [FullAdder].
+  Logic get cOut => output('carry_out');
```
mkorbel1, line 19: Either call this `carryOut` or call the input `cIn`, should be consistent.

lib/src/carry_save_multiplier.dart
```diff
@@ -4,29 +4,19 @@
...
+/// A simple full-adder with inputs `a` and `b` to be added with a `carryIn`.
 class FullAdder extends Module {
-  /// The variable that hold [FullAdder] results.
-  final fullAdderResult = FullAdderResult();
+  /// The result [sum] from [FullAdder].
+  Logic get sum => output('sum');
 
-  /// The results returned from [FullAdder].
-  FullAdderResult get fullAdderRes => fullAdderResult;
+  /// The result [cOut] from [FullAdder].
```
mkorbel1, line 18: Clarify what this is, it's the carry bit.  Calling it `[cOut]` doesnt' help in the docs.

mkorbel1 (changes requested):

doc/multiplier.md
```diff
@@ -0,0 +1,24 @@
...
+
+ROHD HCL provides a multiplier module to get the product from a list of logic. As of now, we have
+
+- [Carry Save Multiplier](#carry-save-multiplier)
+
+## Carry Save Multiplier
+
+Carry save multiplier is a digital circuit used for performing multiplication operations. It is particularly useful in applications that require high speed multip0lication, such as digital signal processing.
+
+The [`CarrySaveMultiplier`](https://intel.github.io/rohd-hcl/rohd_hcl/CarrySaveMultiplier-class.html) module in ROHD-HCl accept four inputs: the clock `clk` signal, reset `reset` signal, a List of Logic()'s `toMultiply` and the name of the module `name`. Note that `toMultiply` must have inputs of two and width of the inputs must be the same or `RohdHclException()` will be thrown.
```
mkorbel1, line 11: Typo "ROHD-HCl" should be "ROHD-HCL"

mkorbel1, line 11: Instead of "Logic()", say `Logic` (in ticks).  This applies in the adder doc also.  There's no need to add `()` after class names (e.g. in `RohdHclException()`)

lib/src/adder.dart
```diff
@@ -0,0 +1,55 @@
+// Copyright (C) 2023 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// carry_save_multiplier.dart
```
mkorbel1, line 4: Check the names of files and file descriptions, copy-paste error in multiple files.

lib/src/ripple_carry_adder.dart
```diff
@@ -0,0 +1,53 @@
...
+    final b = addInput('b', toSum[1], width: toSum[1].width);
+    carry = addInput('carry_in', carry, width: carry.width);
+
+    if (a.width != b.width) {
+      throw RohdHclException('inputs in toSum should have same width.');
+    }
+
+    for (var i = 0; i < a.width; i++) {
+      FullAdder? fullAdder;
+      fullAdder = FullAdder(a: a[i], b: b[i], carryIn: carry);
```
mkorbel1, line 45: Why not just `final fullAdder = ...`?

lib/src/adder.dart
```diff
@@ -0,0 +1,55 @@
...
+
+import 'package:rohd/rohd.dart';
+
+/// An abstract class for all adder module.
+abstract class Adder extends Module {
+  /// The List of inputs to sum. Length of inputs must be two.
+  List<Logic> toSum;
+
+  /// The addition results [sum].
+  List<Logic> get sum;
```
mkorbel1, line 19: The output of an adder should be a `Logic`, not a `List<Logic>`.  Swizzle it in implementations.

lib/src/carry_save_mutiplier.dart
```diff
@@ -0,0 +1,126 @@
...
+/// A binary multiplier using carry-save technique with pipelineing.
+///
+/// Reduces calculation time and complexity by employing the carry save
+/// multiplier, which splits numbers into smaller components and performs
+/// partial multiplications on each component separately, storing the results
+/// in a compact form. The pipeline enhances performance by breaking down the
+/// multiplication process into sequential stages, allowing for concurrent
+/// execution of multiple operations.
+///
+/// The latency of the carry save multiplier is the sum of the two inputs width
```
mkorbel1, line 23: Can you add a getter that calculates the latency of the multiplier, also?

lib/src/ripple_carry_adder.dart
```diff
@@ -0,0 +1,53 @@
...
+  RippleCarryAdder({required super.toSum, super.name = 'ripple_carry_adder'}) {
+    Logic carry = Const(0);
+
+    if (toSum.length != 2) {
+      throw RohdHclException('Length of toSum must be two.');
+    }
+
+    final a = addInput('a', toSum[0], width: toSum[0].width);
+    final b = addInput('b', toSum[1], width: toSum[1].width);
+    carry = addInput('carry_in', carry, width: carry.width);
```
mkorbel1, line 37: This carry is not an input to the module, so delete this line.

test/ripple_carry_adder_test.dart
```diff
@@ -0,0 +1,55 @@
...
+    test('should throw exception if toSum length is not two.', () {
+      final a = Logic(name: 'a', width: 8);
+      final b = Logic(name: 'b', width: 8);
+      final c = Logic(name: 'c', width: 8);
+
+      expect(() => RippleCarryAdder(toSum: [a, b, c]),
+          throwsA(const TypeMatcher<RohdHclException>()));
+    });
+
+    test('should return correct value for ripple carry adder.', () {
```
mkorbel1, line 40: Can we add some corner cases? E.g.

- 0+0
- ?+0
- Carry bit non-zero
- Overflow
- A bunch of random numbers in a loop

lib/src/carry_save_mutiplier.dart
```diff
@@ -0,0 +1,126 @@
...
+
+  /// The list pf carry from every pipeline stages.
+  late final List<Logic> _carry;
+
+  /// The final product of the multiplier module.
+  @override
+  Logic get product => output('product');
+
+  /// The pipeline for [CarrySaveMultiplier].
+  late final Pipeline pipeline;
```
mkorbel1, line 37: pipeline should be private

lib/src/carry_save_mutiplier.dart
```diff
@@ -0,0 +1,126 @@
...
+
+  /// Construct a [CarrySaveMultiplier] that multiply two list of logic in
+  /// [toMultiply].
+  CarrySaveMultiplier(Logic clk, Logic reset,
```
mkorbel1, line 41: let's make `clk` and `reset` named arguments so it's not ambiguous and to avoid people putting `a` and `b` there.

test/carry_save_multiplier_test.dart
```diff
@@ -0,0 +1,82 @@
...
+    unawaited(Simulator.run());
+
+    Future<void> waitCycles(int numCycles) async {
+      for (var i = 0; i < numCycles; i++) {
+        await clk.nextPosedge;
+      }
+    }
+
+    final inputs = List.generate(
+        10, (index) => List.generate(2, (index) => Random().nextInt(100) + 1));
```
mkorbel1, line 67: Instead of 100, make it fill the width (`1<<widthLength`)

test/carry_save_multiplier_test.dart
```diff
@@ -0,0 +1,82 @@
...
+
+    for (final input in inputs) {
+      a.put(input[0]);
+      b.put(input[1]);
+
+      await waitCycles(widthLength * 2).then((value) {
```
mkorbel1, line 73: I think you want `unawaited`, not `await` on this line, so that they are going in parallel

mkorbel1 (changes requested):
This is looking great

doc/multiplier.md
```diff
@@ -0,0 +1,24 @@
...
+An example is shown below to multiply two inputs of signals that have 8-bits of width.
+
+```dart
+const dataWidth = 8;
+final clk = SimpleClockGenerator(10).clk;
+final reset = Logic(name: 'reset');
+final a = Logic(name: 'a', width: dataWidth);
+final b = Logic(name: 'b', width: dataWidth);
+
+final multiply = CarrySaveMultiplier(clk: clk, reset: reset, a, b, name: 'csm_module');
```
mkorbel1, line 22: In this example you should show getting the `product` from it as well?

doc/adder.md
```diff
@@ -0,0 +1,24 @@
...
+
+- [Ripple Carry Adder](#ripple-carry-adder)
+
+## Ripple Carry Adder
+
+An ripple carry adder is a digital circuit used for binary addition. It consists of a series of full adders connected in a chain, with the carry output of each adder linked to the carry input of the next one. Starting from the least significant bit (LSB) to most significant bit (MSB), the adder sequentially adds corresponding bits of two binary numbers.
+
+The [`RippleCarryAdder`](https://intel.github.io/rohd-hcl/rohd_hcl/RippleCarryAdder-class.html) module in ROHD-HCL accept input parameters `Logic`s' a and b as the input pin and the name of the module `name`. Note that the width of the inputs must be the same or `RohdHclException` will be thrown.
+
+An example is shown below to multiply two inputs of signals that have 8-bits of width.
```
mkorbel1, line 13: Typo, you say `multiply` here but this is an adder

lib/src/adder.dart
```diff
@@ -0,0 +1,58 @@
...
+// 2023 June 1
+// Author: Yao Jing Quek <yao.jing.quek@intel.com>
+//
+
+import 'package:rohd/rohd.dart';
+
+/// An abstract class for all adder module.
+abstract class Adder extends Module {
+  /// The input to the adder pin [a].
+  Logic a;
```
mkorbel1, line 16: the `input` signals of the class should _not_ be publicly visible.  they should also probably be `final` or at least `late final`.  At least we should use `@protected` if we don't find a nice way to make them private.

same applies to the multiplier

lib/src/adder.dart
```diff
@@ -0,0 +1,58 @@
...
+
+  /// The input to the adder pin [b].
+  Logic b;
+
+  /// The addition results [sum].
+  Logic get sum;
+
+  /// Takes in input [a] and input [b] and return the [sum] of the addition
+  /// result.
+  Adder(this.a, this.b, {super.name});
```
mkorbel1, line 26: You could take care of the `addInput`s, width checks, etc. in here since they should be common across all adder implementations.

same applies to the multiplier

test/ripple_carry_adder_test.dart
mkorbel1: missing the random numbers in a loop test?

mkorbel1 (changes requested):

test/ripple_carry_adder_test.dart
quekyj:
```dart
test('should return correct value when random numbers is given.', () async {
```

The test is here.

lib/src/adder.dart
```diff
@@ -0,0 +1,66 @@
...
+
+  /// Constructs a [FullAdder] with value [a], [b] and [carryIn] based on
+  /// full adder truth table.
+  FullAdder({
+    required Logic a,
+    required Logic b,
+    required Logic carryIn,
+    super.name = 'full_adder',
+  }) {
+    a = addInput('a', a, width: a.width);
```
mkorbel1, line 53: Move these `addInput`s for `a` and `b` into `Adder`, since you made the protected variables there.  That will also probably force them to be `late final` in `Adder`.

lib/src/carry_save_mutiplier.dart
```diff
@@ -0,0 +1,125 @@
...
+
+  /// The pipeline for [CarrySaveMultiplier].
+  late final Pipeline _pipeline;
+
+  /// Construct a [CarrySaveMultiplier] that multiply input [a] and input [b].
+  CarrySaveMultiplier(super.a, super.b,
+      {required Logic clk,
+      required Logic reset,
+      super.name = 'carry_save_multiplier'}) {
+    final portA = addInput('a', a, width: a.width);
```
mkorbel1, line 47: Same here, these should be moved into `Multiplier` and then use `a` and `b` inherited from there

pubspec.yaml
```diff
@@ -10,6 +10,7 @@ environment:
   sdk: '>=2.18.0 <3.0.0'
 
 dependencies:
+  meta: ^1.9.1
```
mkorbel1, line 13: New dependency, will throw this through the review process

mkorbel1 (changes requested):

lib/src/adder.dart
quekyj: Its seems like I need to removed the final.

```dart
/// An abstract class for all adder module.
abstract class Adder extends Module {
  /// The input to the adder pin [a].
  @protected
  late final Logic a;

  /// The input to the adder pin [b].
  @protected
  late final Logic b;

  /// The addition results [sum].
  Logic get sum;

  /// Takes in input [a] and input [b] and return the [sum] of the addition
  /// result. The width of input [a] and [b] must be the same.
  Adder(this.a, this.b, {super.name}) {
    if (a.width != b.width) {
      throw RohdHclException('inputs of a and b should have same width.');
    }
    a = addInput('a', a, width: a.width);
    b = addInput('b', b, width: b.width);
  }
}
```

Since `a` and `b` will be reinitialized.

mkorbel1: You should make it `late final`

mkorbel1: And don't make the inputs `this.a` and `this.b`, just `Logic a` and `Logic b`, then `this.a = addInput...`

mkorbel1 (approved):
Nice work! Looks good!

PR #30, opened 2023-05-23
quick fix for sch generation (one_hot)

desmonddak:
## Description & Motivation

very quick fix to unblock generation of HTML schematics

## Testing

Tested create_htmls locally.

## Backwards-compatibility

No.

## Documentation

No


mkorbel1 (approved):
looks good to me

PR #33, opened 2023-06-08
Checking and tracking for `Fifo`

mkorbel1:
## Description & Motivation

Adds a ROHD-VF `FifoChecker` and `FifoTracker` that can be used to validate proper usage of the `Fifo` in designs.

Also, refactored `Fifo` a little and removed a few unnecessary gates depending on configuration.

## Related Issue(s)

N/A

## Testing

Added new tests

## Backwards-compatibility

No

## Documentation

Yes, updated docs


mkorbel1:
Tests are affected by https://github.com/intel/rohd-vf/issues/45 but we mute the print statements, so no big deal

PR #38, opened 2023-07-11
Round Robin Arbiter

aherrera97intl:
Added Round Robin Arbiter along with tests

## Description & Motivation

Added a Round Robin component which is required to handle hardware requests in order.

## Testing

Added several tests where requests are sent and an output is expected for each clock cycle

## Backwards-compatibility

No

## Documentation

Not included yet


lib/src/arbiter.dart
```diff
@@ -60,3 +60,57 @@ class PriorityArbiter extends Arbiter {
     ]);
   }
 }
+
+/// Arbiter Roun Robin.
```
mkorbel1, line 64: Typo, should say "Round Robin Arbiter"

lib/src/arbiter.dart
```diff
@@ -60,3 +60,57 @@ class PriorityArbiter extends Arbiter {
...
+class RoundRobinArbiterlogic extends Arbiter {
```
mkorbel1, line 65: Let's just call it `RoundRobinArbiter` without the `logic` at the end

lib/src/arbiter.dart
```diff
@@ -60,3 +60,57 @@ class PriorityArbiter extends Arbiter {
...
+  ///Mask to define pending requests to be attended
+  ///i.e. If bit 4 is granted in a 8 bit request, [requestMask] will be 11110000
```
mkorbel1, line 67: I don't really understand this example, why is bit #4 set to 1 if it's supposed to be cleared?  Maybe a multi-cycle example with how `requestMask` and `grantMask` would help elaborate to readers what these are supposed to do.

lib/src/arbiter.dart
```diff
@@ -60,3 +60,57 @@ class PriorityArbiter extends Arbiter {
...
+  ///turning off the bit granted along with all previous bits
+  List<Logic> requestMask = [];
```
mkorbel1, line 69: We should leave internal implementation details as private by starting with a `_`, so `_requestMask` instead of `requestMask`.  Same for `_grantMask`.

lib/src/arbiter.dart
```diff
@@ -60,3 +60,57 @@ class PriorityArbiter extends Arbiter {
...
+
+  ///Result of masking [_requests] with [requestMask]
+  ///Contains bits of [_requests] but as bits granted, will off
+  List<Logic> grantMask = [];
+
+  /// Initiliazing Round Robin
```
mkorbel1, line 75: Let's elaborate a little more on the constructor, at least mentioning that it arbitrates between `requests`.

lib/src/arbiter.dart
```diff
@@ -60,3 +60,57 @@ class PriorityArbiter extends Arbiter {
...
+  RoundRobinArbiterlogic(super.requests, Logic clk, Logic reset) {
```
mkorbel1, line 76: Positional arguments in this order feel a little confusing.  Maybe some of them should be named parameters?  For consistency with the other arbiters, maybe like this?
```dart
RoundRobinArbiterlogic(super.requests, {required Logic clk, required Logic reset})
```

lib/src/arbiter.dart
```diff
@@ -60,3 +60,57 @@ class PriorityArbiter extends Arbiter {
...
+    reset = addInput('reset', reset);
+    requestMask = List.generate(count, (i) => Logic(name: 'requestMask$i'));
+    grantMask = List.generate(count, (i) => Logic(name: 'grantMask$i'));
+    Sequential(clk, [
+      If(reset, then: [
+        for (var g = 0; g < count; g++) requestMask[g] < 1,
+        for (var g = 0; g < count; g++) _grants[g] < 0,
+        for (var g = 0; g < count; g++) grantMask[g] < _requests[g],
+      ], orElse: [
+        /// Added [_grants] to turn off granted bit when a reset on
```
mkorbel1, line 87: The three `///` is for doc comments on public members and functions, use just `//` for other comments.

mkorbel1, line 87: These comments are a little confusing, can you try clarifying them?

test/arbiter_test.dart
```diff
@@ -35,4 +35,100 @@ void main() {
...
+    vector.put(bin('10000101'));
+    await clk.nextNegedge;
+    expect(grants.value, LogicValue.ofString('00000100'));
+    vector.put(bin('01001000'));
+    await clk.nextNegedge;
+    expect(grants.value, LogicValue.ofString('00001000'));
+    vector.put(bin('10010001'));
+    await clk.nextNegedge;
+    expect(grants.value, LogicValue.ofString('00010000'));
+    vector.put(bin('10000000'));
```
mkorbel1, line 114: Don't we want the grant to be relative to the *current* set of requests, not the *previous cycle's* requests?

mkorbel1 (changes requested):

doc/arbiter.md
```diff
@@ -7,3 +7,8 @@ ROHD HCL implements a generic `abstract` [`Arbiter`](https://intel.github.io/roh
 The [`PriorityArbiter`](https://intel.github.io/rohd-hcl/rohd_hcl/PriorityArbiter-class.html) is a combinational (stateless) arbiter that always grants to the lowest-indexed request.
 
 [PriorityArbiter Schematic](https://intel.github.io/rohd-hcl/PriorityArbiter.html)
+
+## Round Robin Arbiter
+
+The Roun Robin Arbiter is a sequential (stateless) arbiter which grants requests same ways as Priority Arbiter, but keeps track of last request granted.
```
mkorbel1, line 13: It's not "stateless" because it holds state about the previous request granted :)

doc/arbiter.md
```diff
@@ -7,3 +7,8 @@ ROHD HCL implements a generic `abstract` [`Arbiter`](https://intel.github.io/roh
...
+The arbiter uses two masks to determine the next request to be attended.
```
mkorbel1, line 14: The masks are an implementation detail and not relevant to the user, we can remove this.

lib/src/arbiter.dart
```diff
@@ -60,3 +60,71 @@ class PriorityArbiter extends Arbiter {
     ]);
   }
 }
+
+/// Round Robin Arbiter.
+class RoundRobinArbiter extends Arbiter {
+  ///Mask to define pending requests to be attended
```
mkorbel1, line 66: Add a space between `///` and the text

lib/src/arbiter.dart
```diff
@@ -60,3 +60,71 @@ class PriorityArbiter extends Arbiter {
...
+  ///i.e. If bit 4 is granted in a 8 bit request, [_requestMask] will be 11110000
+  ///turning off the bit granted along with all previous bits
+  List<Logic> _requestMask = [];
+
+  ///Result of masking [_requests] with [_requestMask]
+  ///Contains bits of [_requests] but as bits are granted, will turn off
+  List<Logic> _grantMask = [];
+
+  List<Logic> test = [];
```
mkorbel1, line 75: What's this?  Why is it public?

lib/src/arbiter.dart
```diff
@@ -60,3 +60,71 @@ class PriorityArbiter extends Arbiter {
...
+                  for (var g = 0; g < count; g++)
+                    _requestMask[g] < (i < g ? 1 : 0),
+                ])
+        ], defaultItem: [
+          for (var g = 0; g < count; g++) _requestMask[g] < 1,
+        ])
+      ])
+    ]);
+    Combinational([
+      //If(reset, then: [
```
mkorbel1, line 103: remove code you don't need

lib/src/arbiter.dart
```diff
@@ -60,3 +60,71 @@ class PriorityArbiter extends Arbiter {
...
+      //for (var g = 0; g < count; g++) grants[g] < 0,
+      //]),
+      for (var g = 0; g < count; g++)
+        _grantMask[g] < _requestMask[g] & _requests[g] & ~reset,
```
mkorbel1, line 107: why do you need reset in the combinational logic?

test/arbiter_test.dart
```diff
@@ -35,4 +35,129 @@ void main() {
...
+    //expect(grants.value, LogicValue.ofString('00000100'));
+    vector.put(bin('01001000'));
+    await clk.nextNegedge;
+    //expect(grants.value, LogicValue.ofString('00001000'));
+    vector.put(bin('10010001'));
+    await clk.nextNegedge;
+    //expect(grants.value, LogicValue.ofString('00010000'));
+    vector.put(bin('10000000'));
+    await clk.nextNegedge;
+    //expect(grants.value, LogicValue.ofString('10000000'));
```
mkorbel1, line 120: all the `expect`s are commented out?

mkorbel1 (changes requested):

doc/arbiter.md
```diff
@@ -7,3 +7,7 @@ ROHD HCL implements a generic `abstract` [`Arbiter`](https://intel.github.io/roh
 The [`PriorityArbiter`](https://intel.github.io/rohd-hcl/rohd_hcl/PriorityArbiter-class.html) is a combinational (stateless) arbiter that always grants to the lowest-indexed request.
 
 [PriorityArbiter Schematic](https://intel.github.io/rohd-hcl/PriorityArbiter.html)
+
+## Round Robin Arbiter
+
+The Roun Robin Arbiter is a sequential (stateful) arbiter which grants requests same ways as Priority Arbiter, but keeps track of last request granted.
```
mkorbel1, line 13: Typo: "Round Robin"

mkorbel1, line 13: It does more than "keep track", it uses it to try to allocate grants "fairly" between all requestors

lib/src/arbiter.dart
```diff
@@ -60,3 +60,84 @@ class PriorityArbiter extends Arbiter {
...
+/// Round Robin Arbiter.
+class RoundRobinArbiter extends Arbiter {
+  /// Mask to define pending requests to be attended
+  /// the main idea is to mask the requests as they are granted to ensure all
+  /// requests are granted once before resetting the mask
+  /// example:
+  /// [clk] [_requests] [_requestMask]   [_grantMask]   [_grants]
+  ///  0    00100010   &  11111111      =   00100010  ->  00000010
+  ///  1    00100010   &  11111100      =   00100000  ->  00100000
+  ///  2    10000001   &  11000000      =   00100000  ->  10000000
```
mkorbel1, line 73: i'm confused by the grantMask on this line?

lib/src/arbiter.dart
```diff
@@ -60,3 +60,84 @@ class PriorityArbiter extends Arbiter {
...
+      If(reset, then: [
+        for (var g = 0; g < count; g++) _requestMask[g] < 1,
+      ], orElse: [
+        // Use [_grants] to turn the bit granted to 0 along with all bits
+        // to the right to avoid, in case if  [_requests] changes, granting a
+        // previous request
+        // Example:
+        // [_grants] [requestMask]
+        // 00001000   11110000
+        CaseZ(_grants.rswizzle(), [
```
mkorbel1, line 101: We should probably put a `ConditionalType` on this, maybe `unique`?

lib/src/arbiter.dart
```diff
@@ -60,3 +60,84 @@ class PriorityArbiter extends Arbiter {
...
+          for (var i = 0; i < count; i++)
+            CaseItem(
+                Const(LogicValue.filled(count, LogicValue.z)
```
mkorbel1, line 104: I think you can set them to 0 instead of `z` since only one grant will ever be set?

lib/src/arbiter.dart
```diff
@@ -60,3 +60,84 @@ class PriorityArbiter extends Arbiter {
...
+          for (var g = 0; g < count; g++) _requestMask[g] < 1,
+        ])
+      ])
+    ]);
+    Combinational([
+      for (var g = 0; g < count; g++)
+        _grantMask[g] < _requestMask[g] & _requests[g],
+      // CaseZ uses [_grantMask] to set the [_grants] based on the
+      // least significant bit
+      CaseZ(
```
mkorbel1, line 121: we should make this `CaseZ` a `ConditionalType.priority` i think?

lib/src/arbiter.dart
```diff
@@ -60,3 +60,84 @@ class PriorityArbiter extends Arbiter {
...
+  /// requestMask, therefore grantMask changes to grant the next request
+  /// Clock 3 shows when a request changes, the requestMask make sure to
+  /// handle the missing requests before resetting
+  List<Logic> _requestMask = [];
+
+  /// Result of masking [_requests] with [_requestMask]
+  /// Contains bits of [_requests] but as bits are granted, will turn off
+  List<Logic> _grantMask = [];
+
+  /// Round Robin arbiter handles requests keeping track of previous requests.
```
mkorbel1, line 84: Could update this doc comment a little too, to explain that its fair

mkorbel1 (changes requested):
This is looking much better, great work!  The comments really help explain what it's doing and why and the algorithm makes sense.

lib/src/arbiter.dart
```diff
@@ -98,10 +100,10 @@ class RoundRobinArbiter extends Arbiter {
         // Example:
         // [_grants] [requestMask]
         // 00001000   11110000
-        CaseZ(_grants.rswizzle(), [
+        CaseZ(_grants.rswizzle(), conditionalType: ConditionalType.unique, [
```
mkorbel1, line 103: This doesn't need to be a CaseZ anymore if you're not using `z`'s

mkorbel1 (changes requested):
Looks good!  Need to resolve the merge conflicts

test/arbiter_test.dart
```diff
@@ -36,10 +36,9 @@ void main() {
     expect(grantVec.value, LogicValue.ofString('00000100'));
   });
   test('round robin logic', () async {
+    await Simulator.reset();
```
mkorbel1, line 39: Let's add this once at the top of main instead of in each test:

```dart
  tearDown(() async {
    await Simulator.reset();
  });
```

mkorbel1 (approved):
Looks good!

Issue #48, opened 2023-09-29
`RegisterFile` is incorrectly using `reset` in `Combinational`


mkorbel1:
### Describe the bug

There's no need to use `reset` in combinational logic.  It is probably a remnant of an old implementation.  It should be removed.

PR #49, opened 2023-09-29
https://github.com/intel/rohd-hcl/issues/48

chaparalas:
## Description & Motivation

[](https://github.com/intel/rohd-hcl/issues/48)
Remove reset on the rd-data out path.

## Backwards-compatibility

## Documentation

NO


mkorbel1 (approved):
Looks good, thanks!

Issue #12, opened 2023-04-03
Prefix trees


mkorbel1:
### Motivation

Useful for efficient encoders and arithmetic

### Desired solution

Implement generic capabilities for generating prefix trees

PR #51, opened 2023-10-03
parallel prefix operations

desmonddak:
## Description & Motivation

Components based on prefix tree computation: PPAdder, PPIncr, PPDecr, PPOrscan, PriorityEncoder.
Uses Ripple, Sklansky, KoggeStone, and BrentKung prefix tree patterns.

## Related Issue(s)

#12 

## Testing

Tests included for each component with each flavor of prefix tree.

## Backwards-compatibility

No

## Documentation

Yes.  Documentation for the components, plus a generator for one Adder schematic to test publication.


CONTRIBUTING.md
```diff
@@ -151,3 +151,22 @@ Here are some links to help guide you on style as recommended by Dart:
...
+
+1. Create a generator in gen
+
+   1. Add generator to gen/generate.dart
+
+   1. Run dart gen/generate.dart to make sure generation works.
+
+1. Add documentation to doc/name.dart
+
+1. How to test documentation in a dart container
```
mkorbel1, line 169: the dev container already has these installed it seems?

lib/src/parallel_prefix_operations.dart
```diff
@@ -0,0 +1,250 @@
...
+
+/// ParallePrefix is the core parallel prefix tree structure
+class ParallelPrefix extends Module {
+  final List<Logic> _oseq = [];
+
+  /// Output sequence value
+  List<Logic> get val => _oseq;
+
+  /// ParallePrefix recursion
+  ParallelPrefix(List<Logic> inps, String name) : super(name: name) {
```
mkorbel1, line 28: In general, is there a reason why we use `List<Logic>` where each is 1-bit instead of just a single `Logic` bus with varying width?  It seems like it would be easier to use that way as well

lib/src/parallel_prefix_operations.dart
```diff
@@ -0,0 +1,250 @@
...
+      iseq[i] = op(iseq[i - 1], iseq[i]);
+    }
+
+    iseq.forEachIndexed((i, el) {
+      _oseq[i] <= el;
+    });
+  }
+}
+
+/// Or scan based on pp tree
```
mkorbel1, line 154: In both comments and module names, can we write out `ParallelPrefix` instead of `PP` so that it's more clear to those unfamiliar?

lib/src/parallel_prefix_operations.dart
```diff
@@ -0,0 +1,250 @@
...
+            a.width, (i) => [a[i] & b[i], a[i] | b[i]].swizzle()),
+        (lhs, rhs) => [rhs[1] | rhs[0] & lhs[1], rhs[0] & lhs[0]].swizzle());
+    addOutput('out', width: a.width) <=
+        List<Logic>.generate(a.width,
+                (i) => (i == 0) ? a[i] ^ b[i] : a[i] ^ b[i] ^ u.val[i - 1][1])
+            .rswizzle();
+  }
+}
+
+/// Incrementer based on pp tree
```
mkorbel1, line 212: Can you elaborate more on the incrementer and decrementer, what the input should be and what operation is performed?  Comments like "output sequence" are a little too generic

doc/components/parallel_prefix_operations.md
```diff
@@ -0,0 +1,15 @@
+# Parallel Prefix Operations
+
+ROHD HCL implements a set of parallel prefix compute operations using
+different parallel prefix computation trees.
+
+For example, we have unary operations like a word 'or' [`OrScan`] class, and a priority encoder [`PriorityEncoder`] class. We have simple unary arithmetic operations like an increment [`PPIncr`] class, and a decrement [`PPDecr`] class. Finally, we have a binary adder [`PPAdder`] class.
+
+Each of these operations can be implemented with different ['ParallelPrefix'] types:
+
```
mkorbel1, line 9: An example or link to some other information about how to use parallel prefix operations could be helpful in here for people less familiar

pubspec.yaml
```diff
@@ -10,6 +10,7 @@ environment:
   sdk: '>=2.18.0 <3.0.0'
 
 dependencies:
+  collection: ^1.18.0
```
mkorbel1, line 13: NOTE: need to update open-source dependencies prior to merge (I'll take care of it, just marking it here so I don't forget)

test/parallel_prefix_operations_test.dart
```diff
@@ -0,0 +1,236 @@
...
+        b.put(bb);
+        final result = mod.out.value.toInt();
+        //print("adder: $aa $bb $result $golden");
+        expect(result, equals(golden));
+      }
+    }
+  });
+}
+
+BigInt genRandomBigInt(int inBits) {
```
mkorbel1, line 97: ROHD now has a way to generate random `LogicValue`s via an extension on the `Random` object

mkorbel1 (changes requested):
Great addition to the library!

CONTRIBUTING.md
desmonddak: I don't follow.  These are instructions for the component developer to go through in adding a component to ROHD-HCL.  Only the last bullet asks the question: how do you see the output of documentation (including schematics) -- I want to document how to check this out before deployment.  If it means using the dev container, I think we need step-by-step instructions on how to use to check your documentation.

lib/src/parallel_prefix_operations.dart
desmonddak: Each of these items in the List are a multi-bit Logic (2).  See line 202, where the basic operations or AND and OR of the input bits, swizzled into a two-bit representation to feed into the tree.  So each input of the tree is a two-bit Logic and this carries through the core computation.  It might be more clear if this 'Logic' was declared as a type that carried just two bits.

lib/src/parallel_prefix_operations.dart
desmonddak: Sure.  But ParallelPrefixAdder is really long,  Should we do ParPrefixAdder?  Or stick with the full name.

lib/src/parallel_prefix_operations.dart
desmonddak: Yes, will do.

CONTRIBUTING.md
mkorbel1: The file `tool/gh_codespaces/run_setup.sh` does necessary setup including installing open cad suite and d3.  This is the script used in `.devcontainer/devcontainer.json` which is used by VS Code to build a container for ROHD-HCL development, and also is used by GitHub codespaces.  A pointer to how to use the container and/or saying to run the requirements in that script (if needed) might make more sense?

lib/src/parallel_prefix_operations.dart
mkorbel1: Ah I see, ok that makes sense to me.  Maybe a little more explanation on the doc comment for `ParallelPrefix` could help elaborate why it's a list and what sorts of things are expected?

lib/src/parallel_prefix_operations.dart
mkorbel1: I've been trying to follow Dart style and design conventions where possible since they are well organized and thoughtful, and making a new set of conventions for ROHD specifically seems like a daunting task.

Dart recommends avoiding abbreviations where possible:
https://dart.dev/effective-dart/design#avoid-abbreviations

With autocomplete in modern IDEs, I'm inclined to agree with it, and I think it's worth the extra few letters for the clarity it provides as an API

CONTRIBUTING.md
desmonddak: I think that may help with the last line of checking documentation / schematics.  It would be good to have a recipe for seeing the output so people can tweak their documentation / schematic examples locally.

And to add a new Widget, we need to tell people that in addition to adding lib/src/Widget.dart and test/Widget_test.dart, they need to also add gen/Widget_gen.dart to create an example for schematic, then modify gen/generate.dart to call that Widget generator, then reference the schematic in doc/components/Widget.md.  Finally, we need a pointer how to invoke via vscode and/or codespaces to check the schematic output (in my case the output was empty because of a naming bug).

I did file a bug that if you do all this and put a link to the final site in your pull request, it will fail because the link is dead and the checking is strict on that.

lib/src/parallel_prefix_operations.dart
desmonddak: Would it be crazy to create a local two-bit Logic type?

test/parallel_prefix_operations_test.dart
desmonddak: Studying now.

CONTRIBUTING.md
mkorbel1: Ok, I think once the Config generation stuff is merged in this will be more streamlined, but I agree some instructions here are helpful.

I'm not sure the best way to handle that issue.  Perhaps the right way is to add links after the main code has merged in?

lib/src/parallel_prefix_operations.dart
mkorbel1: Are all use cases of prefix trees always 2-bit?

If we want `ParallelLogic` to be an API, we would need to consider usage of the API.  Do we want people to have to construct a special type of `Logic` in order to use the API?

test/parallel_prefix_operations_test.dart
mkorbel1: See https://intel.github.io/rohd/rohd/RandLogicValue.html

lib/src/parallel_prefix_operations.dart
desmonddak: I know the Logic type is a single length, typically 2, but there are longer variants.  These are things like carry-save, generate-propagate or generate-kill.  I found a 4-bit version, for example.  So ParallelPrefix is really just a node in a tree of these, which transform N k-bit quantities in a row to another N k-bit quantities in the next row to form a tree.

CONTRIBUTING.md
```diff
@@ -151,3 +151,22 @@ Here are some links to help guide you on style as recommended by Dart:
...
+
+1. Create testing code in test/
+
+1. Create a generator in gen
+
+   1. Add generator to gen/generate.dart
+
+   1. Run dart gen/generate.dart to make sure generation works.
+
+1. Add documentation to doc/name.dart
```
mkorbel1, line 167: name.md?

doc/components/parallel_prefix_operations.md
```diff
@@ -0,0 +1,16 @@
+# Parallel Prefix Operations
+
+ROHD HCL implements a set of parallel prefix compute operations using different parallel prefix computation trees based on the ['ParallePrefix'] node class carrying carry/save or generate/propagate bits.
+
+For example, we have unary operations like a word-level 'or' [`OrScan`] class, and a priority encoder [`PriorityEncoder`] class which computes the position of the first bit set to '1'. We have simple unary arithmetic operations like an increment [`PPIncr`] class, and a decrement [`PPDecr`] class. Finally, we have a binary adder [`PPAdder`] class. For background on basic parallel prefix adder structures, see <https://en.wikipedia.org/wiki/Kogge%E2%80%93Stone_adder>.
```
mkorbel1, line 5: Update names of classes in here?

lib/src/parallel_prefix_operations.dart
```diff
@@ -0,0 +1,255 @@
+// Copyright (C) 2023 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// parallel-prefix_operations.dart
```
mkorbel1, line 4: file name typo, `-` instead of `_`

lib/src/parallel_prefix_operations.dart
```diff
@@ -0,0 +1,255 @@
...
+import 'dart:math';
+
+import 'package:collection/collection.dart';
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+
+/// This computes the power of 2 less than x
+int largestPow2LessThan(int x) => pow(2, log2Ceil(x) - 1).toInt();
+
+/// ParallePrefix is the core parallel prefix tree structure node
```
mkorbel1, line 20: Typo `ParallePrefix`, and you can put `[]` around it to make the doc pick it up better `[ParallelPrefix]`

lib/src/parallel_prefix_operations.dart
```diff
@@ -0,0 +1,255 @@
...
+/// The output is a List of multi-bit Logic vectors (typicall 2-bit) that
```
mkorbel1, line 21: typo: "typically"

lib/src/parallel_prefix_operations.dart
```diff
@@ -0,0 +1,255 @@
...
+/// represent things like carry-save or generate-propagate signaling in adder
+/// networks.  Each node in a parallel prefix tree transforms a row of inputs
+/// to an equal length row of outputs of these multi-bit Logic values.
+class ParallelPrefix extends Module {
+  final List<Logic> _oseq = [];
+
+  /// Output sequence value
+  List<Logic> get val => _oseq;
```
mkorbel1, line 29: I think wrap `_oseq` so that it cant be modified by users: `UnmodifiableListView(_oSeq)`.

mkorbel1 (changes requested):
Looks really good now!  Just a few things to change

CONTRIBUTING.md
```diff
@@ -151,3 +151,22 @@ Here are some links to help guide you on style as recommended by Dart:
 * Design: <https://dart.dev/guides/language/effective-dart/design>
 
 We recommend following these same guidelines for any of your own packages you may create for the ecosystem, as well.
+
+## Adding a New Component
```
mkorbel1, line 155: We should update this section for the new generation flow with the configurator

mkorbel1: Actually, I copied this section into another branch and updated it based on recent developments, so we can remove it from this PR.

conflicts.txt
mkorbel1, line 1: I think this was mistakenly committed?

doc/onehot.md
mkorbel1, line 1: This file was moved into doc/components/ instead of just doc/.  Merge it to there?

tool/converters/verilog_json.sh
mkorbel1, line 1: incomplete merge?

tool/gh_codespaces/run_setup.sh
mkorbel1, line 1: incomplete merge?

mkorbel1 (changes requested):

lib/src/parallel_prefix_operations.dart
```diff
@@ -0,0 +1,255 @@
...
+//   Borrowed from https://github.com/stevenmburns/rohd_sklansky.git
+
+import 'dart:math';
+
+import 'package:collection/collection.dart';
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+
+/// This computes the power of 2 less than x
+int largestPow2LessThan(int x) => pow(2, log2Ceil(x) - 1).toInt();
```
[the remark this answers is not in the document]
desmonddak: This function is used in the test module as well.

desmonddak:
merge failures highlighted a lot of older code in my branch -- it would be easier to resync my fork and push the latest files for prefix.

PR #76, opened 2024-02-05
New prefix tree

desmonddak:
## Description & Motivation

Components based on prefix-tree computation renamed for more clarityl

## Related Issue(s)

#12 

## Testing

Tests for each component and for each flavor of implementing prefix tree.

## Backwards-compatibility

No

## Documentation

Yes.  Documentation for the components added, plus a schematic generation of an Adder to test publication.


desmonddak:
I appear to have pushed old changes to my main branch and then to my prefix branches and need to reset my fork.

PR #79, opened 2024-02-10
Flutter experiment

desmonddak:
## Description & Motivation

This change adds flutter to devcontainer.

## Related Issue(s)

None.

## Testing

Ran using Codespaces and made sure the Configurator flutter application worked.

## Backwards-compatibility

No.

## Documentation

Yes.  Added a few lines to CONTRIBUTING that documents how to startup flutter to use the component Configurator.

### DIFF
--- a/tool/gh_codespaces/install_flutter.sh
+++ b/tool/gh_codespaces/install_flutter.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+# Copyright (C) 2023 Intel Corporation
+# SPDX-License-Identifier: BSD-3-Clause
+#
+# install_dart.sh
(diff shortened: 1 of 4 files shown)


tool/gh_codespaces/install_flutter.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+# Copyright (C) 2023 Intel Corporation
```
mkorbel1, line 3: header of this script is wrong (copyright year, name of file, description, author, date, etc.)

mkorbel1 (changes requested):

PR #82, opened 2024-03-20
parallel prefix documentation updates

desmonddak:
## Description & Motivation

This PR updates parallel prefix documentation and links to classes that were not possible on first check-in.

During CI: discovered problem with google-fonts where the BKM is to lock it to a recent but not latest version:
https://stackoverflow.com/questions/78090711/error-type-fontfeature-not-found-flutter-google-fonts-package-error

During CI: discovered a problem with flutter code shadowing arguments of a parent class triggering an error entitled use_super_parameters:
https://dart.dev/tools/linter-rules/use_super_parameters

## Related Issue(s)

None

## Testing

I've tested the documentation links in my fork.   I've also added a linear adder test for narrow adders to complement the random test.

## Backwards-compatibility

No

## Documentation

Changes are to the documentation, so they are included.

### DIFF
--- a/confapp/pubspec.yaml
+++ b/confapp/pubspec.yaml
@@ -19,7 +19,7 @@ dependencies:
   sidebarx: ^0.16.0
   flutter_bloc: ^8.1.3
   bloc: ^8.1.2
-  google_fonts: ^6.1.0
+  google_fonts: 6.1.0
 
 dependency_overrides:
   rohd_hcl:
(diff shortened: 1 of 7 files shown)


confapp/pubspec.yaml
```diff
@@ -19,7 +19,7 @@ dependencies:
   sidebarx: ^0.16.0
   flutter_bloc: ^8.1.3
   bloc: ^8.1.2
-  google_fonts: ^6.0.0
+  google_fonts: 6.1.0
```
mkorbel1, line 22: why not `^6.1.0`?

mkorbel1 (approved):

confapp/pubspec.yaml
mkorbel1: Ah thanks, and sorry I missed that in the original description!

mkorbel1 (approved):

PR #89, opened 2024-09-06
`Sum` and `Counter`

### LINKED ISSUE #73: Binary Counter
### Motivation

Configurable counters are useful!

### Desired solution

- Reset with reset value
- Saturation value (maximum value)
- Saturation behavior (overflow vs. hold at saturated value)
- Enable (increment) -> when high: increment, when low: don't increment
- Enable decrement
- Increment amount
- Decrement amount
- number of increments/decrements (different constructors maybe for 1 vs multiple)

### Alternatives considered

_No response_

### Additional details

_No response_

mkorbel1:
## Description & Motivation

A `Sum` and `Counter` component.  The `Sum` adds up all the things input combinationally, and the `Counter` adds a flop (essentially).  The components are super flexible and generic.

## Related Issue(s)

Fix #73 

## Testing

Added new tests

## Backwards-compatibility

No

## Documentation

Yes


doc/components/summation.md
desmonddak, line 1: Is there another obvious use of summation outside the counter?

lib/src/component_config/components/config_summation.dart
desmonddak, line 1: config_counter should be config_summation

lib/src/component_config/config_knobs/int_optional_config_knob.dart
desmonddak, line 1: Do we need to refactor other components to use this knob?  or only for those that have optional inputs?

lib/src/summation/counter.dart
```diff
@@ -0,0 +1,150 @@
...
+            minValue: minValue,
+            width: width,
+            saturates: saturates,
+            name: name);
+
+  /// Creates a [Counter] that counts up by all of the provided [logics],
+  /// including much of the other available configuration in the default
+  /// constructor.
+  ///
+  /// All [logics] are always incrementing and controled optionally by a single
```
desmonddak, line 119: controlled

lib/src/summation/counter.dart
```diff
@@ -0,0 +1,150 @@
...
+  /// width information (e.g. a [LogicValue]), or by making it large enough to
+  /// fit [maxValue], or by inspecting widths of [interfaces]. There must be
+  /// enough information provided to determine the [width].
+  ///
+  /// If no [maxValue] is provided, one will be inferred by the maximum that can
+  /// fit inside of the [width].
+  ///
+  /// The [restart] input can be used to restart the counter to a new value, but
+  /// also continue to increment in that same cycle. This is distinct from
+  /// [reset] which will reset the counter, holding the [count] at [resetValue].
```
desmonddak, line 33: Do you want to define the saturates feature and the conditions like overflow and equalsMax?

lib/src/summation/sum.dart
```diff
@@ -0,0 +1,183 @@
...
+    final lowerSaturation = Logic(name: 'lowerSaturation', width: internalWidth)
+      ..gets(minValueLogicExt + zeroPoint);
+
+    final internalValue = Logic(name: 'internalValue', width: internalWidth);
+    sum <= (internalValue - zeroPoint).getRange(0, width);
+
+    final preAdjustmentValue =
+        Logic(name: 'preAdjustmentValue', width: internalWidth);
+
+    Combinational.ssa((s) => [
```
desmonddak, line 115: I think this needs a comment as to what ssa is doing.  It's a little opaque.  I think you know because you made ssa actually work...

lib/src/summation/sum_interface.dart
```diff
@@ -0,0 +1,79 @@
...
+  /// with a type that contains width information (e.g. a [LogicValue]). There
+  /// must be enough information provided to determine the [width].
+  ///
+  /// If a [fixedAmount] is provided, then [amount] will be tied to a [Const]. A
+  /// provided [fixedAmount] must be parseable by [LogicValue.of]. Note that the
+  /// [fixedAmount] will always be interpreted as a positive value truncated to
+  /// [width]. If no [fixedAmount] is provided, then [amount] will be a normal
+  /// [port] with [width] bits.
+  ///
+  /// If [hasEnable] is `true`, then an [enable] port will be added to the
```
desmonddak, line 52: I do this differently:  if someone passes in an enable in the optional list, then I add it as an input.  This avoids a boolean and gives me the Logic they want to connect as an enable.  So would you prefer to pass in the enable itself (if not passed, don't create an input)

lib/src/summation/summation_base.dart
```diff
@@ -0,0 +1,165 @@
+// Copyright (C) 2024 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// sum.dart
```
desmonddak, line 4: summation_base.dart

desmonddak (commented):
Looks good. I did not walk through all logic as it is hairy.  I learned some things like using inject and negEdge which I hadn't used before.

One thing I didn't check is if the Counter has good defaults for the common case (single increment, wraparound?)

doc/components/summation.md
mkorbel1: The `Sum` is useful as-is and offered as a separate component, is that what you meant?

lib/src/component_config/config_knobs/int_optional_config_knob.dart
mkorbel1: I think other components were able to live with the int being required always because the default could just be populated.  This is one of the first/only components where `int` configuration defaults are `null` (and make sense to be so).  I don't think there's any others to refactor, but this should be generically useful in the future.

lib/src/summation/sum_interface.dart
mkorbel1: The difference in this case is that we're generating an `Interface` object, not a `Module`, and the pattern for these is usually to accept configuration information, not actual `Logic`s, since it's not going to actually use that `Logic` for anything.  Do you still think something is worth changing here?

mkorbel1:
> if the Counter has good defaults for the common case (single increment, wraparound?)

Yes, the default is wrap-around and requires a list of interfaces.  The `Counter.simple` constructor can just be given a clock, reset, and width (or max value) and it would be a simple increment-by-one counter.  You need to provide something so width can be inferred, but beyond that I think the defaults are convenient for what most people expect.

lib/src/summation/counter.dart
mkorbel1: A good idea, adding, thanks

lib/src/summation/sum.dart
mkorbel1: sure, added some comments

lib/src/summation/counter.dart
```diff
@@ -0,0 +1,157 @@
...
+    count <=
+        flop(
+          clk,
+          sum.sum,
+          reset: reset,
+          resetValue: initialValueLogic,
+        );
+
+    // need to flop these since value is flopped
+    overflowed <= flop(clk, sum.overflowed, reset: reset);
```
desmonddak, line 81: If we had a different word for this, we might collapse under/overflow into a single output since the Counter runs in one direction.

Not sure now about equalsMax/Min:  if we have an incrementing counter, do we see equalsMin on reset? Or is this really and 'equalsLimit' and we could collapse these two into one output.

desmonddak (commented):
This PR is in great shape.  My biggest question is: do we need both max and min in the same Counter.  If not, the interface gets way simpler and easier to understand.

mkorbel1:
> do we need both max and min in the same Counter

Yes, because we could have both incrementing and decrementing inputs, with variable amounts and enables.

lib/src/summation/counter.dart
mkorbel1: We need both because we could have both incrementing and decrementing inputs, with variable amounts and enables, so it's possible for the same counter to experience min and max both

lib/src/summation/sum_interface.dart
mkorbel1: Filed #91 to help track this more generally

PR #92, opened 2024-09-16
Clean serialization

desmonddak:
## Description & Motivation

Added deserializer and serializer components.

## Related Issue(s)

None

## Testing

Tested data transfer and clock counts using sequential tests for both.

## Backwards-compatibility

No

## Documentation

serialization.md is included in the PR.


lib/src/deserializer.dart
```diff
@@ -0,0 +1,69 @@
...
+  /// Reset input
+  @protected
+  Logic get reset => input('reset');
+
+  /// Run deserialization whenever [enable] is true
+  @protected
+  Logic? get enable => input('enable');
+
+  /// Serialized input, one data item per clock
+  Logic get serialized => input('serialized');
```
mkorbel1, line 29: this should be `@protected`

lib/src/deserializer.dart
```diff
@@ -0,0 +1,69 @@
...
+
+  /// Aggregated data output
+  LogicArray get deserialized => output('deserialized') as LogicArray;
+
+  /// Valid out when data is reached
+  Logic get done => output('done');
+
+  /// Return the count as an output
+  @protected
```
mkorbel1, line 38: this should not be protected

lib/src/deserializer.dart
```diff
@@ -0,0 +1,69 @@
...
+  /// Serialized input, one data item per clock
+  Logic get serialized => input('serialized');
+
+  /// Aggregated data output
+  LogicArray get deserialized => output('deserialized') as LogicArray;
+
+  /// Valid out when data is reached
+  Logic get done => output('done');
+
+  /// Return the count as an output
```
mkorbel1, line 37: This doc comment could be improved to explain the meaning of the count (you explained it well in the markdown doc)

lib/src/deserializer.dart
```diff
@@ -0,0 +1,69 @@
...
+      {required Logic clk,
+      required Logic reset,
+      Logic? enable,
+      super.name = 'Deserializer'}) {
+    clk = addInput('clk', clk);
+    reset = addInput('reset', reset);
+    if (enable != null) {
+      enable = addInput('enable', enable);
+    } else {
+      enable = Const(1);
```
mkorbel1, line 55: i think you could leave `enable` as `null` for both the `Counter` and `flop`, since they are by default enabled if passed `null` (then you have less port generation)

lib/src/deserializer.dart
```diff
@@ -0,0 +1,69 @@
...
+
+  /// Return the count as an output
+  @protected
+  Logic get count => output('count');
+
+  /// Build a Deserializer that takes serialized input [serialized]
+  /// and aggregates it into one wide output [deserialized], one element per
+  /// clock while [enable] (if connected) is high, emitting [done] when
+  /// completing the filling of wide output [deserialized].
+  Deserializer(Logic serialized, int length,
```
mkorbel1, line 45: could make `length` a field of the class so that it's visible/queryable to others?  `final int length; ... this.length`

lib/src/deserializer.dart
```diff
@@ -0,0 +1,69 @@
...
+  @protected
+  Logic get clk => input('clk');
+
+  /// Reset input
+  @protected
+  Logic get reset => input('reset');
+
+  /// Run deserialization whenever [enable] is true
+  @protected
+  Logic? get enable => input('enable');
```
mkorbel1, line 26: use `tryInput`

lib/src/serializer.dart
```diff
@@ -0,0 +1,92 @@
...
+  @protected
+  Logic get clk => input('clk');
+
+  /// Reset input
+  @protected
+  Logic get reset => input('reset');
+
+  /// Allow serialization onto the output stream when [enable] is true
+  @protected
+  Logic? get enable => input('enable');
```
mkorbel1, line 26: use `tryInput`

lib/src/serializer.dart
```diff
@@ -0,0 +1,92 @@
...
+
+  /// Return the count as an output
```
mkorbel1, line 28: improve doc here

lib/src/deserializer.dart
```diff
@@ -0,0 +1,69 @@
...
+    if (enable != null) {
+      enable = addInput('enable', enable);
+    } else {
+      enable = Const(1);
+    }
+    serialized = addInput('serialized', serialized, width: serialized.width);
+    final cnt = Counter.simple(
+        clk: clk, reset: reset, enable: enable, maxValue: length - 1);
+    addOutput('count', width: cnt.width) <= cnt.count;
+    addOutput('done') <= cnt.overflowed;
```
mkorbel1, line 61: Is this right?  in the serializer you use `equalsMax`, and here you use `overflowed`.  I wonder if one of them is wrong?

It would also be good to be extremely clear in the doc comments for the `done` output when the user should expect the signal to go high (is it on the cycle *after* the last one was sent, or on the *same cycle* as the last one?)

lib/src/serializer.dart
```diff
@@ -0,0 +1,92 @@
...
+      super.name = 'Serializer'}) {
+    clk = addInput('clk', clk);
+    reset = addInput('reset', reset);
+    if (enable != null) {
+      enable = addInput('enable', enable);
+    } else {
+      enable = Const(1);
+    }
+    deserialized = addInputArray('deserialized', deserialized,
+        dimensions: deserialized.dimensions,
```
mkorbel1, line 59: I think we want to be stricter here, throw an exception if the length of `dimensions` is not 1?

lib/src/serializer.dart
```diff
@@ -0,0 +1,92 @@
...
+  Logic? get enable => input('enable');
+
+  /// Return the count as an output
+  Logic get count => output('count');
+
+  /// Return [done] = true when we have processed [deserialized] completely
+  Logic get done => output('done');
+
+  /// Aggregated data to serialize out
+  LogicArray get deserialized => input('deserialized') as LogicArray;
```
mkorbel1, line 35: should be `@protected`

lib/src/serializer.dart
```diff
@@ -0,0 +1,92 @@
...
+
+import 'package:meta/meta.dart';
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+
+/// Serializes wide aggregated data onto a narrower serialization stream
+class Serializer extends Module {
+  /// Clk input
+  @protected
+  Logic get clk => input('clk');
```
mkorbel1, line 18: in general (for both components` you're not using these protected input getters.  are you putting them there for future/user extension of these classes?

lib/src/serializer.dart
```diff
@@ -0,0 +1,92 @@
...
+      required Logic reset,
+      Logic? enable,
+      bool flopInput = false,
+      super.name = 'Serializer'}) {
+    clk = addInput('clk', clk);
+    reset = addInput('reset', reset);
+    if (enable != null) {
+      enable = addInput('enable', enable);
+    } else {
+      enable = Const(1);
```
mkorbel1, line 56: could leave enable null

test/serialization_test.dart
```diff
@@ -0,0 +1,312 @@
...
+//
+// ignore_for_file: invalid_use_of_protected_member
+
+import 'dart:async';
+
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+import 'package:test/test.dart';
+
+void main() {
```
mkorbel1, line 18: maybe add a test where the serializer and deserializer are connected to each other and drive two sequential chunks of data?  it's quite possible the components would be often used in pairs, and it would be good that the first time that happens isn't by a user :)

mkorbel1 (changes requested):

lib/src/deserializer.dart
desmonddak: All inputs should be protected?

lib/src/deserializer.dart
desmonddak: I can add that.  Should all comments that preface classes and methods end in '.'?

mkorbel1: Yes, I think so.  I try to follow the Dart style guides, e.g.
https://dart.dev/effective-dart/documentation#do-format-comments-like-sentences

lib/src/serializer.dart
desmonddak: Can you specify the rules for this?  I am leaning toward leaving everything unprotected for now.

test/serialization_test.dart
desmonddak: I have created a test that links a deserializer->serializer->deserializer.

lib/src/serializer.dart
desmonddak: Should I eliminate the getters or the protected attributes?

lib/src/deserializer.dart
```diff
@@ -0,0 +1,69 @@
...
+    addOutputArray('deserialized',
+            dimensions: [length], elementWidth: serialized.width) <=
```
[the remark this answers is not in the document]
desmonddak: I think this reversed the output, but it was correct.  Perhaps i should have been doing rswizzle.
Still, this is important because I was starting to use the [].swizzle() idiom a lot, not being as familiar with the map approach above.

lib/src/deserializer.dart
desmonddak: This is an input.  But you mention @protected for outputs.  I'm not clear on what we want to do here.

lib/src/deserializer.dart
desmonddak: So I think I found an asymmetry.  Deserialization uses a flop by default to construct the aggregate (it could just mux latest input into the output port), but instead we are flopping.  Serialization has a conditional flop.

So the overflowed is really just a flop of equalsMax.    I think the fix is to create a deserialization that does NOT flop, and then flop conditionally the done and the aggregate array.

desmonddak: I'm also thinking:  should we eliminate conditional flopping. I mean, if someone wants to flop the outputs of these components, then the 'done' signals should be the enables for those flops -- no reason to have the components do the flopping.

desmonddak: Now that I look at it, it makes sense that deserialization must flop to accumulate the result and will necessarily take an extra cycle.

lib/src/serialization/deserializer.dart
```diff
@@ -0,0 +1,64 @@
...
+
+  /// [done] emitted when the last element is committed to [deserialized].
+  /// The timing is that you can latch [deserialized] when [done] is high.
+  Logic get done => output('done');
+
+  /// Return the current count of elements that have been serialized out.
+  Logic get count => output('count');
+
+  /// Internal storage of length of data to be serialized.
+  late final int _length;
```
mkorbel1, line 30: this does not need to be `late` since it's initialized in the constructor line.  also, you could make this public if you want and then avoid the getter above, since a `final int` could never be changed anyways.

lib/src/serialization/deserializer.dart
```diff
@@ -0,0 +1,64 @@
...
+    clk = addInput('clk', clk);
+    reset = addInput('reset', reset);
+    if (enable != null) {
+      enable = addInput('enable', enable);
+    }
+    serialized = addInput('serialized', serialized, width: serialized.width);
+    final cnt = Counter.simple(
+        clk: clk, reset: reset, enable: enable, maxValue: length - 1);
+    addOutput('count', width: cnt.width) <= cnt.count;
+    addOutput('done') <= cnt.overflowed;
```
mkorbel1, line 52: i think this `overflowed` indication (instead of `equalsMax`) works because it happens the cycle after `equalsMax` since the data output `deserialized` is delayed by one `flop` cycle.  however, what if `enable` drops at the end of the data, then the counter would not overflow?  i'm not sure if there's a bug here or not, just wondering if it's safe to use this instead of flopping `equalsMax`?  Although similarly, if you use `equalsMax`, if enable halts at the last one then you'd hold `done` high perpetually without an edge detection.  Have you tested this case where enable drops at the end?

lib/src/serialization/serializer.dart
```diff
@@ -0,0 +1,84 @@
...
+  /// [Serializer] is configured to latch the input data and hold it until
+  /// [done] is asserted after the full `LogicArray` [deserialized] is
+  /// transferred. This will delay the serialized output by one cycle.
+  Serializer(LogicArray deserialized,
+      {required Logic clk,
+      required Logic reset,
+      Logic? enable,
+      bool flopInput = false,
+      super.name = 'Serializer'}) {
+    if (deserialized.dimensions.length != 1) {
```
mkorbel1, line 42: sorry for the flip-flop (lol), but upon reflection, i guess you could have an N dimensional array and just serialize it out by the first dimension.  that seems fair to me?

lib/src/serialization/deserializer.dart
```diff
@@ -0,0 +1,64 @@
...
+  /// and aggregates it into one wide output [deserialized] of length [length].
+  ///
+  /// Updates one element per clock while [enable] (if connected) is high,
+  /// emitting [done] when completing the filling of wide output `LogicArray`
+  /// [deserialized].
+  Deserializer(Logic serialized, this._length,
+      {required Logic clk,
+      required Logic reset,
+      Logic? enable,
+      super.name = 'Deserializer'}) {
```
mkorbel1, line 42: the `name` is used as a default instance name, so typically we write it in lower-case.  As written, this could be a little confusing for generated verilog since `Deserializer Deserializer(.clk(...` is actually legal SystemVerilog (same name for instance and definition).  I recommend changing default to `deserializer` in lower-case.  (same for the serializer)

mkorbel1 (approved):
Looks great!

lib/src/serialization/serializer.dart
```diff
@@ -72,7 +82,11 @@ class Serializer extends Module {
                   clk, reset: reset, en: latchInput, deserialized.elements[i])
               : deserialized.elements[i]);
     }
-    serialized <= dataOutput.elements.selectIndex(count);
+    if (deserialized.dimensions.length > 1) {
```
mkorbel1, line 85: aren't these two statements identical?  why have the `if` at all?

mkorbel1 (changes requested):

lib/src/serialization/serializer.dart
desmonddak: Good catch.  I had moved the addOutputArray down there while experimenting, only to discover that I could use the same name for the output after constructing Logic or LogicArray.

mkorbel1 (approved):
great!

mkorbel1 (approved):

PR #93, opened 2024-09-17
Extrema Component

robtorx:
## Description & Motivation

The component extrema finds min or max of a list of Logic and returns the value and index of the corresponding element.
Has functionality for finding first or last instance of the extrema if multiple of the same exist.
Includes extrema_test file, and config files. 

## Related Issue(s)

N/A

## Testing

Testing includes an extrema_test file with various cases. All rohd-hcl tests passed.

## Backwards-compatibility

No.

## Documentation

Documentation added to doc/components/.  Local API docs generated.


lib/src/extrema.dart
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    // Initializes extremaVal with value from index 0 of toCompareInternal.
+    var extremaVal = toCompareInternal[0];
+
+    // Initializes extremaIndex with index to 0.
+    Logic extremaIndex = Const(0, width: log2Ceil(toCompareInternal.length));
+
+    // If max is true, find max value. Else, find min value.
+    if (max) {
```
mkorbel1, line 66: you could move the condition for max vs. min inside to reduce code duplication.  for example, `final compareVal = max ? ...gt... : ...lt...`

lib/src/extrema.dart
```diff
@@ -0,0 +1,90 @@
...
+      for (var i = 1; i < toCompareInternal.length; i++) {
+        final compareVal = toCompareInternal[i].gt(extremaVal);
+        extremaVal = Logic(name: 'myName$i', width: extremaVal.width)
+          ..gets(mux(compareVal, toCompareInternal[i], extremaVal));
+        extremaIndex = mux(compareVal,
+            Const(i, width: log2Ceil(toCompareInternal.length)), extremaIndex);
```
mkorbel1, line 72: you could calculate the indexWidth once above and reuse to make this more readable

test/extrema_test.dart
```diff
@@ -0,0 +1,112 @@
...
+    expect(() => findMax.val.value.toInt(), throwsA(isA<RohdException>()));
+    expect(findMax.index.value.toInt(), equals(0));
+  });
+
+  test('List containing no elements.', () async {
+    // Create a list of Logic objects with
+    List<Logic> logics = [];
+
+    // Try to create an instance of Extrema
+    expect(() => Extrema(logics), throwsA(isA<ArgumentError>()));
```
mkorbel1, line 110: good test, but we should throw a more helpful error message like a `RohdHclException` explaining that you can't have an empty list

test/extrema_test.dart
```diff
@@ -0,0 +1,112 @@
...
+//
+
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+import 'package:test/test.dart';
+
+void main() {
+  tearDown(() async {
+    await Simulator.reset();
+  });
```
mkorbel1, line 18: some additional cases to consider:
- if the max/min is in there multiple times, does it correctly find the first one?
- is it able to correctly find the max/min when it is the first/last element in the list?

lib/src/component_config/components/config_extrema.dart
```diff
@@ -0,0 +1,45 @@
...
+
+  /// A knob controlling the width of each element to sort.
+  final IntConfigKnob logicWidthKnob = IntConfigKnob(value: 8);
+
+  /// A knob controlling whether to find Max or Min.
+  final ToggleConfigKnob maxKnob = ToggleConfigKnob(value: true);
+
+  @override
+  late final Map<String, ConfigKnob<dynamic>> knobs = UnmodifiableMapView({
+    'max': maxKnob,
```
mkorbel1, line 29: you'll want to expose all three knobs here so that they show up properly in the configurator and web app

mkorbel1 (changes requested):
Really great start!  Nice work!  I commented some suggestions and changes

doc/components/extrema.md
mkorbel1, line 1: you can use the markdown extension suggested in the repo to auto-format this document to resolve errors

doc/components/extrema.md
```diff
@@ -0,0 +1,73 @@
+# Extrema
+
+ROHD-HCL provides a component to find the extrema of a list of Logic.
+
+The component `Extrema` will determine the extremas (maximum or minimum) and their position in a list of [`Logic`]s.
```
mkorbel1, line 5: In markdown files, you don't need to do the `[]` around things since it's not connected to `dart doc` linking capabilities.  Just wrapping with ticks "`" is enough. (remove the brackets)

lib/src/extrema.dart
```diff
@@ -0,0 +1,96 @@
...
+
+  /// Finds an extrema of List<[Logic]> [toCompare]. Inputs need not be the same
+  /// width, and will all be considered positive unsigned numbers.
+  ///
+  /// If [max] is `true`, will find maximum value, else will find minimum.
+  /// If [first] is `true`, will find first extrema, else will find last.
+  ///
+  /// Outputs the [index] and [val] of the extrema in the list of [toCompare].
+
+  Extrema(List<Logic> toCompare, {bool max = true, bool first = true}) {
```
mkorbel1, line 31: i think the `first` argument doesn't really make sense here, since effictively `max` and `first` are controlling the same thing (min vs max).  Addiitonal useful configurability would be something like `n` where you can find the "n'th" biggest or smallest.

lib/src/extrema.dart
```diff
@@ -0,0 +1,96 @@
...
+  Logic get val => output('val');
+
+  /// Finds an extrema of List<[Logic]> [toCompare]. Inputs need not be the same
+  /// width, and will all be considered positive unsigned numbers.
+  ///
+  /// If [max] is `true`, will find maximum value, else will find minimum.
+  /// If [first] is `true`, will find first extrema, else will find last.
+  ///
+  /// Outputs the [index] and [val] of the extrema in the list of [toCompare].
+
```
mkorbel1, line 30: don't leave empty space between the doc comment and function (check multiple places for this)

doc/components/extrema.md
```diff
@@ -0,0 +1,73 @@
...
+`Extrema` will then ouput two `Logic` signals:
+* `index` The index or position of the extrema in the list
+  >`index` will be a `Logic` with the smallest width needed to represent the largest possible index in the list.
+
+* `val` The value of the extrema.
+  >`val` will be a `Logic` with a width equal to the largest width of any element in the list.
+
+The `List<Logic>` may contain `Logic`s of any width. They will all be considered positive unsigned numbers. 
+
+`Extrema` will throw an exception if the `Logic` list is empty or has an empty element.
```
mkorbel1, line 24: what does an "empty element" mean?

doc/components/extrema.md
```diff
@@ -0,0 +1,73 @@
...
+    // Create an Extrema module to find the first maximum value.
+    final firstMax = Extrema(signals);
+    await firstMax.build();
+
+    // Assign the integer representation of the value of the index to a variable.
+    final x = firstMax.index.value.toInt();
+    print('x equals firstMax index: $x');
+
+    // print the index and value of the last minimum
+    print('lastMin index: ${lastMin.index}');
```
mkorbel1, line 51: these two lines will print `Logic`s, rather than values, so aren't really that interesting

doc/components/extrema.md
mkorbel1, line 1: also update `doc/README.md` component list to point to this doc

test/extrema_test.dart
```diff
@@ -0,0 +1,217 @@
...
+  test('List containing an empty element.', () async {
+    // Create a list of Logic objects with different values
+    final logics = [Logic(width: 4)]; // Empty element
+
+    // Create an instance of FindMax
+    final findMax = Extrema(logics);
+    await findMax.build();
+
+    // Verify the minimum value and index
+    expect(() => findMax.val.value.toInt(), throwsA(isA<RohdException>()));
```
mkorbel1, line 206: this test is not interesting because the exception is related to converting an illegal value (Z or X) into an integer, not something related to your component.  It's implicit that invalid inputs to your module may produce invalid outputs.

mkorbel1 (changes requested):

lib/src/extrema.dart
robtorx: The idea here was in the case of multiple instances of the same max or min, one could choose to return the index of the first instance or the last.

doc/components/extrema.md
robtorx: This was in reference to the test and exception thrown for a logic of some width but with no declared LogicValue.

lib/src/extrema.dart
mkorbel1: oh i see, i think it's at least confusing, but also probably not necessary.  i'd remove it

doc/components/extrema.md
```diff
@@ -0,0 +1,71 @@
...
+`Extrema` will take in an input containing a `List<Logic>` along with parameters for `max` and `first`.
+
+* `max` is a boolean indicating whether to find the maximum value (`true`), or the minimum value (`false`). Default is `true`.
+
+* `first` is a boolean indicating whether to find the first instance of the extrema (`true`), or the last instance (`false`). Default is `true`.
+
+`Extrema` will then ouput two `Logic` signals:
+
+* `index` The index or position of the extrema in the list
+  >`index` will be a `Logic` with the smallest width needed to represent the largest possible index in the list.
```
mkorbel1, line 18: i think you can just do `  *` (with some spaces before the `*`) to make an indented bullet.  the `>` is a quote

lib/src/extrema.dart
```diff
@@ -0,0 +1,79 @@
...
+    // Find indexWidth, initialize extremaIndex and extremaVal.
+    final indexWidth = log2Ceil(logics.length);
+    Logic extremaIndex = Const(0, width: indexWidth);
+    var extremaVal = logics[0];
+
+    // If max is true, find max value. Else, find min value.
+    for (var i = 1; i < logics.length; i++) {
+      final compareVal =
+          max ? logics[i].gt(extremaVal) : logics[i].lt(extremaVal);
+      extremaVal = Logic(name: 'myName$i', width: maxWidth)
```
mkorbel1, line 67: let's fix the name so it's something more interesting than "myName"

lib/src/component_config/components/config_extrema.dart
```diff
@@ -0,0 +1,47 @@
...
+  /// A knob controlling the width of each element to sort.
+  final IntConfigKnob logicWidthKnob = IntConfigKnob(value: 8);
+
+  /// A knob controlling whether to find Max or Min.
+  final ToggleConfigKnob maxKnob = ToggleConfigKnob(value: true);
+
+  @override
+  late final Map<String, ConfigKnob<dynamic>> knobs = UnmodifiableMapView({
+    'Length of list (Number of elements)': toCompareKnob,
+    'Width of list (For all elements)': logicWidthKnob,
```
mkorbel1, line 30: maybe "Element width" is better than "width of list"?

mkorbel1 (changes requested):
Very close! couple minor things

mkorbel1 (approved):
Looks great! Nice work!

PR #97, opened 2024-10-01
Floatingpoint

desmonddak:
## Description & Motivation

This PR adds FloatingPoint and FloatingPointValue datatypes as well as a toy FloatingPointAdder.

## Related Issue(s)

None

## Testing

FloatingPointValue has been tested with round-trip testing with the Dart double. The FloatingPointAdder has been subjected to randomized testing with a loose comparison to native Double operations as it has no rounding mode.

## Backwards-compatibility

No

## Documentation

Yes.  A simple introductory documentation markdown page is included in this PR.


doc/components/floating_point.md
```diff
@@ -21,3 +21,7 @@ The `FloatingPoint` type is a `LogicStructure` which comprises the `Logic` bits
 Again, like `FloatingPointValue`, `FloatingPoint64` and `FloatingPoint32` subclasses are provided as these are the most common floating-point number types.
 
 ## FloatingPointAdder
+
+A very basic `FloatingPointAdder` component is available which does not perform any rounding. It takes two `FloatingPoint` `LogicStructure`s and adds them, returning a normalized `FloatingPointValue` on the output.  An option on input is the type of `ParallelPrefixTree` used in the internal addition of the mantissas.
```
mkorbel1, line 25: > normalized `FloatingPointValue` on the output.

Should be `FloatingPoint`?

lib/src/arithmetic/arithmetic_utils.dart
```diff
@@ -0,0 +1,112 @@
...
+/// Helper evaluation methods for printing aligned arithmetic bitvectors.
+extension NumericVector on LogicValue {
+  /// Print aligned bitvector with an optional header.
+  /// [name] is printed at the LHS of the line, trimmed by [prefix].
+  /// [prefix] is the distance from the margin bebore the vector is printed.
+  /// You can align with longer bitvectors by stating the length [align].
+  /// [lowLimit] will trim the vector below this bit position.
+  /// You can insert a separator [sepChar] at position [sep].
+  /// A header can be printed by setting [header] to true.
+  /// Markdown format can be produced by setting [markDown] to true.
```
mkorbel1, line 24: Might be nice to include an example of what this looks like in the doc comment?  You can wrap it in "```" to display it as code

lib/src/arithmetic/arithmetic_utils.dart
```diff
@@ -0,0 +1,112 @@
...
+      bool markDown = false}) {
+    final str = StringBuffer();
+    // ignore: cascade_invocations
+    if (header) {
+      str.write(markDown ? '|Name' : ' ' * prefix);
+
+      for (var col = ((align ?? width) - width) + width - 1;
+          col >= lowLimit;
+          col--) {
+        final bits = col > 9 ? 2 : 1;
```
mkorbel1, line 41: log(col)?  what about >=100 bits?

lib/src/arithmetic/floating_point/floating_point_adder.dart
```diff
@@ -0,0 +1,107 @@
+// Copyright (C) 2024 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// floating_point_test.dart
```
mkorbel1, line 4: file name & description are wrong

lib/src/arithmetic/floating_point/floating_point_value.dart
```diff
@@ -0,0 +1,900 @@
+// Copyright (C) 2024 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// floating_point.dart
```
mkorbel1, line 4: floating_point_value.dart

lib/src/arithmetic/floating_point/floating_point_value.dart
```diff
@@ -0,0 +1,900 @@
...
+  /// Round to nearest, tieas away from zero
+  roundNearestTiesAway,
+
+  /// Round toward zero
+  roundTowardsZero,
+
+  /// Round toward +infinity
+  roundTowardsInfinity,
+
+  /// Round toward zero
```
mkorbel1, line 71: comment wrong?

lib/src/arithmetic/floating_point/floating_point_value.dart
```diff
@@ -0,0 +1,900 @@
...
+    final m = rv.nextLogicValue(width: mantissaWidth).toBigInt();
+    return FloatingPointValue(
+        sign: LogicValue.ofInt(s, 1),
+        exponent: LogicValue.ofBigInt(e, exponentWidth),
+        mantissa: LogicValue.ofBigInt(m, mantissaWidth));
+  }
+
+  /// Convert a floating point number into a [FloatingPointValue]
+  /// representation. This form performs NO ROUNDING.
+  factory FloatingPointValue.fromDoubleIter(double inDouble,
```
mkorbel1, line 385: what is "iter"?

lib/src/arithmetic/floating_point/floating_point_value.dart
```diff
@@ -0,0 +1,900 @@
...
+      LogicValue.ofBigInt(mantissa, mantissaWidth)
+    );
+    return FloatingPoint64Value(
+        sign: signLv, exponent: exponentLv, mantissa: mantissaLv);
+  }
+}
+
+/// A representation of a 8-bit floating point value as defined in
+/// [FP8 Formats for Deep Learning](https://arxiv.org/abs/2209.05433).
+class FloatingPoint8Value extends FloatingPointValue {
```
mkorbel1, line 805: It feels a little odd to me to have one "FP8" type that supports different mixes of exponent and mantissa width.  Why not make them separate types?

mkorbel1 (commented):
Looking really good!

doc/components/floating_point.md
```diff
@@ -6,13 +6,16 @@ Floating-point operations require meticulous precision, and have standards like
 
 The `FloatingPointValue` class comprises the sign, exponent, and mantissa `LogicValue`s that represent a floating-point number. `FloatingPointValue`s can be converted to and from Dart native `Double`s, as well as constructed from integer and string representations of their fields.  They can be operated on (+, -, *, /) and compared.
 
-The various IEEE constants representing corner cases of the field of floating-point values for a given size of `FloatingPointValue`: infinities, zeros, limits for normal (e.g. mantissa in the range of [1,2]) and sub-normal numbers (zero exponent, and mantissa <1).
+A `FloatingPointValue` has a mantissa in $[0,2)$ with $$0 <= exponent <= `maxExponent`$$ A normal `isNormal` `FloatingPointValue`
```
mkorbel1, line 9: what are these `$`'s about?  looking at it in markdown doesn't fully clarify
https://github.com/intel/rohd-hcl/blob/edc71c6569724e99fa8aa9c40968ef75bc95df77/doc/components/floating_point.md

lib/src/arithmetic/floating_point/floating_point_value.dart
desmonddak: This is a converter that iterates in order to convert from double to avoid an overflow.
This routine is general as it does not need to know the format of the native double.
.fromDouble() relies on the exact bit representation of the native double in doing the conversion.

We don't absolutely have to have this routine, but it took a long time to get it to work correctly.

doc/components/floating_point.md
desmonddak: I looked up github markdown for equations.   Inline is bracketing by a single Dollar sign.  Equation is using a double Dollar sign before and after. I probably need to remove the quotes as they do not work in equations.
image

lib/src/arithmetic/arithmetic_utils.dart
desmonddak: I was wondering what to do about these debug routines that I use.  I call them 'evaluate', but they are really introspection into the component or printing live values.  Where should we put these in both code and documentation?
I put some of them in the multiplier_components.md, for example.  I can put an example there.

lib/src/arithmetic/floating_point/floating_point_value.dart
mkorbel1: It's a good thing to keep, but I'm wondering if there's a better name for this that describes the functionality instead of the implementation?

lib/src/arithmetic/arithmetic_utils.dart
desmonddak: The format is currently limited because of the way I print headers.
Should I fix, or throw?  Fixing will slow the PR and it is a debug routine.

I also want to merge the production of ASCII and markdown so that the logic is not replicated.  I did this in my most
recent vecString routine.

doc/components/floating_point.md
mkorbel1: oh wow that looks really cool on yours!  why does it look like this to me in github?
image

perhaps there's a way to make this work nicely in github's markdown parsing as well?

lib/src/arithmetic/floating_point/floating_point_value.dart
desmonddak: I wonder if these should be in the same file or separate files.  FP32, FP64, and then the myriad of 'standard' short formats should have separate files?

I can work with Soner on naming.

lib/src/arithmetic/arithmetic_utils.dart
mkorbel1: Sure that works too, to clarify my suggestion it was something like this:
https://github.com/intel/rohd/blob/3e2d24a78adc7579fb7d4a99452d2105a2e70b65/lib/src/signals/logic.dart#L650-L676

Which then looks like this in generated API docs:
https://intel.github.io/rohd/rohd/Logic/getRange.html

lib/src/arithmetic/arithmetic_utils.dart
mkorbel1: since it's for debug it's probably fine to leave as-is.  throwing an exception might be nice, or even just adding a comment that it "only works for <100 bits" or something would work too

lib/src/arithmetic/floating_point/floating_point_value.dart
mkorbel1: separate files seems better to me, I agree.  perhaps we should meet and talk about in general how to easily add many different types without needing to do much code duplication

lib/src/arithmetic/arithmetic_utils.dart
desmonddak: I need to do this much more often.  There are many cases an invocation example is the best documentation.  I put something in the documentation, but I will work toward this example.

doc/components/floating_point.md
desmonddak: Let me figure out why this is.  Github is supposed to support this format feature.

desmonddak: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/writing-mathematical-expressions

lib/src/arithmetic/arithmetic_utils.dart
desmonddak: I decided to fix.  I want to use this routine in ALL my evaluate routines because it is something I do all the time.
That will give me Markdown and potentially HTML forms of these printouts through reuse.  I have two other routines that print:  the partial product generator  and the compression tree:  we will be looking at both of these to debug VMM
core.

lib/src/arithmetic/floating_point/floating_point_value.dart
desmonddak: We will have to first move away from factory constructors and then get Soner to factor his stuff into multiple files during his PR.

doc/components/floating_point.md
desmonddak: This appears correctly on github:
image

doc/README.md
```diff
@@ -59,7 +59,7 @@ Some in-development items will have opened issues, as well. Feel free to create
     - BFloat16 (16-bit)
     - BFloat8 (8-bit)
     - BFloat4 (4-bit)
-  - [Simple Floating-Point Adder](./componeents/floating_point.md#floatingpointadder)
+  <!-- - [Simple Floating-Point Adder](./componeents/floating_point.md#floatingpointadder) -->
```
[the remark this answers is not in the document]
desmonddak: This isn't the problem.  It is checked in I just cannot link to the document as it fails the link test.

doc/components/floating_point.md
mkorbel1: indeed it does! nice!

doc/README.md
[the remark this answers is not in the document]
mkorbel1: oh, typo "componeents"?

mkorbel1 (approved):
I think this looks ready!

PR #99, opened 2024-10-04
Fixed Point Value system

soneryaldiz:
## Description & Motivation

Introduce a value system to represent unsigned or signed fixed-point values following Q notation.

## Testing

All constructors, methods and operators have unit tests.

## Backwards-compatibility

Not a braking change.

## Documentation

work in progress.

### DIFF
--- a/lib/src/arithmetic/values/fixed_point_value.dart
+++ b/lib/src/arithmetic/values/fixed_point_value.dart
@@ -1,0 +1,167 @@
+// Copyright (C) 2024 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// fixed_point_value.dart
+//
+// 2024 September 24
+// Authors:
+//  Soner Yaldiz <soner.yaldiz@intel.com>
+
+import 'dart:math';
+import 'package:meta/meta.dart';
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+
+/// An immutable representation of (un)signed fixed-point values following
+/// Q notation (Qm.n format) as introduced by
+/// (Texas Instruments)[https://www.ti.com/lit/ug/spru565b/spru565b.pdf].
+@immutable
+class FixedPointValue implements Comparable<FixedPointValue> {
+  /// The full fixed point value bit storage in two's complement.
+  late final LogicValue value;
+
+  /// The sign of the fixed point number.
+  final LogicValue sign;
+
+  /// The integer part of the fixed-point number.
+  final LogicValue integer;
+
+  /// The fractional part of the fixed-point number.
+  final LogicValue fraction;
+
+  /// Constructs [FixedPointValue] from sign, integer and fraction values.
+  FixedPointValue(
+      {this.sign = LogicValue.empty,
+      this.integer = LogicValue.empty,
+      this.fraction = LogicValue.empty}) {
+    if ((integer == LogicValue.empty) & (fraction == LogicValue.empty)) {
+      throw RohdHclException('integer or fraction must be non-empty');
+    }
+    if ((sign == LogicValue.empty) | (sign.isZero)) {
+      value = [sign, integer, fraction].swizzle();
+    } else {
+      value = ~[LogicValue.zero, integer, fraction].swizzle() + 1;
+    }
+  }
+
+  /// Constructs [FixedPointValue] of a Dart [double] rounding away from zero.
+  factory FixedPointValue.ofDouble(double value,
+      {required bool signed, required int m, required int n}) {
+    if (value.abs().floor() > pow(2, m) - 1) {
+      throw RohdHclException('value exceed integer part');
+    }
+    if ((!signed) & (value < 0)) {
+      throw RohdHclException('Negative value must be signed.');
+    }
+    final sign = value >= 0 ? LogicValue.zero : LogicValue.one;
+    final integerPart = value.abs().floor();
+    final fractionalPart = ((value.abs() - integerPart) * pow(2, n)).round();
+
+    return FixedPointValue(
+        sign: signed ? sign : LogicValue.empty,
+        integer: LogicValue.ofInt(integerPart, m),
+        fraction: LogicValue.ofInt(fractionalPart, n));
+  }
+
+  /// Returns the value of the fixed-point number in a Dart [double] type.
+  double toDouble() {
+    final value = integer.toInt().toDouble() +
+        (fraction.toInt().toDouble() / pow(2, fraction.width));
+    return (sign == LogicValue.empty) | (sign == LogicValue.zero)
+        ? value
+        : -value;
+  }
+
+  /// Returns a negative integer if `this` less than [other],
+  /// a positive integer if `this` greater than [other],
+  /// and zero if `this` and [other] are equal.
+  @override
+  int compareTo(Object other) {
+    if (other is! FixedPointValue) {
+      throw RohdHclException('Input must be of type FixedPointValue');
+    }
+    final thisValue = toDouble();
+    final otherValue = other.toDouble();
+    if (thisValue == otherValue) {
+      return 0;
+    } else if (thisValue < otherValue) {
+      return -1;
+    } else {
+      return 1;
+    }
+  }
+
+  /// Equal-to operation that returns a LogicValue.
+  LogicValue eq(FixedPointValue other) =>
+      compareTo(other) == 0 ? LogicValue.one : LogicValue.zero;
+
+  /// Not equal-to operation that returns a LogicValue.
+  LogicValue neq(FixedPointValue other) =>
+      compareTo(other) != 0 ? LogicValue.one : LogicValue.zero;
+
+  /// Less-than operation that returns a LogicValue.
+  LogicValue operator <(FixedPointValue other) =>
+      compareTo(other) < 0 ? LogicValue.one : LogicValue.zero;
+
+  /// Less-than operation that returns a LogicValue.
+  LogicValue operator <=(FixedPointValue other) =>
+      compareTo(other) <= 0 ? LogicValue.one : LogicValue.zero;
+
+  /// Less-than operation that returns a LogicValue.
+  LogicValue operator >(FixedPointValue other) =>
+      compareTo(other) > 0 ? LogicValue.one : LogicValue.zero;
+
+  /// Less-than operation that returns a LogicValue.
+  LogicValue operator >=(FixedPointValue other) =>
+      compareTo(other) >= 0 ? LogicValue.one : LogicValue.zero;
+
+  /// Addition operation that returns a FixedPointValue.
+  /// The result is signed if one of the operands is signed.
+  /// The result integer has the max integer width of the operands plus one.
+  /// The result fraction has the max fractional width of the operands.
+  FixedPointValue operator +(FixedPointValue other) {
+    final res = toDouble() + other.toDouble();
+    final signed =
+        (sign != LogicValue.empty) | (other.sign != LogicValue.empty);
+    final m = max(integer.width, other.integer.width) + 1;
+    final n = max(fraction.width, other.fraction.width);
+    return FixedPointValue.ofDouble(res, signed: signed, m: m, n: n);
+  }
+
+  /// Subtraction operation that returns a FixedPointValue.
+  /// The result is always signed.
+  /// The result integer has the max integer width of the operands plus one.
+  /// The result fraction has the max fractional width of the operands.
+  FixedPointValue operator -(FixedPointValue other) {
+    final res = toDouble() - other.toDouble();
+    final m = max(integer.width, other.integer.width) + 1;
+    final n = max(fraction.width, other.fraction.width);
+    return FixedPointValue.ofDouble(res, signed: true, m: m, n: n);
+  }
+
+  /// Multiplication operation that returns a FixedPointValue.
+  /// The result is signed if one of the operands is signed.
+  /// The result integer width is the sum of integer widths of operands.
+  /// The result fraction width is the sum of fraction widths of operands.
+  FixedPointValue operator *(FixedPointValue other) {
+    final signed =
+        (sign != LogicValue.empty) | (other.sign != LogicValue.empty);
+    final res = toDouble() * other.toDouble();
+    final m = integer.width + other.integer.width;
+    final n = fraction.width + other.fraction.width;
+    return FixedPointValue.ofDouble(res, signed: signed, m: m, n: n);
+  }
+
+  /// Division operation that returns a FixedPointValue.
+  /// The result is signed if one of the operands is signed.
+  /// The result integer width is the sum of integer widths of operands.
+  /// The result fraction width is the sum of fraction widths of operands.
+  FixedPointValue operator /(FixedPointValue other) {
+    final signed =
+        (sign != LogicValue.empty) | (other.sign != LogicValue.empty);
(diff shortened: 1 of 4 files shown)


lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -0,0 +1,167 @@
...
+      {this.sign = LogicValue.empty,
+      this.integer = LogicValue.empty,
+      this.fraction = LogicValue.empty}) {
+    if ((integer == LogicValue.empty) & (fraction == LogicValue.empty)) {
+      throw RohdHclException('integer or fraction must be non-empty');
+    }
+    if ((sign == LogicValue.empty) | (sign.isZero)) {
+      value = [sign, integer, fraction].swizzle();
+    } else {
+      value = ~[LogicValue.zero, integer, fraction].swizzle() + 1;
```
mkorbel1, line 43: doesn't this leave the `integer` and `fraction` with incorrect values that don't match `value`?

lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -0,0 +1,167 @@
...
+    final fractionalPart = ((value.abs() - integerPart) * pow(2, n)).round();
+
+    return FixedPointValue(
+        sign: signed ? sign : LogicValue.empty,
+        integer: LogicValue.ofInt(integerPart, m),
+        fraction: LogicValue.ofInt(fractionalPart, n));
+  }
+
+  /// Returns the value of the fixed-point number in a Dart [double] type.
+  double toDouble() {
```
mkorbel1, line 67: what if it doesn't fit?  should we add exceptions? describe how rounding would occur?

lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -0,0 +1,167 @@
...
+    final value = integer.toInt().toDouble() +
+        (fraction.toInt().toDouble() / pow(2, fraction.width));
+    return (sign == LogicValue.empty) | (sign == LogicValue.zero)
+        ? value
+        : -value;
+  }
+
```
mkorbel1, line 74: should we add a `toFloatingPointValue`?

lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -0,0 +1,167 @@
...
+      return 0;
+    } else if (thisValue < otherValue) {
+      return -1;
+    } else {
+      return 1;
+    }
+  }
+
+  /// Equal-to operation that returns a LogicValue.
+  LogicValue eq(FixedPointValue other) =>
```
mkorbel1, line 95: If we want to stick with the APIs of `LogicValue`, we'd have `operator ==` and `hashCode` overridden which do direct bit comparisons, wheras `eq` and `neq` could potentially have `x`s generated (if inputs have some `x`s).

For example:

`eq` and `neq`
https://github.com/intel/rohd/blob/3e2d24a78adc7579fb7d4a99452d2105a2e70b65/lib/src/values/logic_value.dart#L1062-L1076

`==`
https://github.com/intel/rohd/blob/3e2d24a78adc7579fb7d4a99452d2105a2e70b65/lib/src/values/logic_value.dart#L541-L553

lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -0,0 +1,167 @@
...
+  /// a positive integer if `this` greater than [other],
+  /// and zero if `this` and [other] are equal.
+  @override
+  int compareTo(Object other) {
+    if (other is! FixedPointValue) {
+      throw RohdHclException('Input must be of type FixedPointValue');
+    }
+    final thisValue = toDouble();
+    final otherValue = other.toDouble();
+    if (thisValue == otherValue) {
```
mkorbel1, line 85: could do `toDouble().compareTo(other.toDouble()`?

Although, also, this only scales up to numbers that fit (with enough precision) in a Dart `double`.  Should we implement a comparison using `LogicValue` instead (and considering sign, fraction, etc.) so that it works for any `FixedPoint` instead of only ones reliably representable as a `double`?

lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -0,0 +1,167 @@
...
+  /// Less-than operation that returns a LogicValue.
+  LogicValue operator >=(FixedPointValue other) =>
+      compareTo(other) >= 0 ? LogicValue.one : LogicValue.zero;
+
+  /// Addition operation that returns a FixedPointValue.
+  /// The result is signed if one of the operands is signed.
+  /// The result integer has the max integer width of the operands plus one.
+  /// The result fraction has the max fractional width of the operands.
+  FixedPointValue operator +(FixedPointValue other) {
+    final res = toDouble() + other.toDouble();
```
mkorbel1, line 123: doesn't a standard `LogicValue + LogicValue` (with sufficient extension/padding) give the same operation but support any arbitrary size and precision?  (similar comment for other operators)

test/arithmetic/values/fixed_point_value_test.dart
```diff
@@ -0,0 +1,244 @@
...
+// Authors:
+//  Soner Yaldiz <soner.yaldiz@intel.com>
+
+import 'dart:math';
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+import 'package:test/test.dart';
+
+void main() {
+  test('FXP: construct from sign,int,frac', () {
```
mkorbel1, line 14: nit: unnecessary to prefix test names with "FXP", since they're all in a fixed point test anyways.  If you do want to prefix, a `group` can do it.

test/arithmetic/values/fixed_point_value_test.dart
```diff
@@ -0,0 +1,244 @@
+// Copyright (C) 2024 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+//
```
mkorbel1, line 3: file name and description in header?

lib/src/arithmetic/values/fixed_point_value.dart
soneryaldiz: `value` is always in two's complement where `integer` and `fraction` are not. Accessing `fractional` or `integer` in two's complement fashion is pointless. Assuming that `FloatingPointValue` objects will mostly be constructed using `ofDouble`, this class can be refactored as follows:
```dart
class FixedPointValue implements Comparable<FixedPointValue> {
  /// The full fixed point value bit storage in two's complement.
  late final LogicValue value;
  /// [signed] indicates whether the representation is signed.
  final bool signed;
  /// [m] is the number of bits reserved for the integer part.
  final int m;
  /// [n] is the number of bits reserved for the fractional part.
  final int n;
```
where value is always in two's complement. What do you think?

lib/src/arithmetic/values/fixed_point_value.dart
soneryaldiz: I will add an exception. As long as `m+n <= 52`, rounding is not required.

lib/src/arithmetic/values/fixed_point_value.dart
soneryaldiz: This creates a dependency between the two value systems. One could rather use `ofDouble` and `toDouble` methods to do conversions as needed.

soneryaldiz:
@mkorbel1 , I will refactor this implementation based on your comments:

- [ ] Refactor class members to eliminate integer and fraction
- [ ] Refactor operators to allow arbitrary integer and fraction width and to re-use `LogicValue` operators
- [ ] Add exceptions converting to Dart double (no saturation nor rounding)

lib/src/arithmetic/values/fixed_point_value.dart
mkorbel1: is it not also useful to provide an accessor for the integer and fraction portion of the `value`?  i.e. yes, this looks good, but maybe also add something that `getRange`s the integer and fraction portion?

lib/src/arithmetic/values/fixed_point_value.dart
mkorbel1: I think it's an ok dependency to have since they are in the same library, and it adds support for things that don't fit inside a standard `double`.

Another thing to (maybe?) consider is that Dart can be converted to JavaScript, which has weird characteristics. (https://dart.dev/guides/language/numbers).  The more you lean on `LogicValue`, the more you get for "free", since we've done *extensive* testing to make sure it works well.

lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -2,7 +2,8 @@
 // SPDX-License-Identifier: BSD-3-Clause
 //
 // fixed_point_value.dart
-//
+// Representation of fixed-point values.
+
```
mkorbel1, line 6: nit: shouldn't have a blank line here

lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -0,0 +1,211 @@
...
+  final bool signed;
+
+  /// [m] is the number of bits reserved for the integer part.
+  final int m;
+
+  /// [n] is the number of bits reserved for the fractional part.
+  final int n;
+
+  /// Returns true if the number is negative.
+  bool isNegative() => signed & (value.getRange(-1) == LogicValue.one);
```
mkorbel1, line 34: this works, another option is `signed & value[-1].toBool()`

lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -0,0 +1,211 @@
...
+
+  /// Returns true if the number is positive.
+  bool isPositive() => !isNegative();
```
mkorbel1, line 37: is zero positive?

lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -0,0 +1,211 @@
...
+    final w = signed ? m + n + 1 : m + n;
+    if (w != value.width) {
+      throw RohdHclException('Width must be (sign) + m + n.');
+    }
+  }
+
+  /// Expands the bit width of integer and fractional parts.
+  LogicValue expandWidth({required bool sign, int m = 0, int n = 0}) {
+    if ((m > 0) & (m < this.m)) {
+      throw RohdHclException('Illegal use.');
```
mkorbel1, line 57: these exceptions could be more descriptive

lib/src/arithmetic/values/fixed_point_value.dart
mkorbel1: In general, consider what we should do for "invalid" cases (i.e. values that contain X and Z).  Do we want to just disallow them and throw exceptions everywhere?  Or do we want to handle them somewhat and return a value with X's?

mkorbel1 (commented):
changes looking good!

soneryaldiz:
@mkorbel1 , this PR is ready for final review. I will add randomized testing for operators.

lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -0,0 +1,257 @@
...
+    }
+    final w = signed ? m + n + 1 : m + n;
+    if (w != value.width) {
+      throw RohdHclException('Width must be (sign) + m + n.');
+    }
+  }
+
+  /// Expands the bit width of integer and fractional parts.
+  LogicValue expandWidth({required bool sign, int m = 0, int n = 0}) {
+    if ((m > 0) & (m < this.m)) {
```
mkorbel1, line 53: what if `m` or `n` passed in as arguments are less than 0?

lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -0,0 +1,257 @@
...
+      compareTo(other) > 0 ? LogicValue.one : LogicValue.zero;
+
+  /// Less-than operation that returns a LogicValue.
+  LogicValue operator >=(FixedPointValue other) =>
+      compareTo(other) >= 0 ? LogicValue.one : LogicValue.zero;
+
+  @override
+  int get hashCode => _hashCode;
+
+  int get _hashCode =>
```
mkorbel1, line 134: this seems unnecessary to have the private copy, you could just put `int get hashCode =>  value.hashCode ^ signed.hashCode ^ m.hashCode ^ n.hashCode;`

test/arithmetic/values/fixed_point_value_test.dart
```diff
@@ -0,0 +1,258 @@
...
+                    signed: s1 == 0,
+                    m: m1,
+                    n: n1);
+                fxp2 = FixedPointValue(
+                    value: LogicValue.ofInt(i2, w),
+                    signed: s2 == 0,
+                    m: m2,
+                    n: n2);
+
+                // print('$i1 $i2 s1=$s1 m1=$m1 n1=$n1 s2=$s2 m2=$m2 n2=$n2');
```
mkorbel1, line 201: can remove debug prints from tests

test/arithmetic/values/fixed_point_value_test.dart
```diff
@@ -0,0 +1,258 @@
...
+                    reason: '-');
+                expect(fxp.n, max(n1, n2));
+                expect(fxp.m, max(m1, m2) + 1);
+
+                // multiply
+                fxp = fxp1 * fxp2;
+                expect(fxp.toDouble(), fxp1.toDouble() * fxp2.toDouble(),
+                    reason: '${fxp1.toDouble()}*${fxp2.toDouble()}');
+                expect(fxp.n, n1 + n2);
+                // expect(fxp.m, m1 + m2 + 1);
```
mkorbel1, line 224: did you mean to replace this back?

test/arithmetic/values/fixed_point_value_test.dart
```diff
@@ -0,0 +1,258 @@
...
+                        '${fxp1.toDouble()}/${fxp2.toDouble()} = $expectedValue');
+              }
+            }
+          }
+        }
+      }
+    }
+  });
+
+  test('xxx', () {
```
mkorbel1: this test is WIP?

mkorbel1 (commented):
What do you think of the two remaining comments?

- accessors for the fraction and integer portions
- to/from `FloatingPointValue`

I don't think they are strictly necessary to add (or if we want them, necessarily in this PR), just closing out remaining comments.

test/arithmetic/values/fixed_point_value_test.dart
mkorbel1: nvm see you fixed :)

soneryaldiz:
Accessors for the fraction and integer portions: This is a bit tricky and needs to be defined clearly. Here is an example: Q0.4 does not have an integer part, yet it can represent -1. When one asks for integer portion, what should it return?

to/from `FloatingPointValue`: Given both fixed point and floating point offers to/from double, is there a value add to have dedicated methods in this class?

mkorbel1:
> Accessors for the fraction and integer portions: This is a bit tricky and needs to be defined clearly. Here is an example: Q0.4 does not have an integer part, yet it can represent -1. When one asks for integer portion, what should it return?

Ah good point, ok then!

> to/from FloatingPointValue: Given both fixed point and floating point offers to/from double, is there a value add to have dedicated methods in this class?

What about large fixed point and floating point values that don't fit in double?

soneryaldiz:
Let's revisit from/to floating point after the refactoring is completed. Resolved all other comments in the last commit.

mkorbel1 (approved):
Looks great!

PR #102, opened 2024-10-07
Enhancements for modifying and printing partial product arrays.

desmonddak:
## Description & Motivation

These are changes that enhance PartialProductGenerator for bit-level customization.
Partial product printing (including markdown format) is included.

## Related Issue(s)

None.

## Testing

Tested with differing kinds of sign extension modifications and validated for correctness.

## Backwards-compatibility

No.

## Documentation

Documentation is included in source as well as multiplier_components.md.


doc/components/multiplier_components.md
mkorbel1, line 1: I noticed the formatting looks off in the "Utility: Aligned Vector Formatting" section (not changed this PR, but maybe an easy fix to include)

https://github.com/intel/rohd-hcl/blob/437d2b4f774e49a6505438807d4787719203e119/doc/components/multiplier_components.md

test/arithmetic/multiplier_encoder_test.dart
mkorbel1, line 1: Is it possible to add some sort of basic tests for the new APIs?

desmonddak: A correctness test is pretty tough.  I was thinking if I could get my own dot-product generalization to work, this would be a way to test it, but that will take some time to get working correctly.  We do use them, just not in a test we can use here.
I'll see if there is something simple like a 2-stage dot-product I can make happen.

mkorbel1: maybe correctness is more than needed, is there an easy way to just unit test that the API itself did what it was supposed to do?  just to make sure the APIs are alive and protect them from being broken by later changes?

desmonddak: I think I can do some setAbsolute/getAbsolute round-trips, muxed and not.  
Need a bit more time.
I started down the path of a general dot-product and found an issue with ColumnCompression, but it doesn't show up elsewhere.  So something about when I don't tell it the full width, it freaks out and produces some garbage (not just clipped).

mkorbel1 (approved):
Looks great!!

test/arithmetic/multiplier_encoder_test.dart
mkorbel1: looks great!

PR #104, opened 2024-10-09
[WIP] add float conversion for arbitrary mantissa width and exponent width

saw235:


lib/src/arithmetic/floating_point/floating_point_value.dart
```diff
@@ -526,20 +526,17 @@ class FloatingPointValue implements Comparable<FloatingPointValue> {
...
-    } else {
-      return exponent.toInt() ==
-          computeMaxExponent(exponent.width) + computeBias(exponent.width) + 1;
-    }
-  }
+  bool isExponentAllOnes() =>
+      exponent.toInt() ==
+      computeMaxExponent(exponent.width) + computeBias(exponent.width) + 1;
+
+  bool isNaN() => isExponentAllOnes() && mantissa.toInt() == 0;
```
mkorbel1, line 533: instead of converting to int, check if the value is zero using `isZero` so that you can support mantissas >64 bits

lib/src/arithmetic/floating_point/floating_point_value.dart
```diff
@@ -907,4 +904,117 @@ class FloatingPoint8Value extends FloatingPointValue {
...
+///
+/// This is a binary count of the number of leading zeroes in the
+/// [LogicValue].
+///
+/// The logic value is treated as a binary number, with the rightmost
+/// bit being the least significant bit (bit 0). The leading zeroes
+/// are counted from the most significant bit (bit width-1) towards
+/// the least significant bit.
+///
+/// For example, the binary number 0b00101010 would return 3, as
```
mkorbel1, line 1010: aren't there 2?

lib/src/arithmetic/floating_point/floating_point_value.dart
```diff
@@ -907,4 +904,117 @@ class FloatingPoint8Value extends FloatingPointValue {
...
+/// there are 3 leading zeroes.
+int countLeadingZeroes(LogicValue value) {
```
mkorbel1, line 1012: if this is a good utility in a generic sense, maybe it should be an extension on `LogicValue` or even added to ROHD itself.  If it's mostly just useful for this math stuff, let's keep it private (start name with `_`) so that it's not exposed as a public API

PR #105, opened 2024-10-10
Shift Register Updates

robtorx:
## Description & Motivation

The shift register component was changed to include functionality for asynchronous reset and for individual stages of the shift register to have different reset values. 
Motivation stems from a need for this functionality in implementing SPI for ROHD.

## Related Issue(s)

N/A

## Testing

More tests were added to the existing shift register tests, testing lists of reset values and async reset.
## Backwards-compatibility

No
## Documentation

Component documentation was updated

### DIFF
--- a/test/shift_register_test.dart
+++ b/test/shift_register_test.dart
@@ -209,4 +209,113 @@ void main() {
 
     await Simulator.endSimulation();
   });
+
+  group('list reset value shift register', () {
+    Future<void> listResetTest(
+        dynamic resetVal, void Function(Logic dataOut) check) async {
+      final dataIn = Logic(width: 8);
+      final clk = SimpleClockGenerator(10).clk;
+      const depth = 5;
+      final reset = Logic();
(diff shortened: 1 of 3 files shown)


test/shift_register_test.dart
```diff
@@ -209,4 +209,113 @@ void main() {
...
+
+  group('async reset shift register', () {
+    Future<void> asyncResetTest(
+        dynamic resetVal, void Function(Logic dataOut) check) async {
+      final dataIn = Logic(width: 8);
+      final clk = SimpleClockGenerator(10).clk;
+      const depth = 5;
+      final reset = Logic();
+      final dataOut = ShiftRegister(dataIn,
+              clk: clk,
```
mkorbel1, line 273: replace `clk: clk` with `clk: Const(0)` to tie the clock off

mkorbel1 (changes requested):
Great work!

mkorbel1 (approved):
Nice job!

PR #107, opened 2024-10-12
Pp array

desmonddak:
## Description & Motivation

Exposes PartialProductArray from PartialProductGenerator to allow for array generation outside of a Booth encoder.

## Related Issue(s)

None

## Testing

Ran the existing multiplier tests.

## Backwards-compatibility

No.

## Documentation

We could add documentation for the Array as distinct from the Generator, but it isn't absolutely necessary.


lib/src/arithmetic/addend_compressor.dart
```diff
@@ -168,7 +168,7 @@ class ColumnCompressor {
   late final List<ColumnQueue> columns;
 
   /// The partial product generator to be compressed
```
mkorbel1, line 170: doc comment still says generator? is that intended?

mkorbel1 (approved):

mkorbel1 (approved):
lgtm!

PR #108, opened 2024-10-14
Adds linear feedback shift register component

omoiti:
## Description & Motivation

Added a Galois linear feedback shift register component.

## Related Issue(s)

N/A

## Testing

Added unit tests for new component. Ran unit tests individually and using dart test.

## Backwards-compatibility

No.

## Documentation

Will need to add page for LFSR implementation

### DIFF
--- a/lib/src/linear_feedback_shift_register.dart
+++ b/lib/src/linear_feedback_shift_register.dart
@@ -1,0 +1,114 @@
+// Copyright (C) 2023 Intel Corporation
+
+// SPDX-License-Identifier: BSD-3-Clause
+
+//
+
+// linear_feedback_shift_register.dart
+
+// Implementation of Galois Linear Feedback Shift Register.
+
+//
+
+// 2024 October 1
+
+// Author: Omonefe Itietie <omonefe.itietie@intel.com>
+
+//
+
+import 'dart:collection';
+import 'package:rohd/rohd.dart';
+
+/// Galois Linear Feedback Shift Register
+class LinearFeedbackShiftRegister extends Module {
+  /// Contains polynomial size for LFSR
+  final int width;
+
+  /// Contains seed of LFSR (starting value)
+  Logic state;
+
+  /// Data names for signals
+  final String dataName;
+
+  /// Contains bit string that will be used to calculate the output
+  final Logic taps;
+
+  /// Output for shift register
+  Logic get dataOut => output('${dataName}_out');
+
+  /// The number of stages in this shift register.
+  final int shifts;
+
+  /// A [List] of [output]s where the `n`'th entry corresponds to a version of
+  /// the input data after passing through `n + 1` flops.
+  late final List<Logic> stages = UnmodifiableListView(
+      [for (var i = 0; i < shifts; i++) output(_stageName(i))]);
+
+  /// The name of the signal (and output pin) for the [i]th stage.
+  String _stageName(int i) => '${dataName}_stage_$i';
+
+  LinearFeedbackShiftRegister(Logic dataIn,
+      {required Logic clk,
+      required this.state,
+      required this.shifts,
+      required this.taps,
+      Logic? enable,
+      Logic? reset,
+      dynamic resetValue,
+      this.dataName = 'data'})
+      : width = dataIn.width,
+        super(name: '${dataName}_lfsr') {
+    dataIn = addInput('${dataName}_in', dataIn, width: width);
+    clk = addInput('clk', clk);
+    addOutput('${dataName}_out', width: width);
+
+    Map<Logic, dynamic>? resetValues;
+
+    if (reset != null) {
+      reset = addInput('reset', reset);
+      if (resetValue != null) {
+        if (resetValue is Logic) {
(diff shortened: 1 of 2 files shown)


lib/src/linear_feedback_shift_register.dart
```diff
@@ -0,0 +1,114 @@
+// Copyright (C) 2023 Intel Corporation
```
mkorbel1, line 1: remove blank (uncommented) lines in the header?

lib/src/linear_feedback_shift_register.dart
```diff
@@ -0,0 +1,114 @@
...
+
+  /// A [List] of [output]s where the `n`'th entry corresponds to a version of
+  /// the input data after passing through `n + 1` flops.
+  late final List<Logic> stages = UnmodifiableListView(
+      [for (var i = 0; i < shifts; i++) output(_stageName(i))]);
+
+  /// The name of the signal (and output pin) for the [i]th stage.
+  String _stageName(int i) => '${dataName}_stage_$i';
+
+  LinearFeedbackShiftRegister(Logic dataIn,
```
mkorbel1, line 50: add a doc comment to the constructor

lib/src/linear_feedback_shift_register.dart
```diff
@@ -0,0 +1,114 @@
...
+        super(name: '${dataName}_lfsr') {
+    dataIn = addInput('${dataName}_in', dataIn, width: width);
+    clk = addInput('clk', clk);
+    addOutput('${dataName}_out', width: width);
+
+    Map<Logic, dynamic>? resetValues;
+
+    if (reset != null) {
+      reset = addInput('reset', reset);
+      if (resetValue != null) {
```
mkorbel1, line 69: the original shift register was recently upgraded to allow for a `List` `resetValue` to control each element of the register (https://github.com/intel/rohd-hcl/blob/main/lib/src/shift_register.dart).  However, since the LFSR is always 1-wide, maybe a different API makes sense.  The current `resetValue` just allows you to basically set all flops to 0 or 1.  Should we allow a bitvector of width == `shifts` to initialize the LFSR?

lib/src/linear_feedback_shift_register.dart
```diff
@@ -0,0 +1,114 @@
...
+    for (var i = 0; i < shifts; i++) {
+      final stageI = addOutput(_stageName(i), width: width);
+
+      conds.add(stageI < dataStage);
+      resetValues?[stageI] = resetValue;
+
+      Logic lsb = state.getRange(0, 1); // Get LSB (least significant bit)
+      state = [Const(0, width: 1), state.getRange(1, width)].swizzle();
+
+      If(lsb.eq(Const(1)), then: [
```
mkorbel1, line 91: This `If` (capitalized) creates a `Conditional` object which only executes under certain conditions (e.g. within a `Sequential` or `Combinational`), but the object is not included in one, so it will never take effect.

lib/src/linear_feedback_shift_register.dart
```diff
@@ -0,0 +1,114 @@
...
+        state < state ^ taps // Perform XOR and assign back to state
```
mkorbel1, line 92: If `taps` is a `Logic`, then it should be created as an input to the module.  I'm not sure it's necessary to support dynamic re-muxing of the LFSR, in which case the taps could be a non-`Logic` (e.g. a `List<bool>`, or maybe some `int`/`BigInt` where the bits correspond to the polynomial

lib/src/linear_feedback_shift_register.dart
```diff
@@ -0,0 +1,114 @@
...
+      [for (var i = 0; i < shifts; i++) output(_stageName(i))]);
+
+  /// The name of the signal (and output pin) for the [i]th stage.
+  String _stageName(int i) => '${dataName}_stage_$i';
+
+  LinearFeedbackShiftRegister(Logic dataIn,
+      {required Logic clk,
+      required this.state,
+      required this.shifts,
+      required this.taps,
```
mkorbel1, line 54: Careful using `this.` for inputs of a module, since we want to replace them using `addInput`

test/linear_feedback_shift_register_test.dart
```diff
@@ -0,0 +1,195 @@
...
+    const shifts = 6;
+    final taps = Logic(width: 4)..put(bin('1010')); // tap positions
+    final lfsr = LinearFeedbackShiftRegister(dataIn,
+        clk: clk, state: state, shifts: shifts, taps: taps);
+
+    final dataOut = lfsr.dataOut;
+
+    await lfsr.build();
+
+    final expectedData = [8, 12, 6, 3, 9, 4, 2, 1];
```
mkorbel1, line 33: how did you determine the expected values?

mkorbel1 (commented):
Nice start!

mkorbel1:
@omoiti are you still interested in completing this PR?

omoiti:
> @omoiti are you still interested in completing this PR?

I won't be able to complete the PR. I will go and close it.

PR #110, opened 2024-10-15
Refactoring of `FloatingPointValue` constructors for ease of use and extension

mkorbel1:
## Description & Motivation

Making it easier to extend `FloatingPointValue` and reuse constructors, removing `factory`s, etc.

## Related Issue(s)

N/A

## Testing

Existing tests should cover.

## Backwards-compatibility

No

## Documentation

No

### DIFF
--- a/lib/src/arithmetic/floating_point/floating_point_logic.dart
+++ b/lib/src/arithmetic/floating_point/floating_point_logic.dart
@@ -11,8 +11,7 @@
 //
 
 import 'package:rohd/rohd.dart';
-import 'package:rohd_hcl/src/arithmetic/floating_point/floating_point_value.dart';
-import 'package:rohd_hcl/src/exceptions.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
 
 /// Flexible floating point logic representation
 class FloatingPoint extends LogicStructure {
--- a/lib/src/arithmetic/floating_point/floating_point_values/floating_point_32_value.dart
+++ b/lib/src/arithmetic/floating_point/floating_point_values/floating_point_32_value.dart
@@ -1,0 +1,97 @@
+import 'dart:typed_data';
+
+import 'package:meta/meta.dart';
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+
+/// A representation of a single precision floating point value
+class FloatingPoint32Value extends FloatingPointValue {
+  /// The exponent width
+  static const int exponentWidth = 8;
+
+  /// The mantissa width
+  static const int mantissaWidth = 23;
+
+  @override
+  @protected
+  int get constrainedExponentWidth => exponentWidth;
+
+  @override
+  @protected
+  int get constrainedMantissaWidth => mantissaWidth;
+
+  /// Constructor for a single precision floating point value
+  FloatingPoint32Value(
+      {required super.sign, required super.exponent, required super.mantissa});
+
+  /// Return the [FloatingPoint32Value] representing the constant specified
+  factory FloatingPoint32Value.getFloatingPointConstant(
+          FloatingPointConstants constantFloatingPoint) =>
+      FloatingPointValue.getFloatingPointConstant(
+              constantFloatingPoint, exponentWidth, mantissaWidth)
+          as FloatingPoint32Value;
+
+  /// [FloatingPoint32Value] constructor from string representation of
+  /// individual bitfields
+  FloatingPoint32Value.ofBinaryStrings(
+      super.sign, super.exponent, super.mantissa)
+      : super.ofBinaryStrings();
+
+  /// [FloatingPoint32Value] constructor from a single string representing
+  /// space-separated bitfields
+  FloatingPoint32Value.ofString(String fp, {super.radix})
+      : super.ofString(fp, exponentWidth, mantissaWidth);
+
+  /// [FloatingPoint32Value] constructor from a set of [BigInt]s of the binary
+  /// representation
+  FloatingPoint32Value.ofBigInts(super.exponent, super.mantissa, {super.sign})
+      : super.ofBigInts();
+
+  /// [FloatingPoint32Value] constructor from a set of [int]s of the binary
+  /// representation
+  factory FloatingPoint32Value.ofInts(int exponent, int mantissa,
+      {bool sign = false}) {
+    final (signLv, exponentLv, mantissaLv) = (
+      LogicValue.ofBigInt(sign ? BigInt.one : BigInt.zero, 1),
+      LogicValue.ofBigInt(BigInt.from(exponent), exponentWidth),
+      LogicValue.ofBigInt(BigInt.from(mantissa), mantissaWidth)
+    );
+
+    return FloatingPoint32Value(
+        sign: signLv, exponent: exponentLv, mantissa: mantissaLv);
+  }
+
+  /// Numeric conversion of a [FloatingPoint32Value] from a host double
+  factory FloatingPoint32Value.fromDouble(double inDouble) {
+    final byteData = ByteData(4)
+      ..setFloat32(0, inDouble)
+      ..buffer.asUint8List();
+    final bytes = byteData.buffer.asUint8List();
+    final lv = bytes.map((b) => LogicValue.ofInt(b, 32));
+
+    final accum = lv.reduce((accum, v) => (accum << 8) | v);
+
+    final sign = accum[-1];
+    final exponent =
+        accum.slice(exponentWidth + mantissaWidth - 1, mantissaWidth);
+    final mantissa = accum.slice(mantissaWidth - 1, 0);
+
+    return FloatingPoint32Value(
+        sign: sign, exponent: exponent, mantissa: mantissa);
+  }
+
+  /// Construct a [FloatingPoint32Value] from a Logic word
+  factory FloatingPoint32Value.fromLogic(LogicValue val) {
+    final sign = (val[-1] == LogicValue.one);
+    final exponent =
+        val.slice(exponentWidth + mantissaWidth - 1, mantissaWidth);
+    final mantissa = val.slice(mantissaWidth - 1, 0);
+    final (signLv, exponentLv, mantissaLv) = (
+      LogicValue.ofBigInt(sign ? BigInt.one : BigInt.zero, 1),
+      exponent,
+      mantissa
+    );
+    return FloatingPoint32Value(
+        sign: signLv, exponent: exponentLv, mantissa: mantissaLv);
+  }
+}
--- a/lib/src/arithmetic/floating_point/floating_point_values/floating_point_8_value.dart
+++ b/lib/src/arithmetic/floating_point/floating_point_values/floating_point_8_value.dart
@@ -1,0 +1,103 @@
+import 'dart:math';
+
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+
+/// A representation of a 8-bit floating point value as defined in
+/// [FP8 Formats for Deep Learning](https://arxiv.org/abs/2209.05433).
+class FloatingPoint8Value extends FloatingPointValue {
+  /// The exponent width
+  late final int exponentWidth;
+
+  /// The mantissa width
+  late final int mantissaWidth;
+
+  static double get _e4m3max => 448.toDouble();
+  static double get _e5m2max => 57344.toDouble();
+  static double get _e4m3min => pow(2, -9).toDouble();
+  static double get _e5m2min => pow(2, -16).toDouble();
+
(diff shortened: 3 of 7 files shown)


desmonddak (commented):
This structure looks good.  I think I need to add more tests for the new classes,
but the major restructuring is to have each FP type with its own class.  E.g., the 8-bit classes are now separated into two types.

lib/src/arithmetic/floating_point/floating_point_values/floating_point_value.dart
```diff
@@ -119,135 +118,159 @@ class FloatingPointValue implements Comparable<FloatingPointValue> {
...
+  FloatingPointValue.ofInts(int exponent, int mantissa,
+      {int exponentWidth = 0, int mantissaWidth = 0, bool sign = false})
+      : this(
+            sign: LogicValue.ofBigInt(sign ? BigInt.one : BigInt.zero, 1),
+            exponent: LogicValue.ofBigInt(BigInt.from(exponent), exponentWidth),
+            mantissa:
+                LogicValue.ofBigInt(BigInt.from(mantissa), mantissaWidth));
 
   /// Construct a [FloatingPointValue] from a Logic word
   factory FloatingPointValue.fromLogic(
```
mkorbel1, line 268: Why is this called `fromLogic` instead of `fromLogicValue`?  Also, the comment says "word"? Also, this shouldn't be a factory?

lib/src/arithmetic/floating_point/floating_point_values/floating_point_8_value.dart
```diff
@@ -0,0 +1,177 @@
...
+      : super.ofBigInts();
+
+  /// [FloatingPoint8E4M3Value] constructor from a set of [int]s of the binary
+  /// representation
+  FloatingPoint8E4M3Value.ofInts(super.exponent, super.mantissa, {super.sign})
+      : super.ofInts();
+
+  /// Numeric conversion of a [FloatingPoint8E4M3Value] from a host double
+  factory FloatingPoint8E4M3Value.fromDouble(double inDouble) {
+    if ((inDouble > maxValue) | (inDouble < minValue)) {
```
mkorbel1, line 77: can we move this consideration of max/min into the base `fromDouble` in `FloatingPointLogic`?

lib/src/arithmetic/floating_point/floating_point_values/floating_point_bf16_value.dart
```diff
@@ -0,0 +1,84 @@
...
+// 2024 October 15
+// Authors:
+//  Max Korbel <max.korbel@intel.com>
+//  Desmond A Kirkpatrick <desmond.a.kirkpatrick@intel.com>
+
+import 'package:meta/meta.dart';
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+
+/// A representation of a single precision floating point value
```
mkorbel1, line 16: doc comments on these classes all say "single precision"

lib/src/arithmetic/floating_point/floating_point_values/floating_point_value.dart
```diff
@@ -119,135 +118,159 @@ class FloatingPointValue implements Comparable<FloatingPointValue> {
   final int _maxExp;
   final int _minExp;
 
-  /// Factory (static) constructor of a [FloatingPointValue] from
-  /// sign, mantissa and exponent
-  factory FloatingPointValue(
+  /// A Map from the (exponentWidth, mantissaWidth) pair to the
```
mkorbel1, line 121: I feel like we should update this doc comment to explain how it's used and how people can leverage it to override default behavior

mkorbel1:
Yes, this is looking a lot better!

lib/src/arithmetic/floating_point/floating_point_values/floating_point_value.dart
desmonddak: I am not sure of the utility of the map. We can work without it for now.  I think if people are accessing the constructors from the base class, they won't get the special behavior of the sub-class constructors that I need to add.  We need to figure out what we want here, as FPV(e=4,m=3) is not going to be the same type as FP8E4M3 since the latter does not represent infinity.

desmonddak: Unfortunately the test for E4M3 calls the base class constructor and does not validate the fromDouble routine.  So we have bugs in the very basic tests for the subtypes I need to check.

lib/src/arithmetic/floating_point/floating_point_values/floating_point_value.dart
desmonddak: I'm struggling with having both factory and non-factory constructors.  For example, fromDouble() is a factory constructor as it does a lot of work before it can initialize fields.  But that forces sub-classes to also have a factory method that first checks for some things and then calls the base factory method, possibly then checking for other things.

lib/src/arithmetic/floating_point/floating_point_values/floating_point_bf16_value.dart
desmonddak: I'll sanitize these.

lib/src/arithmetic/floating_point/floating_point_values/floating_point_8_value.dart
desmonddak: This is specific to E4M3.  it's also wrong, so I have fixed it (it fails for 0.0 and negatives).

lib/src/arithmetic/floating_point/floating_point_values/floating_point_32_value.dart
```diff
@@ -0,0 +1,91 @@
...
+        .reduce((accum, v) => (accum << 8) | v);
+
+    return FloatingPoint32Value(
+        sign: accum[-1],
+        exponent: accum.slice(exponentWidth + mantissaWidth - 1, mantissaWidth),
+        mantissa: accum.slice(mantissaWidth - 1, 0));
+  }
+
+  /// Construct a [FloatingPoint32Value] from a Logic word
+  factory FloatingPoint32Value.fromLogic(LogicValue val) =>
```
mkorbel1, line 86: these don't need to be factorys, and also should also be updatd to `fromLogicValue`

lib/src/arithmetic/floating_point/floating_point_values/floating_point_value.dart
```diff
@@ -84,7 +83,7 @@ class FloatingPointValue implements Comparable<FloatingPointValue> {
   final LogicValue value;
 
   /// The sign of the value:  1 means a negative value
-  final LogicValue sign;
+  late final LogicValue sign;
```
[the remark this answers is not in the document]
mkorbel1: fixed

lib/src/arithmetic/floating_point/floating_point_values/floating_point_32_value.dart
mkorbel1: fixed

lib/src/arithmetic/floating_point/floating_point_values/floating_point_value.dart
mkorbel1: the `fromDouble` seems like a special case we can ignore for now.  i updated the docs for the mapped stuff and i think that should all work nicely for everything except the fromDouble

lib/src/arithmetic/floating_point/floating_point_logic.dart
```diff
@@ -84,19 +83,36 @@ class FloatingPoint64 extends FloatingPoint {
...
+    final mantissaWidth = 7 - exponentWidth;
+    if (!FloatingPoint8E5M2Value.isLegal(exponentWidth, mantissaWidth)) {
+      throw RohdHclException('FloatingPoint8 must follow E4M3 or E5M2');
+    } else {
+      return mantissaWidth;
+    }
+  }
+
+  /// Construct an 8-bit floating point number
+  FloatingPoint8E5M2({required super.exponentWidth})
```
soneryaldiz, line 116: Why is `exponentWidth` required? This class is explicitly e5m2.

lib/src/arithmetic/floating_point/floating_point_logic.dart
```diff
@@ -84,19 +83,36 @@ class FloatingPoint64 extends FloatingPoint {
...
   /// Construct an 8-bit floating point number
-  FloatingPoint8({required super.exponentWidth})
+  FloatingPoint8E4M3({required super.exponentWidth})
+      : super(mantissaWidth: _calculateMantissaWidth(exponentWidth));
+}
+
+/// Eight-bit floating point representation for deep learning: E5M2
+class FloatingPoint8E5M2 extends FloatingPoint {
+  /// Calculate mantissa width and sanitize
+  static int _calculateMantissaWidth(int exponentWidth) {
```
soneryaldiz, line 106: This calculation and checking are redundant.

lib/src/arithmetic/floating_point/floating_point_logic.dart
```diff
@@ -84,19 +83,36 @@ class FloatingPoint64 extends FloatingPoint {
             mantissaWidth: FloatingPoint64Value.mantissaWidth);
 }
 
-/// Eight-bit floating point representation for deep learning
-class FloatingPoint8 extends FloatingPoint {
+/// Eight-bit floating point representation for deep learning: E4M3
+class FloatingPoint8E4M3 extends FloatingPoint {
```
soneryaldiz, line 87: The two comments below for E5M2 applies to this class as well.

lib/src/arithmetic/floating_point/floating_point_logic.dart
mkorbel1: good point, thanks!

lib/src/arithmetic/floating_point/floating_point_values/floating_point_32_value.dart
```diff
@@ -0,0 +1,89 @@
...
+      : super.ofBigInts();
+
+  /// [FloatingPoint32Value] constructor from a set of [int]s of the binary
+  /// representation
+  FloatingPoint32Value.ofInts(super.exponent, super.mantissa, {super.sign})
+      : super.ofInts();
+
+  /// Numeric conversion of a [FloatingPoint32Value] from a host double
+  factory FloatingPoint32Value.fromDouble(double inDouble) {
+    final byteData = ByteData(4)..setFloat32(0, inDouble);
```
soneryaldiz, line 73: This code could be moved to parent method as in the other floating point types.

soneryaldiz (approved):
This refactoring looks clean.

lib/src/arithmetic/floating_point/floating_point_values/floating_point_32_value.dart
mkorbel1: only fp32 and fp64 have this code, and they're not quite identical.  is that what you mean?

soneryaldiz: Yes, 32 and 64 exploiting native ways. It is fine to keep here.

lib/src/arithmetic/floating_point/floating_point_values/floating_point_value.dart
```diff
@@ -119,134 +118,185 @@ class FloatingPointValue implements Comparable<FloatingPointValue> {
...
+  /// sign, exponent, and mantissa using the constructor provided in
+  /// [subtypeConstructorMap] if available, otherwise using the default
+  /// constructor.
+  FloatingPointValue toMappedSubtype() => FloatingPointValue.withMappedSubtype(
+      sign: sign, exponent: exponent, mantissa: mantissa);
+
+  /// [constrainedMantissaWidth] is the hard-coded mantissa width of the
+  /// sub-class of this floating-point value
+  @protected
+  int? get constrainedMantissaWidth => null;
```
soneryaldiz, line 201: What is the purpose of these `constrained*Width` given that all subclasses hard code the widths?

lib/src/arithmetic/floating_point/floating_point_values/floating_point_value.dart
```diff
@@ -392,12 +449,13 @@ class FloatingPointValue implements Comparable<FloatingPointValue> {
 
   /// Convert a floating point number into a [FloatingPointValue]
   /// representation. This form performs NO ROUNDING.
-  factory FloatingPointValue.fromDoubleIter(double inDouble,
+  @internal
+  factory FloatingPointValue.ofDoubleUnrounded(double inDouble,
```
soneryaldiz, line 453: The value add of this constructor is not clear given that `ofDouble` lets user to specify rounding mode.

soneryaldiz (approved):
This PR looks good. `ofDoubleUnrounded` could be refactored in a future PR.

PR #111, opened 2024-10-22
Gated counter and toggle gate

mkorbel1:
## Description & Motivation

Introduces two new components:
- A `ToggleGate` which helps avoid toggles on combinational hardware when it's not necessary
- A `GatedCounter` which partitions the flops into two groups and clock gates them independently as needed

## Related Issue(s)

Found https://github.com/intel/rohd/issues/516 during development of random tests on the `GatedCounter`

## Testing

Lots of unit tests and random tests

## Backwards-compatibility

No

## Documentation

Yes, needs API docs and component docs


doc/components/summation.md
```diff
@@ -38,3 +38,11 @@ The `Counter` also has a `Counter.simple` constructor which is intended for very
 // A counter which increments by 1 each cycle up to 5, then rolls over.
 Counter.simple(clk: clk, reset: reset, maxValue: 5);
 ```
+
+## Gated Counter
+
+The `GatedCounter` is a version of a `Counter` which contains a number of power-saving features including clock gating to save on flop power and enable gating to avoid unnecessary combinational toggles.
+
+The `GatedCounter` has a `clkGatePartitionIndex` which determines a dividing line for the counter to be clock gated such that flops at or above that index will be independently clock gated from the flops below that index. This is an effective method of saving extra power on many counters because the upper bits of the counter may change much less frequently than the lower bits (or vice versa).  If the index is negative or greater than or equal to the width of the counter, then the whole counter will be clock gated in unison.
```
desmonddak, line 46: Would it be crazy to make this a list?

lib/src/summation/counter.dart
```diff
@@ -17,6 +18,19 @@ class Counter extends SummationBase {
   /// The output value of the counter.
   Logic get count => output('count');
 
+  /// The main clock signal.
+  @visibleForTesting
```
desmonddak, line 22: This is interesting:  I think we sometimes enable access in test and I did not know there was a specific decorator

lib/src/summation/gated_counter.dart
```diff
@@ -0,0 +1,529 @@
+// Copyright (C) 2024 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// counter.dart
```
desmonddak, line 4: Does not match filename.

desmonddak (approved):
Only a few comments, mostly questions as to new things I learned from this code (like @visibleForTesting).

doc/components/toggle_gate.md
```diff
@@ -0,0 +1,16 @@
+# Toggle Gate
+
+The `ToggleGate` component is intended to help save power by avoiding unnecessary toggles through combinational logic. It accomplishes this by flopping the previous value of data and muxing the previous value to the `gatedData` output if the `enable` is low. By default, the flops within the `ToggleGate` are also clock gated for extra power savings, but it can be controlled via a `ClockGateControlInterface`.
+
+As an example use case, if you have a large arithmetic unit but only care about the result when a `valid` bit is high, you could use a `ToggleGate` so that the inputs to that combinational logic do not change unless `valid` is high.
+
```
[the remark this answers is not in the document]
mkorbel1: can you elaborate? what do you mean?

doc/components/summation.md
mkorbel1: you mean making it accept a list of indices rather than a single split index?  API-wise, perhaps that's good future proofing, but practically speaking it was surprisingly hard to get this working in all cases with even a single cut point.

lib/src/summation/counter.dart
mkorbel1: i also did not know about this until someone else showed me, I think it was @saw235, very useful indeed!

the full list from the `meta` package is here and there are some other good interesting ones too:
https://pub.dev/documentation/meta/latest/meta/meta-library.html

lib/src/summation/gated_counter.dart
mkorbel1: good catch, thanks!

PR #117, opened 2024-10-28
Multi Cycle Divider Enhancements: remainder output and isSigned input

kimmeljo:
## Description & Motivation

Adding a remainder output so that this unit can be used for implementing the modulo operator. Adding an isSigned input so that the unit can handle unsigned operands if the integrating environment wants this.

## Related Issue(s)

N/A

## Testing

Enhanced existing testbench/checker to test the remainder output and mix in the use of the isSigned input.

## Backwards-compatibility

No breaking changes.

## Documentation

Documentation has been updated accordingly.

### DIFF
--- a/doc/components/divider.md
+++ b/doc/components/divider.md
@@ -16,17 +16,24 @@ The inputs to the divider module are:
 The outputs of the divider module are:
 
 * `quotient` => the result of the division
+* `remainder` => the remainder of the division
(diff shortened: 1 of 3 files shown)


mkorbel1 (approved):
looks great!

mkorbel1 (approved):

PR #118, opened 2024-10-30
Flop compressor

desmonddak:
## Description & Motivation

This change contributes an option to the ColumnCompressor:  if a clk is provided, then the output rows will be flopped.

## Related Issue(s)

None.

## Testing

A single test checks to make sure that the output is delayed by a flop and inputs changing are blocked from changing
this value.

## Backwards-compatibility

No.

## Documentation

Yes.  TBD.


lib/src/arithmetic/addend_compressor.dart
```diff
@@ -170,8 +170,17 @@ class ColumnCompressor {
   /// The partial product array to be compressed
   final PartialProductArray pp;
 
+  /// The clk
+  Logic? clk;
+
+  /// Optional reset
+  Logic? reset;
+
+  /// Optional enable
```
mkorbel1, line 179: Maybe some more explanation on what this is enabling in the doc?

lib/src/arithmetic/addend_compressor.dart
```diff
@@ -170,8 +170,17 @@ class ColumnCompressor {
...
+  Logic? enable;
+
   /// Initialize a ColumnCompressor for a set of partial products
```
mkorbel1, line 182: Worth some clarification on what the clock, reset, and enable are used for if they are present/not present.

test/arithmetic/addend_compressor_test.dart
```diff
@@ -163,6 +199,50 @@ void main() {
...
+  test('single compressor evaluate flopped', () async {
+    final clk = SimpleClockGenerator(10).clk;
+    const widthX = 6;
+    const widthY = 6;
+    final a = Logic(name: 'a', width: widthX);
+    final b = Logic(name: 'b', width: widthY);
+
+    var av = 3;
+    const bv = 6;
+    for (final signed in [true]) {
```
mkorbel1, line 211: why a loop here?

PR #129, opened 2024-11-06
[MultiCycleDivider] Bug Fixes for Large Operands and Ready Indication Polarity

kimmeljo:
## Description & Motivation

The MultiCycleDivider has the following bugs:
(1) Polarity of the readyIn output is flipped relative to what it should be.
(2) Unsigned cases of division involving very large operands (MSB=1) do not produce correct results and can even deadlock the unit.

This PR fixes these and adds tests for the exceptional cases moving forward. 

## Related Issue(s)

N/A

## Testing

Added a new sequence for "evil" cases of division stressing large operands in both signed and unsigned. Fixed a bug in the testbench preventing testing of readyIn signal.

## Backwards-compatibility

This change is backward compatible as the interface to the divider has not changed at all. Only the outputs for a given set of inputs have changed in certain cases.

## Documentation

Documentation updated to address the case of signed division overflow which was also added as a test case and needs explicitly clarification on unit behavior.

### DIFF
--- a/lib/src/arithmetic/divider.dart
+++ b/lib/src/arithmetic/divider.dart
@@ -168,9 +168,9 @@ class MultiCycleDivider extends Module {
   void _build() {
     // to capture current inputs
     // as this operation takes multiple cycles
-    final aBuf = Logic(name: 'aBuf', width: dataWidth);
-    final rBuf = Logic(name: 'rBuf', width: dataWidth);
-    final bBuf = Logic(name: 'bBuf', width: dataWidth);
+    final aBuf = Logic(name: 'aBuf', width: dataWidth + 1);
+    final rBuf = Logic(name: 'rBuf', width: dataWidth + 1);
+    final bBuf = Logic(name: 'bBuf', width: dataWidth + 1);
     final signOut = Logic(name: 'signOut');
     final signNum = Logic(name: 'signNum');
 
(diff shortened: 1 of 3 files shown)


lib/src/arithmetic/divider.dart
```diff
@@ -249,13 +252,7 @@ class MultiCycleDivider extends Module {
                   ~bBuf.getRange(0, dataWidth - 2).or() &
                   (signOut ^ signNum),
               then: [
-                // special logic for when a is also most negative #
```
desmonddak, line 252: Since you are in an If block, would you want to stay with If/Else and not insert a mux?
If you have the unit tests to exercise all paths through this logic, this is a quick refactor.

lib/src/arithmetic/divider.dart
```diff
@@ -297,11 +300,11 @@ class MultiCycleDivider extends Module {
           // conditionally convert negative inputs to positive
           // and compute the output sign
           aBuf <
-              mux(intf.dividend[dataWidth - 1] & intf.isSigned,
```
desmonddak, line 300: Optimization:  these computations are using bit inversion and then adding 1 (e.g. 2s complement conversion).
First idea is: is there a way to stay in 1s complement as that +1 is very costly (full carry).
Second idea is:  should we factor out a 2s complement converter so that this operation can be optimized and is very visible -- you don't have to use it, but it does make explicit this operation is not cheap.

lib/src/arithmetic/divider.dart
```diff
@@ -365,16 +368,17 @@ class MultiCycleDivider extends Module {
...
+              mux(extDividendIn[dataWidth - 1] & intf.isSigned,
+                  ~extDividendIn + 1, extDividendIn), // start by matching aBuf
         ]),
         ElseIf(
             currentState.eq(_MultiCycleDividerState
                 .process), // didn't exceed a_buf, so count as success
             [
               If(~tmpDifference[dataWidth - 1], then: [
                 lastSuccess <
-                    (Const(1, width: dataWidth) << currIndex), // capture 2^i
```
desmonddak, line 377: Should we make a feature request for width inference on muxes?  I even created a 'zero' and 'one' variable once to help minimize my typing with muxes, but I also ran into a lot of bugs when forgetting to expand a 0 or a 1 to the correct width.

test/arithmetic/divider_test.dart
```diff
@@ -399,6 +410,84 @@ class MultiCycleDividerBasicSequence extends Sequence {
...
 }
 
+class MultiCycleDividerEvilSequence extends Sequence {
+  late final int numBits;
+
+  MultiCycleDividerEvilSequence(
+      {required this.numBits, String name = 'MultiCycleDividerEvilSequence'})
+      : super(name);
+
+  @override
```
desmonddak, line 420: Great test for the corner cases! Are there other corners that could use a similar sequence (zero, one, etc.)

test/arithmetic/divider_test.dart
```diff
@@ -479,6 +568,7 @@ class MultiCycleDividerTest extends Test {
 
     // Kick off a sequence on the sequencer
     await _divSequencer.start(MultiCycleDividerBasicSequence());
+    await _divSequencer.start(MultiCycleDividerEvilSequence(numBits: 32));
```
desmonddak, line 571: Is it possible to create a narrow divider and do an exhaustive test?

lib/src/arithmetic/divider.dart
desmonddak, line 1: Should MultCycleDividerState be an enum?  ready/process/accumulate/done look like states to an FSM and you could use the FSM construct which is an enum-based way of representing states.
https://intel.github.io/rohd-website/docs/fsm/

This could be a really good example of a component (the first one) using the FSM construct because right now we only have the light controller.  It may not be that big a rewrite since you already have a sequencer wrapping the state transition, you just need to see  if you can output the logic from each state in a way that aligns with the FSM construct.

lib/src/arithmetic/divider.dart
kimmeljo: That's fine with me. Since it's quick and non-functional here, I can add it to this PR.

lib/src/arithmetic/divider.dart
kimmeljo: This might be a good one to discuss further and split out of this PR? Also note that given the way things are flopped, the 2's complement inversions shouldn't really cause timing problems (the longest path delays are pretty short by design). That being said, I'm all for optimization - just might make more sense to break it out as it's pretty unrelated to the change set here.

lib/src/arithmetic/divider.dart
kimmeljo: I like that idea! That's often the source of build errors for me - forgetting to set the width correctly on a Const.

test/arithmetic/divider_test.dart
kimmeljo: Yeah any suggestions? Divide by zero is covered in the basic sequence. Maybe 0 divided by some non-zero number? I tried to cover all of the corners I could think of with this evil sequence already.

test/arithmetic/divider_test.dart
kimmeljo: Absolutely! That's a good idea - I can add try adding that. Maybe the other (future) PR that I alluded to above?

lib/src/arithmetic/divider.dart
kimmeljo: Yeah I think I originally made this thing before the FSM abstraction even existed which is why things are structured like this. I'm happy to refactor here but again probably better suited to a future PR.

PR #141, opened 2024-11-18
[MultiCycleDivider] Patch of prior bug fix to properly support unsigned division

kimmeljo:
## Description & Motivation

This patch addresses a bug in the previous fix for MultiCycleDivider (proper support for unsigned division) that caused certain division operations to be performed incorrectly due to checking of improper bits.

## Related Issue(s)

[](https://github.com/intel/rohd-hcl/pull/129)

## Testing

Added an explicit unit test on a known case where pre-patch functionality was broken.

## Backwards-compatibility

This is fully backwards compatible as the patch is entirely internal to the implementation.

## Documentation

No explicit documentation was required for the patch but documentation was still enhanced to discuss performance of the module in more detail (motivated by this fix).


mkorbel1 (approved):
Looks good to me!

Issue #120, opened 2024-10-31
ColumnCompressor should be support more compressors like 4:2


desmonddak:
### Motivation

Right now, the ColumnCompressor supports only 2:1 (HalfAdder) and 3:2 (FullAdder) column compressors.
Adding more compressor types allows for experimentation, especially in different technologies where different compressor building blocks may win.

### Desired solution

Add a 4:2 compressor and demonstrate the algorithm can select the right compressor in the algorithm;  say there are  8 rows left in a column, the algorithm should end up with 3 4:2s versus a mix of choosing a 4:2, then a 3:2 and ending up imbalanced.

### Additional details

We likely will need to improve the delay-driven nature and find a way to introduce a table of delays for the various compressors.
This infrastructure could lead to also allowing for input delays.  The protocol for initializing this may need to be explored as well.

Issue #86, opened 2024-08-09
Majority function support


desmonddak:
### Motivation

The majority function is very powerful and general.  It is becoming quite interesting as post-CMOS devices can directly implement this function in a dense manner. It is useful in arithmetic operations.

### Desired solution

A LogicValue:majority() function, similar to the unary or() that produces true when the majority of bits are 1.

We can have various components that implement the Logic form of this.

PR #142, opened 2024-11-20
42compress

desmonddak:
## Description & Motivation

Adds 4:2 compressors option to ColumnCompressor as well as a better delay table for optimizing the compression tree.

Added a popCount as a LogicValue extension which may need to be migrated to ROHD itself (along with majority).  Eventually we will need hardware versions of these.

## Related Issue(s)

#120 #86

## Testing

Modified tests to enable the 4:2 compression option.

## Backwards-compatibility

compression trees will take on a different shape with different delay tables as the delay optimization gets more realistic.

## Documentation

Yes.  Code is updated.  We could update the component generation to expose this option as well as we need to update the documentation of multipliers that use this option.

### DIFF
--- a/lib/src/arithmetic/addend_compressor.dart
+++ b/lib/src/arithmetic/addend_compressor.dart
@@ -10,7 +10,6 @@
 import 'package:collection/collection.dart';
 import 'package:meta/meta.dart';
 import 'package:rohd/rohd.dart';
-import 'package:rohd_hcl/src/arithmetic/evaluate_compressor.dart';
 import 'package:rohd_hcl/src/arithmetic/multiplier_lib.dart';
 import 'package:rohd_hcl/src/exceptions.dart';
 
@@ -122,7 +121,7 @@ class CompressTerm implements Comparable<CompressTerm> {
   String toString() {
     final str = StringBuffer();
     final ts = switch (type) {
-      CompressTermType.pp => 'p',
+      CompressTermType.pp => 'pp',
       CompressTermType.carry => 'c',
       CompressTermType.cout => 'o',
       CompressTermType.sum => 's',
@@ -135,8 +134,6 @@ class CompressTerm implements Comparable<CompressTerm> {
   }
 }
 
-// TODO(desmonddak): Could we use List<CompressTerms> in constructor instead
-// PRoblem is we pass BitCOmpressor into CompressTErm constructor for delay
 /// Base class for bit-level column compressor function
 abstract class BitCompressor extends Module {
   /// Input bits to compress
@@ -155,12 +152,11 @@ abstract class BitCompressor extends Module {
   late final List<List<double>> _delays;
 
   /// Construct a column compressor.
-  BitCompressor(Logic compressBits) {
-    this.compressBits = addInput(
-      'compressBits',
-      compressBits,
-      width: compressBits.width,
-    );
+  BitCompressor(this.terms) {
+    compressBits = [
+      for (var pos = 0; pos < terms.length; pos++)
+        addInput('t_$pos', terms[pos].logic)
+    ].swizzle();
     addOutput('sum');
     addOutput('carry');
     _delays = List.filled(CompressTermType.values.length,
@@ -175,7 +171,7 @@ abstract class BitCompressor extends Module {
 /// 2-input column compressor (half-adder)
 class Compressor2 extends BitCompressor {
   /// Construct a 2-input compressor (half-adder).
-  Compressor2(super.compressBits) {
+  Compressor2(super.terms) {
     sum <= compressBits.xor();
     carry <= compressBits.and();
     _delays[CompressTermType.sum.index][CompressTermType.pp.index] = 1.0;
@@ -186,12 +182,13 @@ class Compressor2 extends BitCompressor {
 /// 3-input column compressor (full-adder)
 class Compressor3 extends BitCompressor {
   /// Construct a 3-input column compressor (full-adder).
-  Compressor3(super.compressBits) {
+  Compressor3(super.terms) {
     sum <= compressBits.xor();
     carry <=
         mux(compressBits[0], compressBits.slice(2, 1).or(),
             compressBits.slice(2, 1).and());
     // TODO(desmonddak): wiring different inputs for different delays
+    // means we may need to index by input not just type
     _delays[CompressTermType.sum.index][CompressTermType.pp.index] = 1.0;
     _delays[CompressTermType.carry.index][CompressTermType.pp.index] = 1.5;
   }
@@ -203,14 +200,17 @@ class Compressor4 extends BitCompressor {
   Logic get cout => output('cout');
 
   /// Construct a 4-input column compressor using two 3-input compressors.
-  Compressor4(super.compressBits, List<Logic> cinL) {
+  Compressor4(List<CompressTerm> terms, List<CompressTerm> cinL)
+      : super(terms) {
     for (final cin in cinL) {
-      addInput('cin', cin);
+      addInput('cin', cin.logic);
     }
     addOutput('cout');
-    final c3A = Compressor3(compressBits.slice(2, 0));
+    final c3A = Compressor3(terms.sublist(1, 4));
     cout <= c3A.carry;
-    final c3B = Compressor3([c3A.sum, compressBits[3], cinL[0]].swizzle());
+    final t = CompressTerm(
+        c3A, CompressTermType.sum, c3A.sum, terms.sublist(1, 4), 0, 0);
+    final c3B = Compressor3([t, terms[0], cinL[0]]);
     carry <= c3B.carry;
     sum <= c3B.sum;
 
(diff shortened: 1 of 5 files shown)


lib/src/arithmetic/addend_compressor.dart
```diff
@@ -158,15 +134,115 @@ class CompressTerm implements Comparable<CompressTerm> {
...
+    final c3A = Compressor3(internalTerms.sublist(1, 4));
+    cout <= c3A.carry;
+    final t = CompressTerm(
+        c3A, CompressTermType.sum, c3A.sum, internalTerms.sublist(1, 4), 0, 0);
+    final c3B = Compressor3([t, internalTerms[0], cinL[0]]);
+    carry <= c3B.carry;
+    sum <= c3B.sum;
+
+    // TODO(desmonddak): wiring different inputs for different delays
+    _delays[CompressTermType.sum.index][CompressTermType.pp.index] = 4.0;
```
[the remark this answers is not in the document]
desmonddak: Relative delays of paths through half-adder and full-adder based on or/and/xor transition paths.
We likely need to load these based on technology available.

lib/src/arithmetic/addend_compressor.dart
```diff
@@ -270,9 +271,9 @@ class ColumnCompressor {
   ColumnCompressor(this.pp,
       {this.use42Compressors = false, this.clk, this.reset, this.enable}) {
     columns = List.generate(pp.maxWidth(), (i) => ColumnQueue());
-    if (use42Compressors) {
-      carryColumns = List.generate(pp.maxWidth(), (i) => ColumnQueue());
-    }
+    // if (use42Compressors) {
```
mkorbel1, line 274: delete unused if statement code?

lib/src/arithmetic/addend_compressor.dart
```diff
@@ -285,23 +286,26 @@ class ColumnCompressor {
...
         final value = colList[row].logic;
 
         rowBits.add(
             clk != null ? flop(clk!, value, reset: reset, en: enable) : value);
+      } else {
+        rowBits.add(Const(0));
       }
     }
-    rowBits.addAll(List.filled(pp.rowShift[row], Const(0)));
+    // rowBits.addAll(List.filled(pp.rowShift[row], Const(0)));
```
mkorbel1, line 308: delete?

lib/src/utils.dart
```diff
@@ -49,6 +49,13 @@ extension LogicValueMajority on LogicValue {
     }
     return result;
   }
+
+  /// Return the populationCount of 1s in a LogicValue
+  int popCount() {
```
mkorbel1, line 54: I'm a little concerned about continuing to grow these extensions as public API additions from ROHD-HCL.  If we add them to ROHD later, these will conflict, and we'll have some annoying decisions to make on deprecation and backwards incompatibility.  Can we make these (at least new) extensions private to ROHD-HCL, or even to the files that are using them?  I only see this one used in one test.

desmonddak: Agreed.  I'll make these private.  Some are experimental and then end up being useful in a single case.  We can decide later what the best way to support these is if they end up being useful more broadly.

lib/src/arithmetic/addend_compressor.dart
```diff
@@ -12,60 +12,66 @@ import 'package:meta/meta.dart';
...
-class Compressor3 extends BitCompressor {
-  /// Construct a 3-input column compressor (full-adder)
-  Compressor3(super.compressBits, {super.name = 'compressor_3'}) {
-    sum <= compressBits.xor();
-    carry <=
-        mux(compressBits[0], compressBits.slice(2, 1).or(),
-            compressBits.slice(2, 1).and());
-  }
-}
+// /// Base class for bit-level column compressor function
```
mkorbel1, line 15: Accidentally left this old code in here commented out?

lib/src/arithmetic/addend_compressor.dart
```diff
@@ -158,15 +177,116 @@ class CompressTerm implements Comparable<CompressTerm> {
...
+  Logic get cout => output('cout');
+
+  /// Construct a 4-input column compressor using two 3-input compressors.
+  Compressor4(List<CompressTerm> terms, List<CompressTerm> cinL,
+      {super.name = 'bitcompressor4'})
+      : super(terms) {
+    // We need to use internal Logic and regenerate Term lists inside
+    cinL = [
+      for (final cin in cinL)
+        CompressTerm(this, cin.type, addInput('cin', cin.logic), cin.inputs,
```
mkorbel1, line 252: this `addInput` is in a loop but uses the same name each time, which seems to me that either:
- this shouldn't be in a loop, i.e. `cinL` length must be <=1, or
- `cinL` may be >1, but tests are not covering it, as it would fail

lib/src/arithmetic/addend_compressor.dart
```diff
@@ -158,15 +177,116 @@ class CompressTerm implements Comparable<CompressTerm> {
...
+  /// Input terms to compress
+  late final List<CompressTerm> terms;
+
+  /// The addition results [sum] including carry bit
+  Logic get sum => output('sum');
+
+  /// The carry results [carry].
+  Logic get carry => output('carry');
+
+  late final List<List<double>> _delays;
```
mkorbel1, line 195: do you think it may be useful to expose a `@protected` API for setting these delays rather than private, so that if someone has a technology-specific delay they want to set, they could just extend one of the compressors and change the delay numbers?  this also leads to another question: should there be an argument to generate compressors in `ColumnCompressor` (even the smaller ones) rather than just a flag on whether to use 4:2 or not?

test/arithmetic/addend_compressor_test.dart
```diff
@@ -8,13 +8,56 @@
...
+    var result = 0;
+    while (shiftedValue != zero) {
+      result += (shiftedValue[0] & LogicValue.one == LogicValue.one) ? 1 : 0;
+      shiftedValue >>>= 1;
+    }
+    return result > (width ~/ 2);
+  }
+
+  /// Compute the first One find operation on LogicValue, returning its position
+  int? firstOne() {
```
mkorbel1, line 37: you could still import that other non-public file to gain access to this, rather than copy-paste the function

test/arithmetic/multiplier_encoder_test.dart
```diff
@@ -530,12 +530,4 @@ void main() {
     }
     expect(pp.evaluate(), equals(product));
   });
-
-  test('majority function', () async {
```
mkorbel1, line 534: might as well keep this test and just move it over to addend_compressor_test?

lib/src/arithmetic/addend_compressor.dart
desmonddak: I should look at the tests as well as output Verilog.  line 157 does it correctly and that is the common case, not the 4-input case.

lib/src/arithmetic/addend_compressor.dart
desmonddak: The bare minimum is to use 2:2 and 3:2 compressors (half and full-adders).  But you have a point -- people have invented even wider compressors.

lib/src/arithmetic/addend_compressor.dart
ganewto: Turns out for 42 compressors only one input is added via the cinL.  But if I use this pattern for wider compressors, I will need to have differently named inputs.  The tests were working and I checked that the Verilog made sense.   I migrated this to a cinL.indexed  iteration to get both index and Logic and I named it 'cin_$index'.

test/arithmetic/addend_compressor_test.dart
mkorbel1: i still see it in both "evaluate_partial_product.dart" and "logicvalue_extension.dart"

lib/src/arithmetic/multiplier.dart
```diff
@@ -357,7 +357,15 @@ class CompressionTreeMultiplier extends Multiplier {
...
-      super.name = 'compression_tree_multiplier'}) {
+      super.name = 'compressison_tree_multiplier'})
+      : super(
+            definitionName: 'CompressionTreeMultiplier_W${a.width}x'
+                '${b.width}_'
+                '${signedMultiplicand ? 'SD_' : ''}'
+                '${signedMultiplier ? 'SM_' : ''}'
+                '${selectSignedMultiplicand != null ? 'SSD_' : ''}'
+                '${selectSignedMultiplier != null ? 'SSM_' : ''}'
+                'with${adderGen(a, b).definitionName}') {
```
mkorbel1, line 368: this call to `adderGen` i think will actually construct an adder connected to those ports, which could lead to additional verilog generation (_outside_ of the module, even) and also slow down simulation since it's simulating a whole additional adder.  i think i remember seeing this pattern in a few other places as well and didnt think it through.

lib/src/arithmetic/addend_compressor.dart
mkorbel1: do you want to include changes to support arbitrary delays and user-defined compressors in this PR?  one consideration is this flag to "use 4:2 compressors" is an API that wouldn't really make sense if we later replace it with a list of compressor types that can be used, for example

Issue #143, opened 2024-11-21
CompressionTreeMultiplier should support mixed signed/unsigned operands


desmonddak:
### Motivation

It would be useful to have the multiplier components support a mix of signed and unsigned multiplier and multiplicand.
Today, it supports operands that are either both signed or both unsigned.  
A workaround exists to support mixed by widening the unsigned operand by 1 bit.  But in some cases this is not possible to do, or expensive for shorter operands.

### Desired solution

Change the signed parameter to both operands (multiplier and multiplicand).
Change the signedOperands input to both operand types (say signedMultiplierOp and signedMultiplicandOp).
Modify the logic to separate the support of sign extension versus booth encoding of sign.

### Additional details

Consider the potential for using fixed sign versus runtime sign configuration for the parameter versus runtime-based configuration (do we need to support both or can we presume fully fixed and fully runtime configurable).  Should we have, say, signedMultiplier (hardcoded) and signedMultiplicandOperand runtime configuration.

PR #144, opened 2024-12-03
Full set of sign variants (signed/unsigned/selected) for multiplier operands

desmonddak:
## Description & Motivation

This updates the compression tree mulitplication components to fully handle variations in sign for multiplicand and multiplier.
This includes separate fixed signs for both, as well as a select signal for both and all valid combinations.  When select is enabled for an operand, the fixed sign input parameter is false..

The core parameters are:
- signedMultiplicand
- signedMultiplier
- selectSignedMultiplicand
- selectSignedMultiplier

These parameters are enabled for the CompressionTreeMultiplier and CompressionTreeMultiplyAccumulate.

Major cleanup of the testing routines to ensure solid coverage and reduce testing time.

## Related Issue(s)

#143 

## Testing

Sweep testing of the PartialProductGenerators for all combinations of signed/selected operands and all sign extension methods, across all radix settings.

## Backwards-compatibility

Yes.  The signed parameter is now split into signedMultiplicand and signedMultiplier.

## Documentation

Yes.  Documentation for the Multipliers is updated.  The configurator is also updated.

### DIFF
--- a/lib/src/arithmetic/multiplier.dart
+++ b/lib/src/arithmetic/multiplier.dart
@@ -6,7 +6,8 @@
 // need to inherit this module to ensure consistency.
 //
 // 2023 May 29
-// Author: Yao Jing Quek <yao.jing.quek@intel.com>
+// Author: Yao Jing Quek <yao.jing.quek@intel.com>, Desmond Kirkpatrick
+// <desmond.a.kirkpatrick@intel.com>
 
 import 'package:meta/meta.dart';
 import 'package:rohd/rohd.dart';
@@ -17,28 +18,36 @@ import 'package:rohd_hcl/src/arithmetic/partial_product_sign_extend.dart';
 abstract class Multiplier extends Module {
   /// The multiplicand input [a].
   @protected
-  late final Logic a;
+  Logic get a => input('a');
 
   /// The multiplier input [b].
   @protected
-  late final Logic b;
+  Logic get b => input('b');
 
   /// The multiplier treats input [a] always as a signed input.
-  bool signedMultiplicand;
+  @protected
+  final bool signedMultiplicand;
 
   /// The multiplier treats input [b] always as a signed input.
-  bool signedMultiplier;
+  @protected
+  final bool signedMultiplier;
 
   /// If not null, use this signal to select between signed and unsigned
   /// multiplicand [a].
-  late Logic? selectSignedMultiplicand;
+  @protected
+  Logic? get selectSignedMultiplicand => tryInput('selectSignedMultiplicand');
 
   /// If not null, use this signal to select between signed and unsigned
   /// multiplier [b]
-  late Logic? selectSignedMultiplier;
+  @protected
+  Logic? get selectSignedMultiplier => tryInput('selectSignedMultiplier');
 
   /// The multiplication results of the multiplier.
   Logic get product;
+
+  /// Logic that tells us [product] is signed.
+  @protected
+  Logic get isProductSigned => output('isProductSigned');
 
   /// Take input [a] and input [b] and return the
   /// [product] of the multiplication result.
@@ -57,13 +66,29 @@ abstract class Multiplier extends Module {
   /// signed or unsigned operation, overriding the [signedMultiplier] static
   /// configuration.
   Multiplier(Logic a, Logic b,
-      {required this.signedMultiplicand,
-      required this.signedMultiplier,
-      this.selectSignedMultiplicand,
-      this.selectSignedMultiplier,
+      {this.signedMultiplicand = false,
+      this.signedMultiplier = false,
+      Logic? selectSignedMultiplicand,
+      Logic? selectSignedMultiplier,
       super.name}) {
-    this.a = addInput('a', a, width: a.width);
-    this.b = addInput('b', b, width: b.width);
+    a = addInput('a', a, width: a.width);
+    b = addInput('b', b, width: b.width);
+
+    selectSignedMultiplicand = (selectSignedMultiplicand != null)
+        ? addInput('selectSignedMultiplicand', selectSignedMultiplicand)
+        : null;
+    selectSignedMultiplier = (selectSignedMultiplier != null)
+        ? addInput('selectSignedMultiplier', selectSignedMultiplier)
+        : null;
+
+    addOutput('isProductSigned') <=
+        (signedMultiplicand | signedMultiplier ? Const(1) : Const(0)) |
+            ((selectSignedMultiplicand != null)
+                ? selectSignedMultiplicand
+                : Const(0)) |
+            ((selectSignedMultiplier != null)
+                ? selectSignedMultiplier
+                : Const(0));
   }
 }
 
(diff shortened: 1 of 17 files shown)


lib/src/arithmetic/multiplier.dart
```diff
@@ -227,31 +353,64 @@ class MutiplyOnly extends MultiplyAccumulate {
...
+    super.signedMultiplier = false,
+    super.signedAddend = false,
+    super.selectSignedMultiplicand,
+    super.selectSignedMultiplier,
+    super.selectSignedAddend,
+  }) // Will be overrwridden by multiplyGenerator
+  : super(
+            // ignore: prefer_interpolation_to_compose_strings
+            name: 'Multiply Only: ' +
+                _genName(mulGen, a, b, selectSignedMultiplicand,
```
mkorbel1, line 389: is this constructing a temporary piece of hardware just for the sake of getting its name?

lib/src/arithmetic/partial_product_generator.dart
```diff
@@ -202,35 +202,52 @@ abstract class PartialProductGenerator extends PartialProductArray {
   /// multiples of the multiplicand and generate partial products
   late final MultiplicandSelector selector;
 
-  /// Operands are signed
-  late final bool signed;
+  /// [multiplicand] operand is always signed
+  late final bool signedMultiplicand;
```
mkorbel1, line 206: these don't need to be `late`

lib/src/arithmetic/multiplier.dart
```diff
@@ -15,23 +15,53 @@ import 'package:rohd_hcl/src/arithmetic/partial_product_sign_extend.dart';
...
+  /// If not null, use this signal to select between signed and unsigned
+  /// multiplicand [a].
+  late Logic? selectSignedMultiplicand;
+
+  /// If not null, use this signal to select between signed and unsigned
+  /// multiplier [b]
+  late Logic? selectSignedMultiplier;
 
   /// The multiplication results of the multiplier.
   Logic get product;
```
mkorbel1, line 41: Is the product signed or unsigned?  Clarify in the API doc?

test/arithmetic/multiplier_test.dart
```diff
@@ -15,28 +15,69 @@ import 'package:rohd_hcl/src/arithmetic/evaluate_compressor.dart';
...
+}
+
 /// Simple multiplier to demonstrate instantiation of CompressionTreeMultiplier
 class SimpleMultiplier extends Module {
   /// The output of the simple multiplier
   late final Logic product;
 
   /// Construct a simple multiplier with runtime sign operation
-  SimpleMultiplier(Logic a, Logic b, Logic multASigned)
+  SimpleMultiplier(
```
mkorbel1, line 62: is this `SimpleMultiplier` used anywhere? should it be?

test/arithmetic/multiplier_test.dart
```diff
@@ -198,17 +306,18 @@ void main() {
...
                   width,
                   10,
                   curryMultiplierAsMultiplyAccumulate(radix, ppTree,
                       ppGen: PartialProductGeneratorStopBitsSignExtension.new,
-                      signed: !signedOperands && signedTest,
-                      selectSign: signedSelect),
-                  signedTest: signedTest);
+                      signedMultiplicand: !signedOperands && signedTest,
+                      signedMultiplier: !signedOperands && signedTest,
+                      selectSignedMultiplicand: signedSelect,
```
mkorbel1, line 319: i see some places where both the multiplicand and multiplier are tied to the same signal -- are you sure you've tested all the combinations of signed and unsigned here?

mkorbel1:
This is awesome stuff!!

lib/src/arithmetic/multiplier.dart
```diff
@@ -15,23 +15,53 @@ import 'package:rohd_hcl/src/arithmetic/partial_product_sign_extend.dart';
...
 
-  /// The input to the multiplier pin [b].
+  /// The multiplier input [b].
   @protected
   late final Logic b;
 
-  /// The multiplier treats operands and output as signed
-  bool signed;
+  /// The multiplier treats input [a] always as a signed input.
+  bool signedMultiplicand;
```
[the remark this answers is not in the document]
desmonddak: done.

lib/src/arithmetic/multiplier.dart
```diff
@@ -51,16 +81,52 @@ abstract class MultiplyAccumulate extends Module {
...
+
+  /// The MAC treats multiplier [b] as always signed.
+  bool signedMultiplier;
+
+  /// The MAC treats addend [c] as always signed.
+  bool signedAddend;
+
+  /// If not null, use this signal to select between signed and unsigned
+  /// multiplicand [a].
+  late Logic? selectSignedMultiplicand;
```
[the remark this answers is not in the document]
desmonddak: I migrated these to inputs using getters with tryInput.

lib/src/arithmetic/multiplier.dart
desmonddak: Yes.  It is done only at the beginning of each test to give me a clear idea of the parameters passed, which can affect the name.  A bit experimental, but the name of a module could be extended with some of its parameter settings -- very useful in testing, but maybe not so much in instantiation, though perhaps somewhat useful.

lib/src/arithmetic/multiplier.dart
desmonddak: I decided to extend the API to compute this value of signed or unsigned since it is not easy to do with a mix of boolean and null/non-null selectors.  I think there is probably some functional way to map/reduce this logic to a oneliner, but this is what I have:
```
   addOutput('isAccumulateSigned') <=
        (signedMultiplicand | signedMultiplier | signedAddend
                ? Const(1)
                : Const(0)) |
            ((selectSignedMultiplicand != null)
                ? selectSignedMultiplicand
                : Const(0)) |
            ((selectSignedMultiplier != null)
                ? selectSignedMultiplier
                : Const(0)) |
            ((selectSignedAddend != null) ? selectSignedAddend : Const(0));
```

test/arithmetic/multiplier_test.dart
desmonddak: I was using it to catch trace errors in instantiation.  I wrapped it in a test that helps to put other multipliers in the test harness that expects a MultiplyAccumulate wrapper.

test/arithmetic/multiplier_test.dart
desmonddak: I have the most extensive tests in multiply_encoder_test.dart which does this at the partial product generation level.
But I added complete signage tests at the multiplier level now.  One exhaustive test for all permutations of parameters is too long (signed/unsigned/selected per operand, sign extension methods, final adder generator, widthX, widthY, radix).

desmonddak:
Do I need to protect getters?  I just realized I left protectors in place when I converted inputs declarations to getters.

mkorbel1:
> Do I need to protect getters? I just realized I left protectors in place when I converted inputs declarations to getters.

You should generally mark `@protected` on any variable or function that should only be accessed by sub classes which extend that class.  If nothing should access them, then making them private is better.

mkorbel1 (approved):
Looks great!! Awesome!

PR #154, opened 2025-01-12
Refactor signextend

desmonddak:
## Description & Motivation

Refactored Sign extension out of the PartialProductGenerator to operate on it instead.  You can choose which one you want
or build a new one.  This makes the system more modular and names more clear.

## Related Issue(s)

None

## Testing

Ran existing regressions with sweep of sign extensions.

## Backwards-compatibility

Yes.  This changes the API of the PartialProductGenerator, Multiplier, and Mac.

## Documentation

Yes.  Code documentation (options) were updated as well as the markdown documenting multiplier components.


lib/src/arithmetic/partial_product_sign_extend.dart
```diff
@@ -31,40 +31,186 @@ enum SignExtension {
...
+
+  // is multiplicand.width == entry.length?
+  // width=> multiplicand.width + shift - 1;
+  /// Only used to get width as above
+  MultiplicandSelector get selector => ppg.selector; // selector.width accessed
+
+  /// Sign Extension class that operates on a [PartialProductGeneratorBase]
+  /// and sign-extends the entries.
+  PartialProductSignExtension(this.ppg, {this.name = 'no_sign_extension'}) {
+    //
```
mkorbel1, line 95: nit: extra comment floating?

lib/src/arithmetic/partial_product_sign_extend.dart
```diff
@@ -31,40 +31,186 @@ enum SignExtension {
...
+      SignExtension.stopBits => StopBitsSignExtension.new,
+      SignExtension.compact => CompactSignExtension.new,
+      SignExtension.compactRect => CompactRectSignExtension.new,
     };
 
 /// These other sign extensions are for assisting with testing and debugging.
 /// More robust and simpler sign extensions in case
 /// complex sign extension routines obscure other bugs.
+///
+/// /// A Partial Product Generator using None Sign Extension
```
mkorbel1, line 152: nit: extra `///`

lib/src/arithmetic/partial_product_sign_extend.dart
```diff
@@ -31,40 +31,186 @@ enum SignExtension {
...
+    signExtend();
+  }
+
+  @override
+  void signExtend() {
+    extender.signExtend();
+  }
+}
+
+/// A wrapper class for [NoneSignExtension] we used
```
mkorbel1, line 185: "we used during refactoring" is a confusing description for a public API.  Do we want to mark these `@Deprecated` or do they have other uses beyond just backwards compatibility for refactoring?

mkorbel1 (commented):
looks good generally to me, few minor things/questions

lib/src/arithmetic/partial_product_sign_extend.dart
desmonddak: I intend to delete the old classes rather than maintain backward compatibility.  I could keep them and deprecate. I forgot to delete this one.  I maintained the old ones for a while to switch between old and new for testing (basically my loop would just use one of the new ones along with the other old ones until I migrated them all.

They have basically the same functionality it is just that instead of extending the PartialProductGenerator class with subclasses for different sign extensions, I decided to have another set of classes that operate on the PPG class and do the sign extension.  I didn't like the naming convention (PartialProductGenerationCompactSignExtension).  Instead, just do product generation and then do sign extension of your choice, since these operations happen completely sequentially and have no interdependence (so should not be in the same class).

PR #156, opened 2025-01-15
Naming fp

desmonddak:
## Description & Motivation

Added names to Logic signals in key arithmetic blocks to make SystemVerilog (SV) more readable.  Most effort was on the FloatingPointAdders, some on the FloatingPointMultiplierSimple.

A key simplification of names was in the prefix tree generators which are now far more readable.

But this exposes a big challenge in the column compression and booth encoding areas where Logic names are all synthesized and complex.  This will require a separate PR.

## Related Issue(s)

None

## Testing

Currently only visually inspected the output SystemVerilog code.  A true test would be to read it into a consuming tool.  A potential bug may be a name was forced and conflicts with another and was not caught in the writer.

## Backwards-compatibility

No.  We do not provide any contract on SV output, just that it is functional.  These are naming changes only to the SV variables.

## Documentation

No.


desmonddak:
This branch was off of multfp so many changes are in common.  We will need to merge multfp first.

lib/src/arithmetic/adder.dart
```diff
@@ -76,13 +76,16 @@ class NativeAdder extends Adder {
     if (a.width != b.width) {
       throw RohdHclException('inputs of a and b should have same width.');
     }
+    final ax = a.zeroExtend(a.width + 1).named('ax', naming: Naming.mergeable);
```
mkorbel1, line 79: what's the naming convention with "ax" instead of "a"?

lib/src/arithmetic/multiplier_encoder.dart
```diff
@@ -21,17 +21,26 @@ class RadixEncode extends LogicStructure {
   /// 'sign' of multiple
   final Logic sign;
 
+  /// The
```
mkorbel1, line 24: incomplete comment

confapp/pubspec.yaml
```diff
@@ -24,6 +24,10 @@ dependencies:
 dependency_overrides:
   rohd_hcl:
     path: ../
+  rohd:
```
mkorbel1, line 27: Reminder: remove this before merging

lib/src/arithmetic/signals/floating_point_logic.dart
```diff
@@ -24,21 +24,31 @@ class FloatingPoint extends LogicStructure {
   /// [sign] bit with '1' representing a negative number
   final Logic sign;
 
+  static String _nameJoin(String? structName, String signalName) {
```
mkorbel1, line 27: doc comment? (even though private, helpful)

lib/src/arithmetic/signals/floating_point_logic.dart
```diff
@@ -47,42 +57,46 @@ class FloatingPoint extends LogicStructure {
...
+            'isNaN',
+            naming: Naming.mergeable,
+          );
 
   /// Return a Logic true if this FloatingPoint is an infinity
   /// by having its exponent field set to the NaN value (typically all
   /// ones) and a zero mantissa.
-  late final isInfinity = Logic(name: 'isInfinity', naming: Naming.mergeable)
-    ..gets(exponent.eq(floatingPointValue.infinity.exponent) & ~mantissa.or());
+
```
mkorbel1, line 76: nit: extra blank line here between doc comment and code

lib/src/utils.dart
```diff
@@ -89,8 +89,9 @@ Logic condFlop(
 }) =>
     (clk == null)
         ? d
-        : flop(clk, d,
-            en: en,
-            reset: reset,
-            resetValue: resetValue,
-            asyncReset: asyncReset);
+        : (Logic(name: '${d.name}_flopped', width: d.width)
```
mkorbel1, line 92: use `named` here too?

pubspec.yaml
```diff
@@ -18,3 +18,10 @@ dependencies:
 dev_dependencies:
   logging: ^1.0.1
   test: ^1.25.0
+
+dependency_overrides:
```
mkorbel1, line 22: reminder: remove before merging (switch to 0.6.1)

mkorbel1 (commented):
Looks great!

lib/src/arithmetic/signals/floating_point_logic.dart
mkorbel1, line 1: can you expose the `name` argument in sub-classs of FloatingPoint as well? e.g. `FloatingPoint32`'s constructore has no `name` argument)

lib/src/arithmetic/adder.dart
ganewto: x = extended.

lib/src/arithmetic/signals/floating_point_logic.dart
mkorbel1: also, the clones should return the right type

lib/src/utils.dart
ganewto: I don't think it is needed.  Here is the output
```
ParallelPrefixDecr  exp_decrement(.inp(larger_exponent_flopped),.out(out_0));
ParallelPrefixIncr  exp_increment(.inp(larger_exponent_flopped),.out(out));
```

lib/src/arithmetic/signals/floating_point_logic.dart
ganewto: Done for all the sub-classes.

lib/src/utils.dart
mkorbel1: I think what you have is functional (for the naming purposes), but you're doing the `Logic()..gets()` approach, which I thought we preferred `.named` instead?

lib/src/arithmetic/signals/floating_point_logic.dart
mkorbel1: can you override `clone` also so that it returns th eproper type?

lib/src/arithmetic/signals/fixed_point_logic.dart
mkorbel1, line 1: add _nameJoin for things like `isZero`

ganewto: done

diff.txt
mkorbel1, line 1: accidental check-in?

lib/src/arithmetic/signals/floating_point_logic.dart
mkorbel1: they could if you update the `+` operator for floatingpoint to create a floatingpointadder and use clone to create a copy, but i don't think you're doing that operator overloading right now.  you could achieve it with the floatingpointadder by constructing via a clone, though

lib/src/arithmetic/adder.dart
```diff
@@ -76,13 +76,19 @@ class NativeAdder extends Adder {
...
+    final bx = b.zeroExtend(a.width + 1).named('bx', naming: Naming.mergeable);
+    final apb = Add(ax, bx).sum.named('ax_plus_bx', naming: Naming.mergeable);
```
mkorbel1, line 81: not much savings here, could replace `Add(ax, bx).sum` with `(ax + bx)`

lib/src/utils.dart
ganewto: I didn't see this, now I understand.  Clearly it repeats itself getting another Logic specifying a width, etc, instead of adding .named().  Done.

lib/src/arithmetic/signals/floating_point_logic.dart
ganewto: I meant to talk about FloatingPointValue.  I have tried a couple of different iterations of trying to get it to pass back the actual type, not the base type of FPV, but the subclass (FPBF16V) when using the '+' operator on values. 
Once we added clones, I think I have a proposal for removing so much boilerplate in the subclasses.

diff.txt
ganewto: ugh -- yes.

lib/src/arithmetic/adder.dart
ganewto: Done.  Plus used better names that are less cryptic.  You asked me, so I knew they were bad.

lib/src/arithmetic/floating_point/floating_point_multiplier.dart
```diff
@@ -47,9 +47,15 @@ abstract class FloatingPointMultiplier extends Module {
...
         ..gets(output('product'));
 
+  /// The internal FloatingPoint logic to set
+  late final FloatingPoint internalProduct;
+
   /// Multiply two floating point numbers [a] and [b], returning result in
   /// [product].
   ///
+  /// If you specify the optional  [outProduct], the multiplier
+  /// will output into the specified output allowing for a wider output.
```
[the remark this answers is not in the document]
desmonddak: I did not test the narrowing case.  I can throw for now until I validate (e.g. product widths >= input widths).  There are 4 cases to check with wider and narrower exponent and mantissa.

lib/src/arithmetic/multiplier.dart
```diff
@@ -16,6 +16,15 @@ import 'package:rohd_hcl/src/arithmetic/partial_product_sign_extend.dart';
 
 /// An abstract class for all multiplier implementations.
 abstract class Multiplier extends Module {
+  /// The clk for the pipelined version of column compression.
+  Logic? clk;
```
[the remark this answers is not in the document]
desmonddak: I will work on this. 

Let me know which issues are gating merge.

lib/src/arithmetic/signals/floating_point_logic.dart
mkorbel1: ah, yes, i think there should be a way to do that (perhaps using clone).  we could file another issue tracking this if we dont want to squeeze it in this PR

lib/src/arithmetic/floating_point/floating_point_multiplier.dart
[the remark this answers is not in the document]
mkorbel1: throwing might be good, documentation should be there ideally about which things are supported -- if we want to add it later in another PR it would be good to file an issue so we don't lose track of it

mkorbel1 (approved):
Looks great to me once remaining unresolved comments are addressed

lib/src/arithmetic/multiplier.dart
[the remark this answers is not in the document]
mkorbel1: i think this change, at least `@protected`, is important since:
- without it, this is a public API
- even with protected, there's an implication that any implementation would need to call the right `addInput`s for it

lib/src/arithmetic/multiplier.dart
```diff
@@ -240,13 +258,19 @@ abstract class MultiplyAccumulate extends Module {
...
+      Logic? reset,
+      Logic? enable,
+      this.signedMultiplicand = false,
       this.signedMultiplier = false,
       this.signedAddend = false,
       Logic? selectSignedMultiplicand,
       Logic? selectSignedMultiplier,
       Logic? selectSignedAddend,
       super.name}) {
+    this.clk = (clk != null) ? addInput('clk', clk) : null;
```
mkorbel1, line 271: doesn't this similar code also need to be removed from the `CompressionTreeMultiplyAccumulate` with super calls instead?

lib/src/arithmetic/signals/floating_point_logic.dart
mkorbel1: filed a ticket so we can review it later: https://github.com/intel/rohd-hcl/issues/157

PR #159, opened 2025-01-29
AXI4 Interface and Functional Modeling

kimmeljo:
## Description & Motivation

AXI4 is a common protocol and thus adding a ROHD implementation for its interface and some of its behavior. The structure is very similar to what was done for APB4.

## Related Issue(s)

N/A

## Testing

Unit tests were created very similar to the unit tests for APB4.

## Backwards-compatibility

Fully backwards compatible.

## Documentation

Documentation is forthcoming.


mkorbel1 (commented):
Looks great so far!  Added some initial comments

lib/src/interfaces/axi4.dart
```diff
@@ -0,0 +1,643 @@
...
+// Definitions for the AXI interface.
+//
+// 2025 January
+// Author: Josh Kimmel <joshua1.kimmel@intel.com>
+
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/src/exceptions.dart';
+
+/// A grouping of signals on the [Axi4ReadInterface] interface based on direction.
+enum Axi4Direction {
```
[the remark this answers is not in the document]
kimmeljo: Yeah there wasn't much intentionality behind my choice - more so just copying how APB did it. We can discuss to decide if there's an advantage to making the switch.

lib/src/models/axi4_bfm/axi4_main.dart
```diff
@@ -0,0 +1,62 @@
...
+
+/// An agent for sending requests on [Axi4ReadInterface]s and [Axi4WriteInterface]s.
+///
+/// Driven read packets will update the returned data into the same packet.
+class Axi4MainAgent extends Agent {
+  /// AXI4 System Interface.
+  final Axi4SystemInterface sIntf;
+
+  /// AXI4 Read Interface.
+  final Axi4ReadInterface rIntf;
```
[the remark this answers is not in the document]
kimmeljo: That's actually a good question. I definitely know of instances where the block has a vectorized number of AXI interfaces. I can't think of an instance where the number of read instances doesn't match the number of write instances but maybe the spec further clarifies this?

Say we just support a vectorized number. Would we just instantiate that many drivers and sequencers?

lib/src/models/axi4_bfm/axi4_main.dart
```diff
@@ -0,0 +1,62 @@
...
+    required this.sIntf,
+    required this.rIntf,
+    required this.wIntf,
+    required Component parent,
+    String name = 'axiMainAgent',
+    this.timeoutCycles = 500,
+    this.dropDelayCycles = 30,
+  }) : super(name, parent) {
+    sequencer = Sequencer<Axi4RequestPacket>('sequencer', this);
+
```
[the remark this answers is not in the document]
kimmeljo: I guess yeah it probably should, right? I was just copying what APB did here (it also doesn't have a monitor).

lib/src/models/axi4_bfm/axi4_subordinate.dart
```diff
@@ -0,0 +1,374 @@
...
+
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+import 'package:rohd_hcl/src/models/axi4_bfm/axi4_bfm.dart';
+import 'package:rohd_vf/rohd_vf.dart';
+
+/// A model for the subordinate side of
+/// an [Axi4ReadInterface] and [Axi4WriteInterface].
+class Axi4SubordinateAgent extends Agent {
+  /// The system interface.
```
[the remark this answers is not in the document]
kimmeljo: This change would follow from our decision to split out the read and write drivers presumably.

lib/src/models/axi4_bfm/axi4_monitor.dart
```diff
@@ -0,0 +1,160 @@
...
+
+import 'dart:async';
+
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+import 'package:rohd_hcl/src/models/axi4_bfm/axi4_bfm.dart';
+import 'package:rohd_vf/rohd_vf.dart';
+
+/// A monitor for [Axi4ReadInterface]s and [Axi4WriteInterface]s.
+class Axi4Monitor extends Monitor<Axi4RequestPacket> {
```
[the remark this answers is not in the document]
kimmeljo: I guess it could? Doesn't seem like the logic is too complicated though for a single monitor to handle both reads and writes. Maybe there's another benefit though?

lib/src/interfaces/axi4.dart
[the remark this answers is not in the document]
mkorbel1: i think it's fine as-is

lib/src/models/axi4_bfm/axi4_main.dart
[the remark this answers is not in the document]
mkorbel1: Yes, you could perhaps group things (e.g. a "write agent" has a driver and sequencer bundled together).  I think a configurable number makes a lot of sense independently for reads and writes unless the spec specifically calls out that they must be the same number.

lib/src/models/axi4_bfm/axi4_main.dart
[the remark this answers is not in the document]
mkorbel1: sorry for the bad example haha, i think it makes sense probably to have a monitor included

lib/src/models/axi4_bfm/axi4_main_driver.dart
```diff
@@ -0,0 +1,198 @@
...
+
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/src/interfaces/interfaces.dart';
+import 'package:rohd_hcl/src/models/axi4_bfm/axi4_bfm.dart';
+import 'package:rohd_vf/rohd_vf.dart';
+
+/// A driver for the [Axi4ReadInterface] and [Axi4WriteInterface] interfaces.
+///
+/// Driving from the perspective of the Main agent.
+class Axi4MainDriver extends PendingClockedDriver<Axi4RequestPacket> {
```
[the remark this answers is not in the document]
mkorbel1: If you have independent numbers of write and read interfaces, then you'd want them split apart

lib/src/models/axi4_bfm/axi4_subordinate.dart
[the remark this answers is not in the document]
mkorbel1: right

lib/src/models/axi4_bfm/axi4_monitor.dart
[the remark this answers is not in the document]
mkorbel1: same consideration as above -- if you have different numbers of read and write interfaces

mkorbel1 (commented):
changes since last review look good

lib/src/models/axi4_bfm/axi4_compliance_checker.dart
```diff
@@ -0,0 +1,270 @@
...
+//
+// 2025 January
+// Author: Josh Kimmel <joshua1.kimmel@intel.com>
+
+import 'dart:async';
+
+import 'package:rohd_hcl/rohd_hcl.dart';
+import 'package:rohd_vf/rohd_vf.dart';
+
+// /// A checker for some of the rules defined in the AXI4 interface specification.
```
mkorbel1, line 15: delete old code?

lib/src/models/axi4_bfm/axi4_main.dart
```diff
@@ -0,0 +1,179 @@
...
+  final Axi4WriteInterface? wIntf;
+
+  /// Constructor.
+  Axi4Channel({
+    this.channelId = 0,
+    this.hasRead = true,
+    this.hasWrite = true,
+    this.rIntf,
+    this.wIntf,
+  })  : assert(
```
mkorbel1, line 43: asserts only execute in tests and debug mode -- should these be exceptions?

lib/src/models/axi4_bfm/axi4_main.dart
```diff
@@ -0,0 +1,179 @@
...
+  /// Read interface.
+  final Axi4ReadInterface? rIntf;
+
+  /// Write interface.
+  final Axi4WriteInterface? wIntf;
+
+  /// Constructor.
+  Axi4Channel({
+    this.channelId = 0,
+    this.hasRead = true,
```
mkorbel1, line 39: can't hasRead and hasWrite be determined automatically based on whether the corresponding interface was provided (non-null)?

lib/src/models/axi4_bfm/axi4_main.dart
```diff
@@ -0,0 +1,179 @@
...
+/// An agent for sending requests on
+/// [Axi4ReadInterface]s and [Axi4WriteInterface]s.
+///
+/// Driven read packets will update the returned data into the same packet.
+class Axi4MainAgent extends Agent {
+  /// AXI4 System Interface.
+  final Axi4SystemInterface sIntf;
+
+  /// Channels that the agent can send requests on.
+  final List<Axi4Channel> channels;
```
mkorbel1, line 62: The spec calls a "channel" something else?
https://developer.arm.com/documentation/102202/0300/Channel-signals

Conceptually, is there a reason why we should have a list of channels rather than just a list of read interfaces and a list of write interfaces provided to the main agent?  With read and write agents created per read/write interface?  That would result in collection of the corresponding sequencer, driver, and monitor on a per-interface basis without the need to keep track of a channel index across multiple independent `List`s.

lib/src/models/axi4_bfm/axi4_monitor.dart
```diff
@@ -0,0 +1,330 @@
...
+            _pendingReadResponseData.removeAt(targIdx);
+          }
+        }
+      }
+    });
+  }
+}
+
+/// A monitor for [Axi4WriteInterface]s.
+class Axi4WriteMonitor extends Monitor<Axi4WriteRequestPacket> {
```
mkorbel1, line 245: prefer 1 class per file unless they are tightly related (e.g. a private class as a utility)

lib/src/models/axi4_bfm/axi4_subordinate.dart
```diff
@@ -0,0 +1,654 @@
...
+/// and can be leveraged for wrapping bursts.
+class AxiAddressRange {
+  /// Starting address of the range.
+  final LogicValue start;
+
+  /// Ending address of the range (exclusive).
+  final LogicValue end;
+
+  /// Secure region.
+  final bool isSecure;
```
mkorbel1, line 28: do we need to do any extra testing for secure and privileged?

lib/src/models/axi4_bfm/axi4_compliance_checker.dart
```diff
@@ -0,0 +1,270 @@
...
+          rIntf.arValid.previousValue!.toBool()) {
+        final id = rIntf.arId?.previousValue?.toInt() ?? 0;
+        final len = (rIntf.arLen?.previousValue?.toInt() ?? 0) + 1;
+        readReqMap[id] = [len, 0];
+      }
+
+      // track read response flits
+      if (rIntf.rValid.previousValue!.isValid &&
+          rIntf.rValid.previousValue!.toBool()) {
+        final id = rIntf.rId?.previousValue?.toInt() ?? 0;
```
mkorbel1, line 177: `previousValue` will always be non-null after the first tick in the simulator -- since you're waiting for reset before you can get here, it must be non null (use `!`)

lib/src/models/axi4_bfm/axi4_packet.dart
```diff
@@ -0,0 +1,231 @@
...
+    _completer.complete();
+  }
+
+  @override
+  String? trackerString(TrackerField field) {
+    switch (field.title) {
+      case Axi4Tracker.timeField:
+        return Simulator.time.toString();
+      case Axi4Tracker.idField:
+        return Simulator.time.toString();
```
mkorbel1, line 95: typo -- should be id?

lib/src/models/axi4_bfm/axi4_subordinate.dart
```diff
@@ -0,0 +1,654 @@
...
+  // assesses the input ready signals and drives them appropriately
+  void _driveReadReadys({int index = 0}) {
+    // for now, assume we can always handle a new request
+    channels[index].rIntf!.arReady.put(true);
+  }
+
+  // assesses the input ready signals and drives them appropriately
+  void _driveWriteReadys({int index = 0}) {
+    // for now, assume we can always handle a new request
+    channels[index].wIntf!.awReady.put(true);
```
mkorbel1, line 227: you probably want to use `inject` for a BFM -- main impact is waveforms will make more sense

lib/src/models/axi4_bfm/axi4_main.dart
kimmeljo: Yeah fair enough! I can change that

lib/src/models/axi4_bfm/axi4_main.dart
kimmeljo: Yeah I guess? Can we throw exceptions in the initialization list like the assertions now? What's the syntax?

lib/src/models/axi4_bfm/axi4_main.dart
kimmeljo: So the notion of a "channel" as I've defined it has a few implications. Essentially, for certain actions like locking we need to keep track of which "main" performed the action. This is what I use the notion of a channel for. It is also a nice logical distinction/categorization of the interfaces to understand "who" they are associated with.

That being said, sounds like we might need a new for "channel"? I just find this metadata to be useful to help reason about the system/network.

lib/src/models/axi4_bfm/axi4_subordinate.dart
kimmeljo: Yeah we absolutely do. I think there is some coverage through sheer randomness but we will want directed tests at some point.

lib/src/models/axi4_bfm/axi4_packet.dart
kimmeljo: Ah yup!

lib/src/models/axi4_bfm/axi4_main.dart
mkorbel1: you can't throw exceptions there, but you can do checks within the body of the constructor and throw an exception if they fail (a little more verbose/clunky, but probably appropriate for this case)

lib/src/models/axi4_bfm/axi4_main.dart
mkorbel1: Ah I see, I didn't realize that there was a relationship between a "pair" of read and write agents.  Does it not make sense to still group, for example, a write sequencer, write driver, and write monitor into one "write agent", and similar for read?

mkorbel1 (commented):
Looking really good now!

test/axi4_bfm_test.dart
```diff
@@ -0,0 +1,722 @@
...
+    Simulator.registerEndOfSimulationAction(() async {
+      await tracker.terminate();
+
+      // final jsonStr =
+      //     File('$outFolder/axi4Tracker.tracker.json').readAsStringSync();
+      // final jsonContents = json.decode(jsonStr);
+
+      // Here can do any checking against the tracker contents...
+
+      // Directory(outFolder).deleteSync(recursive: true);
```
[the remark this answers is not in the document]
kimmeljo: Would you say it's sufficient to just enable the tracker generation but not really check its contents? Or should we actually be checking it? Not really sure what checks are worth doing...

lib/src/models/axi4_bfm/axi4_write_compliance_checker.dart
```diff
@@ -0,0 +1,83 @@
...
+
+import 'dart:async';
+
+import 'package:rohd_hcl/rohd_hcl.dart';
+import 'package:rohd_vf/rohd_vf.dart';
+
+/// A checker for some of the rules defined in the AXI4 interface specification.
+///
+/// This does not necessarily cover all rules defined in the spec.
+class Axi4WriteComplianceChecker extends Component {
```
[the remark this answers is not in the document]
kimmeljo: Ah yeah fair. I mostly have been relying on the compliance checkers to cause tests that should pass to fail. But I don't really have any explicit negative space testing here. In theory the Subordinate BFM shouldn't ever violate the compliance checkers though. Would the suggestion then be to create a set of tests that just exclude the Subordinate BFM entirely?

test/axi4_bfm_test.dart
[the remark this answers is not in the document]
mkorbel1: parsing the json output is one way to determine that your `Trackable`s are generating the strings you want, but you could also test that directly.  just generating is probably fine

lib/src/models/axi4_bfm/axi4_write_compliance_checker.dart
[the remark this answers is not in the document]
mkorbel1: Yeah you could even just shove specific values on signals manually to make the compliance checker fail if you want :)

Imagine someone else down the line was modifying the compliance checker, we want tests to ensure that what it checks for now never breaks

lib/src/models/axi4_bfm/axi4_subordinate.dart
```diff
@@ -87,7 +87,7 @@ class Axi4SubordinateAgent extends Agent {
   // to handle response read responses
   final List<List<Axi4ReadRequestPacket>> _dataReadResponseMetadataQueue = [];
   final List<List<List<LogicValue>>> _dataReadResponseDataQueue = [];
-  final List<List<bool>> _dataReadResponseErrorQueue = [];
+  // final List<List<bool>> _dataReadResponseErrorQueue = [];
```
mkorbel1, line 90: did you mean to leave this commented out?

kimmeljo: Yeah. I'm thinking we might want that in the future but for now we have no use for it.

lib/src/models/axi4_bfm/axi4_write_compliance_checker.dart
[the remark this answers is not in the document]
kimmeljo: Yeah, makes sense! I'll add some.

lib/src/models/axi4_bfm/axi4_main_read_driver.dart
```diff
@@ -0,0 +1,123 @@
...
+      rIntf.arProt.put(packet.prot);
+      rIntf.arQos?.put(packet.qos);
+      rIntf.arRegion?.put(packet.region);
+      rIntf.arUser?.put(packet.user);
+      rIntf.rReady.put(1);
+    });
+
+    // need to hold the request until receiver is ready
+    await sIntf.clk.nextPosedge;
+    if (!rIntf.arReady.value.toBool()) {
```
mkorbel1: i think you want `previousValue` here (on both read and write driver)?

mkorbel1 (changes requested):
Looks just about ready!

mkorbel1 (approved):
Looks great!

test/axi4_bfm_test.dart
```diff
@@ -0,0 +1,1257 @@
...
+    final tracker = Axi4Tracker(
+      dumpTable: false,
+      outputFolder: outFolder,
+    );
+
+    Simulator.registerEndOfSimulationAction(() async {
+      await tracker.terminate();
+
+      final jsonStr =
+          File('$outFolder/axi4Tracker.tracker.json').readAsStringSync();
```
mkorbel1: capital A in Axi to fix test issue

test/axi4_bfm_test.dart
```diff
@@ -0,0 +1,1257 @@
...
+}
+
+void main() {
+  tearDown(() async {
+    await Test.reset();
+  });
+
+  setUp(() async {
+    // Set the logger level
+    Logger.root.level = Level.SEVERE;
```
mkorbel1: set to off

mkorbel1 (approved):

PR #109, opened 2024-10-15
WIP: Float conv1

saw235:
## Description & Motivation
Software and Hardware Arbitrary Precision Float Converter

PR #161, opened 2025-01-30
FloatingPointConverter

desmonddak:
## Description & Motivation

Creates a new component that does general floating-point logic conversion for arbitrary widths.

## Testing

Exhaustive testing at small widths of exponent and mantissa, but with high variation.  Captures infinity results, subnormal to normal and normal to subnormal.

## Backwards-compatibility

No.

## Documentation

yes:  floating_point.md has documentation with and example.  API is documented in code.


mkorbel1:
To confirm, does this supersede #109?

doc/components/floating_point.md
mkorbel1, line 1: > yes: floating_point.md has documentation with and example. API is documented in code.

Where?

mkorbel1, line 1: > yes: floating_point.md has documentation with and example. API is documented in code.

Where?

mkorbel1 (commented):
awesome!

doc/components/floating_point.md
desmonddak: Ugh.  In my vscode buffer, unsaved.  I wish vscode had a clear marker of having unsaved files.  'Save' is always available.  'SaveAll' shows up highlighted when multiple files are not saved.
Perhaps I should learn to do my git operations from within VSCode (assuming it would check for save), but I do them in a separate terminal.

mkorbel1: The "X" for the tab turns into a white circle when a file is unsaved :)

mkorbel1 (approved):

PR #167, opened 2025-02-03
rounding and fixed point fixes

desmonddak:
## Description & Motivation

Missed adding code to FPAdderRound to check for infinity after adding.
Fixed FloatingPointValue to properly check for infinity in the rounding case.

## Testing

Existing FP round tests now pass (they were passing before, but FPV and the same bug as FPAdderRound).

## Backwards-compatibility

No
## Documentation

No change required.

### DIFF
--- a/lib/src/arithmetic/floating_point/floating_point_adder_round.dart
+++ b/lib/src/arithmetic/floating_point/floating_point_adder_round.dart
@@ -314,6 +314,7 @@ class FloatingPointAdderRound extends FloatingPointAdder {
     final isR = (deltaFlopped.gte(Const(2, width: delta.width)) |
             ~effectiveSubtractionFlopped)
         .named('isR');
+    final infExponent = outputSum.inf(sign: largerSignFlopped).exponent;
 
     Combinational([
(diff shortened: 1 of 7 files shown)


lib/src/arithmetic/floating_point/floating_point_adder_round.dart
```diff
@@ -309,9 +309,9 @@ class FloatingPointAdderRound extends FloatingPointAdder {
     final isR = (deltaFlopped.gte(Const(2, width: delta.width)) |
             ~effectiveSubtractionFlopped)
         .named('isR');
+    final infExponent = outputSum.inf(sign: largerSignFlopped).exponent;
```
mkorbel1, line 312: what does this change do? it looks equivalent?

desmonddak: It's equivalent.  I didn't pull from main correctly and ended up reimplementing the check for infinity in a slightly different but equivalent way.

mkorbel1 (approved):
Looks good!

PR #172, opened 2025-02-07
Fixed lossy

desmonddak:
## Description & Motivation

We need converters from float to fixed that are lossy (e.g. are constrained by fixed parameters m, n.  This PR creates this.

## Testing

Exhaustive testing at narrow widths, e.g., BF16->18.16.

## Backwards-compatibility

No.

## Documentation

Yes.  Markdown updated to document the addition of optional inputs m,n and the checkOverflow flag.

### DIFF
--- a/doc/components/fixed_point.md
+++ b/doc/components/fixed_point.md
@@ -12,13 +12,13 @@ The [FixedPoint](https://intel.github.io/rohd-hcl/rohd_hcl/FixedPoint-class.html
 
 ## FixedToFloat
 
-This component converts a fixed-point signal to a floating point signal specified by exponent and mantissa width. The output is rounded to the nearest even (RNE) when applicable and set to infinity if the input exceed the representable range.
+The [FixedToFloat](https://intel.github.io/rohd-hcl/rohd_hcl/FixedToFloat-class.html) component converts a fixed-point signal to a floating point signal specified by exponent and mantissa width. The output is rounded to the nearest even (RNE) when applicable and set to infinity if the input exceed the representable range.
 
 ## FloatToFixed
 
 This component converts a floating-point signal to a signed fixed-point signal. Infinities and NaN's are not supported. The integer and fraction widths are auto-calculated to achieve lossless conversion.
 
-If the `m` and `n` integer and fraction widths are supplied, then lossy conversion is performed to fit the floating-point value into the fixed-point value. For testing, [FixedPointValue] has a `canStore` method to predetermine if a given double can fit.  For execution, [FloatToFixed] can perform overflow detection by setting a `checkOverflow` option.
+If the `m` and `n` integer and fraction widths are supplied, then lossy conversion is performed to fit the floating-point value into the fixed-point value. For testing, [FixedPointValue](https://intel.github.io/rohd-hcl/rohd_hcl/FixedPointValue-class.html) has a `canStore` method to predetermine if a given double can fit.  For execution, [FloatToFixed](https://intel.github.io/rohd-hcl/rohd_hcl/FloatToFixed-class.html) can perform overflow detection by setting a `checkOverflow` option, which is a property of the class and set in the constructor (default is false as it must add significant logic to do the check).
 
 ## Float8ToFixed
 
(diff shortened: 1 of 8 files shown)


lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -143,18 +143,58 @@ class FixedPointValue implements Comparable<FixedPointValue> {
     return compareTo(other) == 0;
   }
 
-  /// Constructs [FixedPointValue] of a Dart [double] rounding away from zero.
+  /// Return a string representation of FloatingPointValue.
+  ///  return sign, exponent, mantissa as binary strings.
+  @override
+  String toString() =>
+      "(${signed ? '${value[-1].toString(includeWidth: false)} ' : ''}"
+      "${(m > 0) ? '${value.slice(m + n - 1, n).bitString} ' : ''}"
```
mkorbel1, line 151: why does the `m` piece not use `.toString(includeWidth: false)`?

lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -143,18 +143,58 @@ class FixedPointValue implements Comparable<FixedPointValue> {
...
+      '${value.slice(n - 1, 0).toString(includeWidth: false)})';
+
+  /// Return true if double [val] be stored in FixedPointValue with [m] and [n]
```
mkorbel1, line 154: typo: "val be stored"

also, does this mean "without precision loss"?

lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -143,18 +143,58 @@ class FixedPointValue implements Comparable<FixedPointValue> {
...
+      final l = (val < 0.0)
+          ? max(bigIntegerValue.bitLength, negBigIntegerValue.bitLength)
+          : bigIntegerValue.bitLength;
+      return l <= w;
+    }
+    return false;
+  }
+
+  /// Constructs [FixedPointValue] from a Dart [double] rounding away from zero.
   factory FixedPointValue.ofDouble(double val,
```
mkorbel1, line 171: should this have a "rounding mode" option instead of having two separate functions?

doc/components/fixed_point.md
```diff
@@ -12,12 +12,14 @@ The [FixedPoint](https://intel.github.io/rohd-hcl/rohd_hcl/FixedPoint-class.html
...
+This component converts a fixed-point signal to a floating point signal specified by exponent and mantissa width. The output is rounded to the nearest even (RNE) when applicable and set to infinity if the input exceed the representable range.
 
 ## FloatToFixed
 
-This component converts a floating-point signal to a signed fixed-point signal. Infinities and NaN's are not supported. The integer and fraction widths are auto-calculated to achieve lossles conversion.
+This component converts a floating-point signal to a signed fixed-point signal. Infinities and NaN's are not supported. The integer and fraction widths are auto-calculated to achieve lossless conversion.
+
+If the `m` and `n` integer and fraction widths are supplied, then lossy conversion is performed to fit the floating-point value into the fixed-point value. For testing, [FixedPointValue] has a `canStore` method to predetermine if a given double can fit.  For execution, [FloatToFixed] can perform overflow detection by setting a `checkOverflow` option.
```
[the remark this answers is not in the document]
desmonddak: Should I not instead add the html link to these classes? 
```markdown
[FixedPointValue](https://intel.github.io/rohd-hcl/rohd_hcl/FixedPointValue-class.html)
```

lib/src/arithmetic/values/fixed_point_value.dart
desmonddak: bitString = toString(includeWidth:false).    In this case, I run into the 80 character margin with the interpolated string.

lib/src/arithmetic/values/fixed_point_value.dart
desmonddak: overflow.  There will be precision loss.  I'll update.
Unfortunately, fixed-point has no Infinity support.

lib/src/arithmetic/values/fixed_point_value.dart
desmonddak: That's a good idea.  I will need to refactor ofDouble, ofDoubleUnrounded in both FixedPointValue and FloatingPointValue.

I'd like to do that after figuring out removing factory methods and getting overrides working in FPV and FXPV to kill all subtype references in the base classes and as much boilerplate as possible in the subclasses.

desmonddak: Can I file this as an issue?  Could be an 'easy' one.

desmonddak:
I wonder if I should add a wrapper component called 'FloatToFixedPoint' which takes both a FloatingPoint and a FixedPoint, extending this Module and passing in the m, n from the FixedPoint.   For the lossless case, we still need this class.

lib/src/arithmetic/values/fixed_point_value.dart
desmonddak: Filed as Issue #173

PR #175, opened 2025-02-19
FloatingPointValue refactor for simpler resuable construction

mkorbel1:
## Description & Motivation

Major refactor of `FloatingPointValue` to remove factory constructors and put automation related to construction into "populators".

## Related Issue(s)

N/A

## Testing

Existing tests cover most of it, plus some new ones to cover additional functionality

## Backwards-compatibility

Not relative to a released version of ROHD-HCL, but a major set of backwards-incompatible changes relative to what was previously on `main`

## Documentation

Yes, included


lib/src/arithmetic/signals/floating_point_logic.dart
```diff
@@ -152,64 +173,88 @@ class FloatingPoint32 extends FloatingPoint {
   /// Construct a 32-bit (single-precision) floating point number
   FloatingPoint32({super.name})
```
desmonddak, line 174: Should we factor these into other files?

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
```diff
@@ -14,625 +14,201 @@ import 'package:meta/meta.dart';
...
-  largestPositiveSubnormal,
-
-  /// Smallest possible positive number, most negative exponent, mantissa is 0
-  smallestPositiveNormal,
-
-  /// Largest number smaller than one
-  largestLessThanOne,
-
-  /// The number one
-  one,
```
desmonddak, line 41: As silly as it seems, we should add zero.

desmonddak (approved):
This is so much nicer than before.  We need to get this merged.

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
mkorbel1: Is that different from positive zero?

```
  /// The number zero, negative form.
  negativeZero,

  /// The number zero, positive form.
  positiveZero,
```

lib/src/arithmetic/signals/floating_point_logic.dart
mkorbel1: yes, i'll do it now

mkorbel1: we were missing tf32!

desmonddak: ack!

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
desmonddak: oh that's right.  I forgot we had added.

PR #177, opened 2025-02-20
Allowing CSR Width to Exceed Frontdoor Interface Data Width

kimmeljo:
## Description & Motivation

Adds an optional configuration on the CSR block and top modules in which CSR widths can be greater than the frontdoor interface data width. In this case, frontdoor accesses are broken out until logical "chunks" with their own addresses. This can be convenient for implementations with parametrizable register widths (as well as implementations where this happens to be the mode of operation). 

## Related Issue(s)

N/A

## Testing

Enhanced/added to the existing CSR unit tests.

## Backwards-compatibility

This is a fully backwards compatible change due to the fact that the new behavior is controlled by a new configuration parameter.

## Documentation

Documentation has been updated accordingly.


lib/src/memory/csr/csr_block.dart
```diff
@@ -226,11 +243,39 @@ class CsrBlock extends Module {
...
-      throw CsrValidationException(
-          'Frontdoor read interface data width must be '
-          'at least ${_config.maxRegWidth()}.');
+      if (allowLargerRegisters) {
+        // must check for collisions in logical register addresses
+        final regCheck = <int>[];
+        for (final csr in csrs) {
+          if (csr.config.width > _frontRead.dataWidth) {
+            final rem = csr.width % _frontRead.dataWidth;
+            final targ = rem == 0
```
mkorbel1, line 252: should you use math `ceil` instead?

mkorbel1 (commented):
Looks really good!!

lib/src/memory/csr/csr_block.dart
kimmeljo: Yeah good call! Just fixed.

mkorbel1 (approved):
Looks good to me! Awesome work!

PR #178, opened 2025-02-21
faster priority encoder

desmonddak:
## Description & Motivation

Adds a new recursive priority encoder that more directly encodes priority and is faster (fewer logic levels) as it doesn't first
find the trailing-1 and then OneHot encode it, it does a more direct assembly of the bit position.

## Related Issue(s)

None.

## Testing

Ran the priority encoder tests, integrated into all existing arithmetic blocks and ran those tests (exhaustive).

## Backwards-compatibility

Yes.  We added a new PriorityEncoder base class, and mapped the existing ParallelPrefixPriorityEncoder to that in addition to the new class.  We also changed the APIs of the major floating point and converter classes to use a new functor (closure) for the
PriorityEncoder instead of the ParallelPrefixTree option.

## Documentation

Yes.  All API changes are documented in code and the markdown files have been updated for all components touched as well as adding the new PriorityEncoder markdown file.

### DIFF
--- a/doc/components/priority_encoder.md
+++ b/doc/components/priority_encoder.md
@@ -1,0 +1,30 @@
+# Priority Encoder
+
+A priority encoder is used to find the leading `1` in a Logic (or, equivalently, count the number of leading 0s). In ROHD HCL, our `PriorityEncoder` abstract class searches from the LSB to find the leading-1, returning the index of its position from the LSB. If there is no `1` in the Logic, it returns 1 past the width of the Logic.  Additionally, if a Logic `valid` output is provided, it will set that to `1` if a leading-1 is found, otherwise `0`.
+
+## Parallel-Prefix Priority Encoder
+
+We provide one implementation based on `ParallelPrefix` trees, which finds the leading-1 position using and then encodes that position.  This `ParallelPrefixPriorityEncoder` allows you to select the prefix-tree of your choice to use in the find position portion, which is followed by a `OneHotToBinary` module to encode that position.
+
+```dart
+    final bitVector = Logic(width: 5);
+    // ignore: cascade_invocations
+    bitVector.put(8);
+    final valid = Logic();
+    final encoder = ParallelPrefixPriorityEncoder(bitVector,
+        ppGen: BrentKung.new, valid: valid);
+    // encoder.out.value.toInt() will be 3 and valid.out.toBool() will be true
+```
+
+## Recursive Priority Encoder
+
+We provide a more direct encoding implementation that builds up the leading-1 position using a binary recursion tree.  There are two flavors of this implementation: `RecursivePriorityEncoder` is a straight recursion that builds up the binary tree construction using straightforward logic, whereas `RecursiveModulePriorityEncoder` uses the same algorithm, but produces a `Module` for each node of the tree (this results in more readable output Verilog).
+
+```dart
+    final bitVector = Logic(width: 5);
+    // ignore: cascade_invocations
+    bitVector.put(8);
+    final valid = Logic();
+    final encoder = RecursiveModulePriorityEncoder(bitVector, valid: valid);
+    // encoder.out.value.toInt() will be 3 and valid.out.toBool() will be true
+  ```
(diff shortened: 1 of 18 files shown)


doc/components/priority_encoder.md
```diff
@@ -0,0 +1,30 @@
+# Priority Encoder
+
+A priority encoder is used to find the leading `1` in a Logic (or, equivalently, count the number of leading 0s). In ROHD HCL, our `PriorityEncoder` abstract class searches from the LSB to find the leading-1, returning the index of its position from the LSB. If there is no `1` in the Logic, it returns 1 past the width of the Logic.  Additionally, if a Logic `valid` output is provided, it will set that to `1` if a leading-1 is found, otherwise `0`.
+
+## Parallel-Prefix Priority Encoder
+
+We provide one implementation based on `ParallelPrefix` trees, which finds the leading-1 position using and then encodes that position.  This `ParallelPrefixPriorityEncoder` allows you to select the prefix-tree of your choice to use in the find position portion, which is followed by a `OneHotToBinary` module to encode that position.
```
mkorbel1, line 7: typo: "which finds the leading-1 position using and then"

lib/src/priority_encoder.dart
```diff
@@ -0,0 +1,242 @@
...
+// 2025 February 13
+// Author: Desmond Kirkpatrick <desmond.a.kirkpatrick@intel.com>
+
+import 'dart:math';
+
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+
+/// Base class for priority encoders.
+class PriorityEncoder extends Module {
```
mkorbel1, line 16: this should be `abstract`?

lib/src/arithmetic/floating_point/floating_point_adder_simple.dart
```diff
@@ -16,17 +16,16 @@ class FloatingPointAdderSimple extends FloatingPointAdder {
...
   FloatingPointAdderSimple(super.a, super.b,
       {super.clk,
       super.reset,
       super.enable,
       Adder Function(Logic a, Logic b, {Logic? carryIn}) adderGen =
           NativeAdder.new,
-      ParallelPrefix Function(
-              List<Logic> inps, Logic Function(Logic term1, Logic term2) op)
-          ppTree = KoggeStone.new,
+      PriorityEncoder Function(Logic bitVector, {Logic? valid, String name})
```
mkorbel1, line 27: why require the function to support `name` if you're not using it?  probably good to keep it, but then use the name argument.

lib/src/arithmetic/floating_point/floating_point_multiplier_simple.dart
```diff
@@ -91,12 +90,12 @@ class FloatingPointMultiplierSimple extends FloatingPointMultiplier {
...
     final isNaNLatch = localFlop(isNaN);
 
-    final leadingOnePosPre = ParallelPrefixPriorityEncoder(mantissa.reversed,
-            ppGen: ppTree, name: 'leading_one_encoder')
-        .out
-        .named('leadingOneRaw')
-        .zeroExtend(exponentWidth + 2)
-        .named('leadingOneRawExtended', naming: Naming.mergeable);
+    final leadingOnePosPre =
+        priorityGen(mantissa.reversed, name: 'leading_one_encoder')
```
mkorbel1, line 94: not using the `valid` here?  should it not be required then in the argument, or should it have been used?

test/priority_encoder_test.dart
```diff
@@ -0,0 +1,112 @@
...
+
+  final generators = [Ripple.new, Sklansky.new, KoggeStone.new, BrentKung.new];
+  test('RecursivePriorityEncoder quick test', () {
+    final inp = Logic(width: 4)..put(8);
+    final valid = Logic();
+    final dut = RecursivePriorityEncoder(inp, valid: valid);
+    expect(dut.out.value.toInt(), equals(3));
+    expect(valid.value.toBool(), equals(true));
+  });
+  test('RecursivePriorityModuleEncoder quick test', () async {
```
mkorbel1, line 54: nit: new line between tests/groups

mkorbel1 (commented):
looks great!

PR #179, opened 2025-02-22
Compound adder

desmonddak:
## Description & Motivation

Addition of a compound adder that also does ones complement subtraction for use in floating point.

## Related Issue(s)

None

## Testing

Exhaustive tests of compound adder with ones complement using signed and spinning through exp and mantissa combinations.

## Backwards-compatibility

API changes in the base class.

## Documentation

Yes.  Markdown and code documentation updated.


lib/src/arithmetic/compound_adder.dart
```diff
@@ -68,27 +77,55 @@ class CarrySelectCompoundAdder extends CompoundAdder {
...
+  /// Generator of splitter algorithm
+  static Splitter splitSelectAdderAlgorithmNBit(int n) =>
+      (adderWidth) => _splitAdderNBitFunctor(adderWidth, n);
+
+  /// Default adder functor to use.
+  static Adder defaultAdder(Logic a, Logic b,
+          {Logic? carryIn, Logic? subtractIn, String name = ''}) =>
+      ParallelPrefixAdder(a, b, carryIn: carryIn, name: name);
+
   /// Constructs a [CarrySelectCompoundAdder].
```
mkorbel1, line 100: Would be good to add to the constructor documentation explaining the arguments (e.g. subtractIn behavior)

lib/src/arithmetic/compound_adder.dart
```diff
@@ -141,6 +178,101 @@ class CarrySelectCompoundAdder extends CompoundAdder {
     sumList1.add(carry1!);
 
     sum <= sumList0.rswizzle();
-    sum1 <= sumList1.rswizzle();
+    sumP1 <= sumList1.rswizzle();
+
+    // print('s  =${sum.value.bitString}');
```
mkorbel1, line 183: remove old/debug commented out code

lib/src/arithmetic/compound_adder.dart
```diff
@@ -68,27 +77,55 @@ class CarrySelectCompoundAdder extends CompoundAdder {
...
+  static Adder defaultAdder(Logic a, Logic b,
+          {Logic? carryIn, Logic? subtractIn, String name = ''}) =>
+      ParallelPrefixAdder(a, b, carryIn: carryIn, name: name);
+
   /// Constructs a [CarrySelectCompoundAdder].
   CarrySelectCompoundAdder(super.a, super.b,
-      {Adder Function(Logic a, Logic b, {Logic? carryIn, String name})
-          adderGen = ParallelPrefixAdder.new,
+      {Adder Function(Logic a, Logic b,
+              {Logic? carryIn, Logic? subtractIn, String name})
```
mkorbel1, line 103: do all adders that work with this carry select adder have to support `subtractIn` for it to behave properly?

lib/src/arithmetic/ones_complement_adder.dart
```diff
@@ -36,27 +40,30 @@ class OnesComplementAdder extends Adder {
           ParallelPrefixAdder.new,
       Logic? subtractIn,
       Logic? carryOut,
-      Logic? carryIn,
-      bool? subtract,
+      super.carryIn,
+      bool subtract = false,
+      bool chainable = false,
```
mkorbel1, line 45: explain `chainable` in doc?

test/arithmetic/adder_test.dart
```diff
@@ -318,6 +318,41 @@ void main() {
     }
   });
 
+  test('ones complement adder: random with boolean subtract', () {
+    const width = 4;
```
mkorbel1, line 322: why random if this is only adding 4-bit numbers?

test/arithmetic/adder_test.dart
```diff
@@ -318,6 +318,41 @@ void main() {
...
+    final a = Logic(width: width);
+    final b = Logic(width: width);
+    const nSamples = 100;
+
+    for (var i = 0; i < nSamples; i++) {
+      final av = Random().nextLogicValue(width: width);
```
mkorbel1, line 328: prefer to create a single `Random` object with a fixed seed for test reproducibility

lib/src/arithmetic/compound_adder.dart
desmonddak: This is a challenging API:   if the compound adder receives a subtractIn signal, then the adder generator must supply that signal.  But if it doesn't get one at its api, the sub-adders don't need it.

lib/src/arithmetic/compound_adder.dart
```diff
@@ -136,11 +179,103 @@ class CarrySelectCompoundAdder extends CompoundAdder {
...
+    if (carryOut != null) {
+      addOutput('carryOut');
+      carryOut <= this.carryOut!;
+    }
+    if (carryOutP1 != null) {
+      addOutput('carryOutP1');
+      carryOutP1 <= this.carryOutP1!;
+    }
+
+    final doSubtract = subtractIn ?? (subtract ? Const(subtract) : Const(0));
```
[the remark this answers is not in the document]
desmonddak: Ugh, I just saw that I don't have the same API for the CSCompoundAdder vs the OnesComplement version.
I have to think this through a bit: have I fixed enough of the API so that the CSCompoundAdder can do subtraction properly (not ones' complement).

lib/src/arithmetic/ones_complement_adder.dart
```diff
@@ -12,76 +12,95 @@ import 'package:rohd/rohd.dart';
...
-      carryOut <= this.carryOut!;
+      : super(
+            definitionName:
+                definitionName ?? 'OnesComplementAdder_W${a.width}') {
+    if (endAroundCarry != null) {
+      addOutput('endAroundCarry');
+      endAroundCarry <= this.endAroundCarry!;
     }
-    if ((subtractIn != null) & (subtract != null)) {
+    if ((subtractIn != null) & subtract) {
```
mkorbel1, line 62: why not make `subtract` a `bool?`?

test/ones_complement_adder_test.dart
```diff
@@ -0,0 +1,487 @@
...
+  final b = Logic(name: 'b', width: width);
+
+  final adder = SignMagnitudeAdder(aSign, a, bSign, b,
+      adderGen: fn, largestMagnitudeFirst: sortOperands);
+  test('random Sign Magnitude: ${adder.name}_W${a.width}_N$sortOperands',
+      () async {
+    await adder.build();
+
+    for (var i = 0; i < nSamples; i++) {
+      final aa = Random().nextLogicValue(width: width);
```
mkorbel1, line 121: make a single random object, with a fixed seed, so that tests are reproducible

test/ones_complement_adder_test.dart
```diff
@@ -0,0 +1,487 @@
...
+
+    for (final t in tests) {
+      aSign.put(t.aSign);
+      a.put(t.aMag);
+      bSign.put(t.bSign);
+      b.put(t.bMag);
+
+      final adder = SignMagnitudeAdder(aSign, a, bSign, b,
+          largestMagnitudeFirst: t.sorted, endAroundCarry: t.carry);
+      // expect(adder.sign.value, equals(t.sign));
```
mkorbel1, line 279: did you intend to leave this commented out?

lib/src/arithmetic/ones_complement_adder.dart
desmonddak: If subtract is a bool?  then I have to check for both null and false for the same code.  It seems redundant.  For a hard parameter (not a Logic),  we would only use true or false and default to false; in that case we would probably not be tempted to convert our parameters to nullable throughout, so why do it here?  Adjacent to the subtractIn, it seems odd,
but against regular hard parameters, it seems reasonable.

lib/src/arithmetic/compound_adder.dart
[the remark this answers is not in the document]
desmonddak: A bit of ugliness:   the CSCompoundAdder needs to pass 'subtractIn' as it is used by the CSOnesComplementCompoundAdder to perform subtraction.  It cannot do it directly without a special adderGen() because each sub-adder cannot do the end-around-carry, which most adders do when subtracting.  There's no way to hide this argument, so I documented this not to be used.

lib/src/arithmetic/ones_complement_adder.dart
mkorbel1: It seems like there are 3 options:
- this adder does not do subtraction always (`subtract = false`)
- this adder does do subtraction always (`subtract = true`)
- this adder can do addition or subtraction, based on whether `subtractIn` is driven to a 1 or 0

In the third case, it seems like either a value of `true` or `false` would be confusing.

There are other APIs i could also see as reasonable:
- a single `dynamic subtract` which must either be a `bool` or `Logic`, with default to `false`
- just a `subtractIn` and the default could be `Const(0)` (if null is provided?).  This would generate the additional SV, but synthesis would squash it down to equivalent

desmonddak: This seems like an issue we should add because there are multiple places I've used this pattern (fp components, integer multiply, etc.) and it threads through them.

Another alternative would be a small class that is BoolOrSignal  to carry that configuration, rather than dynamic (I see that and I think (void*)).  

Another pattern that may be an issue that I have been using is optional Logic outputs, where we might have used a bool instead and generated the Logic output, but I chose to pass in the Logic, which generates the output and is connected to it.  Another single-bit case.

So we could decide on a common pattern for HCL and I could implement that pattern throughout.

mkorbel1: sure, we can split it off in another issue.  so then we need to clarify at least the documentation.

```
  /// - Either an optional Logic [subtractIn] or a boolean [subtract] can enable
  /// subtraction, but providing both non-null will result in an exception.
```

we should be explicit that `subtract == false` and `subtractIn != null` is legal and that `subtractIn` wins, if you want to keep it that way for now

PR #180, opened 2025-02-22
Reduction module

desmonddak:
## Description & Motivation

Modified tree reduction to create modules instead of inline assignments.

## Related Issue(s)

None

## Testing

ran existing tree reduction tests

## Backwards-compatibility

No.  API and naming match.

## Documentation

None required.

### DIFF
--- a/lib/src/reduction_tree.dart
+++ b/lib/src/reduction_tree.dart
@@ -13,9 +13,11 @@ import 'package:meta/meta.dart';
 import 'package:rohd/rohd.dart';
 import 'package:rohd_hcl/rohd_hcl.dart';
 
-/// A generator which constructs a tree of radix-input / 1-output modules.
+/// Recursive Node for Reduction Tree
 class ReductionTree extends Module {
-  /// The radix-sized input operation to be performed at each node.
+  /// Operation to be performed at each node. Note that [operation] can widen
+  /// the output. The logic function must support the operation for [2 to radix]
+  /// inputs.
   @protected
   final Logic Function(List<Logic> inputs, {String name}) operation;
 
@@ -54,9 +56,17 @@ class ReductionTree extends Module {
   /// The flop depth of the tree from the output to the leaves.
   int get latency => _computed.flopDepth;
 
+  /// The input sequence
+  @protected
+  late final List<Logic> sequence;
+
   /// Capture the record of compute: the final value, its depth (from last
   /// flop or input), and its flopDepth if pipelined.
   late final ({Logic value, int depth, int flopDepth}) _computed;
+
+  /// Local conditional flop using module reset/enable
+  Logic localFlop(Logic d, {bool doFlop = false}) =>
+      condFlop(doFlop ? clk : null, reset: reset, en: enable, d);
 
   /// Generate a tree based on dividing the input [sequence] of a node into
   /// segments, recursively constructing [radix] child nodes to operate
@@ -64,12 +74,12 @@ class ReductionTree extends Module {
   /// - [sequence] is the input sequence to be reduced using the tree of
   /// operations.
   /// - Logic Function(List<Logic> inputs, {String name}) [operation]
-  /// is the operation to be
-  /// performed at each node. Note that [operation] can widen the output. The
-  /// logic function must support the operation for 2 to radix inputs.
+  /// is the operation to be performed at each node. Note that [operation]
+  /// can widen the output. The logic function must support the operation for
+  /// (2 to [radix]) inputs.
   /// - [radix] is the width of reduction at each node in the tree (e.g.,
   /// binary: radix=2).
-  /// - [signExtend] if true, use sign-extension to widen [Logic] values as
+  /// - [signExtend] if true, use sign-extension to widen Logic values as
   /// needed in the tree, otherwise use zero-extension (default).
   ///
   /// Optional parameters to be used for creating a pipelined computation tree:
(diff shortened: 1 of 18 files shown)


lib/src/reduction_tree.dart
```diff
@@ -13,9 +13,11 @@ import 'package:meta/meta.dart';
 import 'package:rohd/rohd.dart';
 import 'package:rohd_hcl/rohd_hcl.dart';
 
-/// A generator which constructs a tree of radix-input / 1-output modules.
+/// Recursive Node for Reduction Tree
 class ReductionTree extends Module {
-  /// The radix-sized input operation to be performed at each node.
+  /// Operation to be performed at each node. Note that [operation] can widen
+  /// the output. The logic function must support the operation for [2 to radix]
```
mkorbel1, line 19: nit: `[2 to radix]` is not a link to anything, maybe also worth clarifying that this means any number from 2 to [radix].

lib/src/reduction_tree.dart
```diff
@@ -54,22 +56,30 @@ class ReductionTree extends Module {
...
+  /// The input sequence
+  @protected
+  late final List<Logic> sequence;
+
   /// Capture the record of compute: the final value, its depth (from last
   /// flop or input), and its flopDepth if pipelined.
   late final ({Logic value, int depth, int flopDepth}) _computed;
 
+  /// Local conditional flop using module reset/enable
+  Logic localFlop(Logic d, {bool doFlop = false}) =>
```
mkorbel1, line 68: this should be private

lib/src/reduction_tree.dart
```diff
@@ -83,41 +93,43 @@ class ReductionTree extends Module {
...
+    if (this.sequence.length <= radix) {
+      final value = operation(this.sequence);
+      addOutput('out', width: value.width) <= value;
+      _computed = (value: output('out'), depth: 0, flopDepth: 0);
     } else {
       final results = <({Logic value, int depth, int flopDepth})>[];
-      final segment = seq.length ~/ radix;
+      final segment = this.sequence.length ~/ radix;
+
+      // final divisor = (log(sequence.length - 1) / log(radix)).floor();
```
mkorbel1, line 114: remove old code

mkorbel1 (changes requested):

lib/src/reduction_tree.dart
```diff
@@ -83,41 +93,43 @@ class ReductionTree extends Module {
...
+            this
+                .sequence
+                .getRange(
+                    pos, (i < radix - 1) ? pos + segment : this.sequence.length)
+                .toList(),
+            operation,
+            radix: radix,
+            signExtend: signExtend,
+            depthToFlop: depthToFlop,
+            clk: this.clk,
```
mkorbel1, line 129: suggestion: keeping this logic generation in another function (e.g. `_buildLogic()`) makes it easier to avoid accidentally using the wrong version of variables without needing to put `this`.  Additionally, you could consider making some of these variable (e.g. clk, enable, reset, sequence) private instead of protected, unless there's a reason to extend this.

lib/src/reduction_tree.dart
```diff
@@ -54,91 +24,130 @@ class ReductionTree extends Module {
   /// The flop depth of the tree from the output to the leaves.
   int get latency => _computed.flopDepth;
 
+  /// Operation to be performed at each node. Note that [_operation] can widen
+  /// the output. The logic function must support the operation for 2 and up to
+  /// [_radix] inputs.
+  final Logic Function(List<Logic> inputs, {String name}) _operation;
+
+  /// Specified width of input to each reduction node (e.g., binary: radix=2)
+  late final int _radix;
```
mkorbel1, line 33: primitive types like `int` and `bool`, or really any non-hardware-signal/module thing (even `Function`s like `operation`), especially when they are directly passed from the user, can be just public `final` variables with `this.radix` type of syntax in the constructor.  That's also better for when you reference those variables in documentation comments `[radix]` vs `[_radix]`

lib/src/reduction_tree.dart
```diff
@@ -54,91 +24,130 @@ class ReductionTree extends Module {
...
+  ReductionTree(List<Logic> sequence, this._operation,
+      {int radix = 2,
+      bool signExtend = false,
+      int? depthToFlop,
       Logic? clk,
       Logic? enable,
       Logic? reset,
       super.name = 'reduction_tree'})
-      : super(definitionName: 'ReductionTree_R${radix}_L${sequence.length}}') {
-    if (sequence.isEmpty) {
```
mkorbel1, line 87: does this now support empty sequences?

lib/src/reduction_tree.dart
```diff
@@ -9,136 +9,159 @@
...
+        final end1 =
+            _nextSplitDistance(_sequence.length - pos, radix, radix - i) + pos;
+        final tree = ReductionTree(
+            _sequence.getRange(pos, end1).toList(), _operation,
+            radix: radix,
+            signExtend: signExtend,
+            depthToFlop: depthToFlop,
+            clk: _clk,
+            enable: _enable,
+            reset: _reset);
```
mkorbel1, line 140: do you want to pass `name` down the hierarchy? maybe a modified version of it based on level or something?

lib/src/reduction_tree.dart
```diff
@@ -9,136 +9,159 @@
...
 class ReductionTree extends Module {
-  /// The radix-sized input operation to be performed at each node.
-  @protected
-  final Logic Function(List<Logic> inputs, {String name}) operation;
+  /// The final output of the tree computation.
+  Logic get out => output('out');
+
+  /// The combinational depth since the last flop. The total compute depth of
+  /// the tree is: depth + flopDepth * depthToflop;
+  int get depth => _computed.depth;
```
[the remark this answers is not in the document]
desmonddak: Yes.  I shouldn't talk about total compute depth since it really isn't relevant and it is confusing.  If you wanted to know the total amount of logic depth (say if you were going to re-flop it), then you might want to do that calculation.  But it isn't accurate since the tree does not have to be balanced perfectly.

lib/src/reduction_tree.dart
desmonddak: I will just pass the top level name down.  I don't want to modify that name (say with depth) as that causes Nodes to be uniquified and results in multiple Cells that are basically the same but different because of a name inside.  Because of splitting, we can reuse a Cell Master at different levels of hierarchy when the tree is imbalanced.

mkorbel1 (approved):

PR #182, opened 2025-02-25
`FloatingPointUtilities` extraction for `swap`/`sort`, and parameterization of some FP components

mkorbel1:
## Description & Motivation

The `swap` and `sort` routines for `FloatingPoint` signals are useful beyond just inside of ROHD-HCL `FloatingPointAdder`s, so this PR extracts them as static functions to a separate `FloatingPointUtilities` class.

Additionally, add some parameterization on inputs for floating point adders and multipliers.  Held off on parameterization of the outputs, since that could cause some backwards incompatible problems when we upgrade so that outputs can be different types than the inputs (as is already the case for multipliers).

## Related Issue(s)

N/A

## Testing

Existing tests should cover, this is not a functional change

## Backwards-compatibility

No

## Documentation

No


desmonddak (approved):
These look good.

Issue #169, opened 2025-02-06
Find N'th pattern from the start/end


mkorbel1:
### Motivation

Similar to the [Find N'th bit (0 or 1) from the start/end](https://github.com/intel/rohd-hcl/blob/main/doc/components/find.md#find-nth), but instead, find an entire multi-bit pattern

### Desired solution

A component that can find a fixed-width pattern, the N'th instance of it, from the start or the end

gus-bonilla-86:
Hi, we would like to work on this issue [ROHD_Team2] [Intel Open Source Software Hackathon 2025].

mkorbel1:
Awesome, assigned!

PR #187, opened 2025-03-07
Add Find N'th Pattern component

izzianyramli:
## Description & Motivation

Similar to Find N'th bit from start/end, but FindPattern find an entire multi-bit pattern from start/end and returns it index

## Related Issue(s)

https://github.com/intel/rohd-hcl/issues/169 

## Testing

Test is included in test/find_pattern_test.dart 

## Backwards-compatibility

No

## Documentation

Documentation added in doc/components/find.md

### DIFF
--- a/doc/components/find.md
+++ b/doc/components/find.md
@@ -36,3 +36,14 @@ To find the nth occurrence just pass in `bus` along with Logic value `n` passed 
 ### Find Nth Zero
 
 To find the nth occurrence just pass in `bus` along with Logic value `n` passed in as an argument. In addition to finding occurrence of Zero (`0`), set the argument `countOne` to `false`.
+
+## Find Nth Pattern
+To find the nth pattern occurrence just pass in `bus` along with Logic pattern `n` passed in as an argument. Note that the pattern needs to be smaller than bus.
+
+### Find Nth Pattern from start
+To find the nth pattern occurrence just pass in `bus` along with Logic pattern `n` passed in as an argument. Note that the pattern needs to be smaller than bus.
+In addition to finding occurrence of pattern `n`, set the argument `start` as true.
+
+### Find Nth Pattern from end
+To find the nth pattern occurrence just pass in `bus` along with Logic pattern `n` passed in as an argument. Note that the pattern needs to be smaller than bus.
+In addition to finding occurrence of pattern `n`, set the argument `start` as false.
--- a/doc/components/find.md
+++ b/doc/components/find.md
@@ -38,12 +38,12 @@ To find the nth occurrence just pass in `bus` along with Logic value `n` passed 
 To find the nth occurrence just pass in `bus` along with Logic value `n` passed in as an argument. In addition to finding occurrence of Zero (`0`), set the argument `countOne` to `false`.
 
 ## Find Nth Pattern
-To find the nth pattern occurrence just pass in `bus` along with Logic pattern `n` passed in as an argument. Note that the pattern needs to be smaller than bus.
+To find the nth pattern occurrence just pass in `bus` along with Logic pattern `pattern` passed in as an argument. Note that the `pattern` needs to be smaller than `bus`.
 
 ### Find Nth Pattern from start
-To find the nth pattern occurrence just pass in `bus` along with Logic pattern `n` passed in as an argument. Note that the pattern needs to be smaller than bus.
-In addition to finding occurrence of pattern `n`, set the argument `start` as true.
+To find the nth pattern occurrence just pass in `bus` along with Logic pattern `pattern` passed in as an argument. Note that the `pattern` needs to be smaller than `bus`.
+In addition to finding occurrence of `pattern`, set the argument `start` as true.
 
 ### Find Nth Pattern from end
-To find the nth pattern occurrence just pass in `bus` along with Logic pattern `n` passed in as an argument. Note that the pattern needs to be smaller than bus.
-In addition to finding occurrence of pattern `n`, set the argument `start` as false.
+To find the nth pattern occurrence just pass in `bus` along with Logic pattern `pattern` passed in as an argument. Note that the `pattern` needs to be smaller than `bus`.
+In addition to finding occurrence of `pattern`, set the argument `start` as false.
--- a/doc/components/find.md
+++ b/doc/components/find.md
@@ -36,14 +36,3 @@ To find the nth occurrence just pass in `bus` along with Logic value `n` passed 
 ### Find Nth Zero
 
 To find the nth occurrence just pass in `bus` along with Logic value `n` passed in as an argument. In addition to finding occurrence of Zero (`0`), set the argument `countOne` to `false`.
-
-## Find Nth Pattern
-To find the nth pattern occurrence just pass in `bus` along with Logic pattern `pattern` passed in as an argument. Note that the `pattern` needs to be smaller than `bus`.
-
-### Find Nth Pattern from start
-To find the nth pattern occurrence just pass in `bus` along with Logic pattern `pattern` passed in as an argument. Note that the `pattern` needs to be smaller than `bus`.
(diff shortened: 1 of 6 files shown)


lib/src/find_pattern.dart
```diff
@@ -0,0 +1,95 @@
...
+        minBit = bus.width - i - pattern.width;
+        maxBit = bus.width - i;
+      }
+
+      // Check if pattern matches
+      final busVal = bus.getRange(minBit, maxBit);
+      final valCheck = busVal.eq(pattern);
+
+      // Check if pattern matches, count if found
+      count += ((valCheck.value.toInt() == 1) ? 1 : 0);
```
mkorbel1, line 75: I think you mean `count += valCheck` directly -- it's already a 1-bit signal and you want to assign it synthesizably, not based on `.toInt()`

lib/src/find_pattern.dart
```diff
@@ -0,0 +1,95 @@
...
+
+      // If count matches n, break and return index
+      if (nVal.value.toInt() == count.value.toInt()) {
```
mkorbel1, line 78: This is unfortunately not synthesizable in the way you intend.  Comparisons in hardware should be done with something like `nVal.eq(count)`.  You'll need to have all the hardware present at all times, then have a result based on hardware computations in the signals.  The `for` and `if` here are "generation time" constructs, so the generated hardware from this code will not consider properly whether the n'th one has been found yet.  Let me know if you need some more guidance on how to build this hardware.

lib/src/find_pattern.dart
```diff
@@ -0,0 +1,95 @@
...
+        addOutput('index', width: bus.width);
+        index <= Const(i, width: bus.width);
+        break;
+      }
+    }
+
+    if (generateError) {
+      // If pattern is not found, return error
+      final isError =
+          (count.value.toInt() < nVal.value.toInt() || count.value.toInt() == 0)
```
mkorbel1, line 88: Again, operate on the signals themselves.  Something like `isError = count.lt(nVal) | count.eq(0)`

test/find_pattern_test.dart
```diff
@@ -0,0 +1,119 @@
...
+//
+// 2025 March 6
+// Author: Louiz Ang Zhi Lin <louiz.zhi.lin.ang@intel.com>
+//
+
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+import 'package:test/test.dart';
+
+void main() {
```
mkorbel1, line 15: suggestion: try writing at least one test where you do multiple different values on the bus, pattern, and `n` for the same hardware.  this will help test against the type of `.toInt()` bugs that you have in your implementation

lib/src/find_pattern.dart
```diff
@@ -0,0 +1,95 @@
...
+
+import 'package:rohd/rohd.dart';
+
+/// [FindPattern] functionality
+///
+/// Takes in a [Logic] to find location of a fixed-width pattern.
+/// Outputs pin `index` contains position.
+class FindPattern extends Module {
+  /// [index] is a getter for output of FindPattern
+  Logic get index => output('index');
```
mkorbel1, line 19: It would be good to clarify in the doc comment what the index means exactly.  For example, if you have a pattern `01`, is the index the position of the `0` or the `1`?  Does it depend on the direction of `start`?

mkorbel1, line 19: What does the index output when there is an error?  Would be good to document (even if you claim "undefined behavior")

lib/src/find_pattern.dart
```diff
@@ -0,0 +1,95 @@
...
+  ///
+  /// By default, [FindPattern] will look for the first occurrence
+  /// of the pattern.
+  /// If [n] is given, [FindPattern] will find the N'th occurrence
+  /// of the pattern.
+  /// [n] starts from `0` as the first occurrence.
+  ///
+  /// Outputs pin `index` contains the position. Position starts from `0` based.
+  FindPattern(Logic bus, Logic pattern,
+      {bool start = true, Logic? n, this.generateError = false})
```
mkorbel1, line 43: suggestion:  a name of `fromStart` or something might be more clear

mkorbel1 (changes requested):

lib/src/find_pattern.dart
izzianyramli: I think I get what you mean. I will implement it similar way to how Find component was implemented by appending the result to a list and get the index value from the binary of the list

doc/components/find.md
```diff
@@ -36,3 +36,14 @@ To find the nth occurrence just pass in `bus` along with Logic value `n` passed
 ### Find Nth Zero
 
 To find the nth occurrence just pass in `bus` along with Logic value `n` passed in as an argument. In addition to finding occurrence of Zero (`0`), set the argument `countOne` to `false`.
+
+## Find Nth Pattern
+To find the nth pattern occurrence just pass in `bus` along with Logic pattern `pattern` passed in as an argument. Note that the `pattern` needs to be smaller than `bus`.
+
+### Find Nth Pattern from start
+To find the nth pattern occurrence just pass in `bus` along with Logic pattern `pattern` passed in as an argument. Note that the `pattern` needs to be smaller than `bus`.
```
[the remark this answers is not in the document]
izzianyramli: Will create a separate doc in find_pattern.md as the component description will be a bit different than as describe for Find component in find.md

test/find_pattern_test.dart
izzianyramli: > suggestion: try writing at least one test where you do multiple different values on the bus, pattern, and `n` for the same hardware. this will help test against the type of `.toInt()` bugs that you have in your implementation

Can you describe more what do you mean by multiple values for the same hardware here?

doc/components/find_pattern.md
```diff
@@ -0,0 +1,84 @@
...
+final pattern = Const(bin('1010'), width: 4);
+final findPattern = FindPattern(bus, pattern);
+expect(findPattern.index.value.toInt(), equals(4));
+```
+
+### Find Pattern from End
+
+To get the `index` location of the pattern from end of the bus (MSB), pass the Logic `bus`, Logic `pattern` and set the `fromStart` to `false`. Ensure that the `pattern` width is smaller than the `bus` width. By default, it will find the first occurrence of the `pattern` from end of the `bus`.
+
+For example, if `bus` is `11101010` and `pattern` is `11101`, the output `index` will return `40` as the first occurrence of the pattern is found at the 0th-bit of the bus from the MSB.
```
mkorbel1, line 41: typo: you have `40` instead of `0`

lib/src/find_pattern.dart
```diff
@@ -0,0 +1,102 @@
...
+
+    if (n != null) {
+      n = addInput('n', n, width: n.width);
+    }
+
+    // Initialize counter pattern occurrence to 0
+    Logic count = Const(0, width: log2Ceil(bus.width + 1));
+    final nVal = ((n ?? Const(0)) + 1).zeroExtend(count.width);
+
+    int minBit;
```
mkorbel1, line 71: nit: these variable declarations could be made inside the `for` loop

test/find_pattern_test.dart
mkorbel1: Yeah, so the hardware can exist with input signals changing, and the outputs would dynamically change.

For example, a test could look something like this:

```dart
final bus = Logic(width:8);
final pattern = Logic(width:2);

final findPattern = FindPattern(bus, pattern);

bus.put(bin('10000001'));
pattern.put(bin('01'));
expect(findPattern.index.value.toInt(), equals(0));

bus.put(bin('10001001'));
pattern.put(bin('10'));
expect(findPattern.index.value.toInt(), equals(2));
```

this way, if you accidentally have some generation-time calculations done, the hardware outputs wont dynamically update when the inputs changs, so it's a good test for the types of issues you had in your implementation previously

Issue #171, opened 2025-02-06
Floating point square root


mkorbel1:
### Motivation

Floating point units sometimes need to be able to perform a square root operation.

### Desired solution

A floating-point square root component.

lucasphillips:
As part of the 2025 OSS Hackathon, ROHD_Team1 is working on this issue.

mkorbel1:
Awesome!  Assigned!

PR #188, opened 2025-03-07
Floating Point Square Root Module with Variable Odd Mantissa and No Rounding

lucasphillips:
## Description & Motivation

Sometimes we want to be able to do square root for floating points.
This is especially true for ML hardware.
This PR addresses that by adding a Floating Point Square Root module and extending
Fixed Point to have multiplication

## Related Issue(s)

#171

## Testing

Test cases were created for the Fixed Point Square Root and Floating Point.

Fixed Square Root tested:
1) Tested case of 1.0
2) Tested case of 1.5
3) Tested case of 1.7
4) Tested case of 1.125
5) Tested case of 2.25
6) Tested case of 3.999
7) Tested case of 3.999998

Floating Point Tested:
1) Non-normalized numbers
2) Error conditions (negative numbers and negative infinity)
3) Targeted normalized numbers
4) Randomized 4Ki values within rounding error

## Backwards-compatibility

No.

## Documentation

Yes. This required changes to the Floating Point and Fixed Point docs within the HCL library. They are included within this change


lib/src/arithmetic/fixed_sqrt.dart
```diff
@@ -0,0 +1,100 @@
...
+library;
+
+import 'package:meta/meta.dart';
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+
+/// Abstract base class
+abstract class FixedPointSqrtBase extends Module {
+  /// Width of the input and output fields.
+  final int numWidth;
```
desmonddak, line 21: Why not just 'width'?

test/arithmetic/floating_point/floating_point_sqrt_test.dart
```diff
@@ -0,0 +1,200 @@
...
+      final compResult = sqrtDUT.sqrtR;
+      final compError = sqrtDUT.error;
+
+      final expResult = FloatingPointValue.populator(
+              exponentWidth: exponentWidth, mantissaWidth: mantissaWidth)
+          .ofDouble(sqrt(fv.toDouble()));
+      final expError = Const(0);
+
+      expect(compResult.floatingPointValue.withinRounding(expResult), true,
+          reason: '\t${fp.floatingPointValue} '
```
desmonddak, line 188: Here's an opportunity to use 
```
reason:   '''
${fp.floatingPointValue} 
<avoiding having to inject newlines into a string>
'''
);

File: test/arithmetic/floating_point/floating_point_sqrt_test.dart [hunk 3]
```diff
@@ -0,0 +1,200 @@
...
+  test('FP: random number sqrt', () {
+    const exponentWidth = 5;
+    const mantissaWidth = 9;
+
+    final fp = FloatingPoint(
+        exponentWidth: exponentWidth, mantissaWidth: mantissaWidth);
+
+    final sqrtDUT = FloatingPointSqrtSimple(fp);
+    final rand = Random(513);
+    for (var i = 0; i < 4096; i++) {
```
desmonddak, line 167: Can you compute the limit (4096) from eWidth and mWidth?

test/arithmetic/floating_point/floating_point_sqrt_test.dart
```diff
@@ -0,0 +1,200 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Indentifier: BSD-3-Clause
+//
+// floating_point_sqrt.dart
+// An abstract base class defining the API for floating-point square root.
+//
+// 2025 March 5
+// Authors: James Farwell <james.c.farwell@intel.com>,
```
desmonddak, line 8: nit:  Align your names to get top billing -- like to see your names clearly.

lib/src/arithmetic/floating_point/floating_point_sqrt_simple.dart
```diff
@@ -0,0 +1,102 @@
...
+  /// [sqrtR] and [error]
+  FloatingPointSqrtSimple(super.a,
+      {super.clk,
+      super.reset,
+      super.enable,
+      super.name = 'floatingpoint_square_root_simple'})
+      : super(
+            definitionName: 'FloatingPointSquareRootSimple_'
+                'E${a.exponent.width}M${a.mantissa.width}') {
+    final outputSqrt = FloatingPoint(
```
desmonddak, line 28: Noticed the 'normal' tests:  are subnormals supported?  If not, you should check and throw.

lib/src/arithmetic/floating_point/floating_point_sqrt_simple.dart
```diff
@@ -0,0 +1,102 @@
...
+    // shift exponent
+    final shiftedExp =
+        [deBiasExp[-1], deBiasExp.slice(a.exponent.width - 1, 1)].swizzle();
+
+    // check if exponent was odd
+    final isExpOdd = deBiasExp[0];
+
+    // use fixed sqrt unit
+    final aFixed = FixedPoint(signed: false, m: 3, n: a.mantissa.width);
+    aFixed <= [Const(1, width: 3), a.mantissa.getRange(0)].swizzle();
```
desmonddak, line 55: If you have a chance, in a test you can check out your naming of Verilog upon output.  You can create good names by attaching .named('aFixed');   especially after swizzles.
```
File('name.sv').writeAsStringSync(mod.generateSynth());

File: lib/src/arithmetic/floating_point/floating_point_sqrt.dart [hunk 7]
```diff
@@ -0,0 +1,83 @@
...
+  /// flops.
+  @protected
+  late final Logic? enable;
+
+  /// The value [a], named this way to allow for a local variable 'a'.
+  @protected
+  late final FpType a;
+
+  /// getter for the computed [FloatingPoint] output.
+  late final FloatingPoint sqrtR = (a.clone(name: 'sqrtR') as FpType)
```
mkorbel1, line 44: question: what's the "R" mean?  why not "sqrt"?

lib/src/arithmetic/floating_point/floating_point_sqrt.dart
```diff
@@ -0,0 +1,83 @@
...
+    ..gets(output('sqrtR'));
+
+  /// getter for the [error] output.
```
mkorbel1, line 47: I'm a little confused by this setup here with `error` and `errorSig`.  Since the `error` is already a single-bit `Logic`, there's no need for this style with an internal and external version of the signal, like there is with the "casting" of `LogicStructure`s.  Unless I'm misunderstanding, I think you can just do a single `late final Logic error = output('error');` and use it for both internal and external purposes

lib/src/arithmetic/floating_point/floating_point_sqrt.dart
```diff
@@ -0,0 +1,83 @@
...
+/// An abstract API for floating point square root.
+abstract class FloatingPointSqrt<FpType extends FloatingPoint> extends Module {
+  /// Width of the output exponent field.
+  final int exponentWidth;
+
+  /// Width of the output mantissa field.
+  final int mantissaWidth;
+
+  /// The [clk] : if a non-null clock signal is passed in, a pipestage is added
+  /// to the square root to help optimize frequency.
```
mkorbel1, line 25: I don't see the `clk` and `reset` used anywhere?  No sequentials  in here (yet?)? Did you mean to use it or is this plumbed for future upgrades?

lib/src/arithmetic/signals/fixed_point_logic.dart
```diff
@@ -123,6 +123,13 @@ class FixedPoint extends Logic {
     }
   }
 
+  /// Multiply
+  Logic fpMultiply(dynamic other) {
```
mkorbel1, line 127: I think you can either move this directly into the implementation of `operator *`, or at least make this private.  If you keep it as a separate function, why not just `_multiply` since it's already applicable to FixedPoint?

lib/src/arithmetic/signals/fixed_point_logic.dart
```diff
@@ -123,6 +123,13 @@ class FixedPoint extends Logic {
...
+    _verifyCompatible(other);
+    final product = Multiply(this, other).out;
+    return FixedPoint.of(product, signed: false, m: product.width - n, n: n);
```
mkorbel1, line 130: since `Multiply` has the same output width as inputs, `product.width - n` should be the same as `m` right?  Or is it 1 bigger due to sign bit (if signed)?

Something to consider: do we want `FixedPoint` to widen or keep the same width when multiplied by another `FixedPoint`?  To keep with API consistency, it seems reasonable to keep the same width, since `FixedPoint` is a type of `Logic`.  But do we need a way to appropriately "expand" a `FixedPoint`, sort of like a `signExtend` but that makes it the "right" shape (m and n) for an output of a multiplication?

This is probably beyond the scope of this PR, but since this change is here and you're the first user of this new `*`, thought I'd ask for thoughts :)

test/arithmetic/fixed_sqrt_test.dart
```diff
@@ -0,0 +1,110 @@
...
+import 'dart:math';
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/src/arithmetic/arithmetic.dart';
+import 'package:test/test.dart';
+
+void main() {
+  tearDown(() async {
+    await Simulator.reset();
+  });
+  test('sqrt(1)', () async {
```
mkorbel1, line 21: suggestion: since most of these tests look very similar, you could consider making a list of "test vectors" and then generate all these tests, to avoid copy-paste

test/arithmetic/fixed_sqrt_test.dart
```diff
@@ -0,0 +1,110 @@
...
+    final fixed = FixedPoint(signed: false, m: 3, n: 23);
+    final dut = FixedPointSqrt(fixed);
+    await dut.build();
+    fixed.put(FixedPointValue.ofDouble(3.9999998,
+        signed: fixed.signed, m: fixed.m, n: fixed.n));
+    final fpvResult = dut.sqrtF.fixedPointValue;
+    final fpvExpected = FixedPointValue.ofDouble(sqrt(3.9999998),
+        signed: fixed.signed, m: fixed.m, n: fixed.n);
+    expect(fpvResult, fpvExpected);
+  });
```
mkorbel1, line 109: question: how did you arrive at these numbers for the tests?  are there any special corner cases to consider?  what if there's a negative input?

mkorbel1 (commented):
This is awesome work!  Got me thinking about expanding for multi-cycle or pipelined versions :)

test/arithmetic/floating_point/floating_point_sqrt_test.dart
sweeksBigBlue: Possibly? Not 100% the question here. I wanted to choose something reasonably large since we do end up discarding negative numbers

lib/src/arithmetic/floating_point/floating_point_sqrt_simple.dart
sweeksBigBlue: Good catch! I forgot that! I'll fix that this weekend though I may need pointers on how to do that

lib/src/arithmetic/floating_point/floating_point_sqrt_simple.dart
sweeksBigBlue: I wasn't sure if that was "too many names" when writing this, so tried to keep it to named structures.

lib/src/arithmetic/fixed_sqrt.dart
```diff
@@ -0,0 +1,100 @@
...
+
+  /// The value [a], named this way to allow for a local variable 'a'.
+  @protected
+  late final FixedPoint a;
+
+  /// getter for the computed output.
+  late final FixedPoint sqrtF = a.clone(name: 'sqrtF')..gets(output('sqrtF'));
```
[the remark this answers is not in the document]
sweeksBigBlue: Okay, this answers your question below! We did originally have both do sqrt. However, dart was having naming conflicts with math.sqrt() that we needed to run our tests. And honestly, sqrt didn't seem great after that. So I opted for "sqrtF" for sqrt fixed and "sqrtR" which is sqrt result. Though that one should probably be "sqrtFP".

lib/src/arithmetic/floating_point/floating_point_sqrt.dart
sweeksBigBlue: I originally was planning to use the internal error signal to also be used for logic testing throughout when we had our first rough draft of the block diagram. I think I just missed this.

lib/src/arithmetic/floating_point/floating_point_sqrt.dart
sweeksBigBlue: This is to make future work easier. In most cases, like all ALUs, you get your best bang for buck when it is pipelined. I wasn't sure how to do that yet, but I wanted to have the ground work. It also keeps it consistent with the other FloatingPoint modules that are also capable of pipeline features. So setup for future work to make it a bit easier (and once I figure out how to do it in ROHD).

lib/src/arithmetic/signals/fixed_point_logic.dart
sweeksBigBlue: So, I spent a lot of time on this last night. You do want your Fixed point to grow, or you loose A LOT of precision. Specifically, when we multiply to `N` values we end up with `2*N - 1` result. Because our n doesn't move, that means our m grows by the difference of the output width and n: `m = total_width - n`. So you could sign extend, but you will lose data. If you look at most multipliers in general your product usually grows with it. You can than leave it to the user to truncate how they will.

I'm hesitant to force it to the same size as the incoming products because you just loose so much information. Instead of opting to give precise precision to the user to then truncate as they want (or make an override to do this?)

lib/src/arithmetic/signals/fixed_point_logic.dart
sweeksBigBlue: This I will probably need some Dart education on. I like how we can use `.eq()` and `.lte()` methods or the override. That could just be my preference so not married to it. Would like to set it private though.

lib/src/arithmetic/floating_point/floating_point_sqrt_simple.dart
desmonddak: I think you have an error flag, so you would just need to assign it to ~fp.isNormal();

test/arithmetic/floating_point/floating_point_sqrt_test.dart
desmonddak: Thinking ahead to a variable width component for narrow and wider sqrts -- it would be great if the testing framework scaled.

lib/src/arithmetic/signals/fixed_point_logic.dart
sweeksBigBlue: I did my math here wrong (darn being out of practice for so long). It is still 2*N-1 but the results are `2*m` and `2*n`. So yes, we either need to sign extend or not. Going to push a fix for that.

sweeksBigBlue: So for example, if we do 3.5 * 1.25 we expect 4.375. These numbers are (11.10 *01.01) = 0100.0110. If we truncate that from the Q4.4 format back to Q2.2 format we get 00.01 which is 0.5 and inaccurate. So I DO need to fix this to be `m: 2*m, n: 2*n` but I think we should keep it widened. (Never not double your check after coding late at night so I'm glad you called this out).

lib/src/arithmetic/fixed_sqrt.dart
```diff
@@ -0,0 +1,100 @@
...
+//
+// fixed_point_sqrt.dart
+// An abstract base class defining the API for floating-point square root.
+//
+// 2025 March 3
+// Authors: James Farwell <james.c.farwell@intel.com>, Stephen
+// Weeks <stephen.weeks@intel.com>
+
+/// An abstract API for fixed point square root.
+library;
```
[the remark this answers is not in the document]
jcfarwe: Added to clear lint warning.  No known purpose, can remove.

lib/src/arithmetic/fixed_sqrt.dart
jcfarwe: Fixed

lib/src/arithmetic/fixed_sqrt.dart
```diff
@@ -0,0 +1,100 @@
...
+
+import 'package:meta/meta.dart';
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+
+/// Abstract base class
```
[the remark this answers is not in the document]
jcfarwe: Fixed

test/arithmetic/fixed_sqrt_test.dart
jcfarwe: For the purpose of floating point square root, the maximum value to be square rooted is 4.0 (exclusive).  Regarding inputting negative numbers, I have added a test to throw an exception if a signed number is attempted, and updated the tests.

test/arithmetic/fixed_sqrt_test.dart
jcfarwe: Done.

lib/src/arithmetic/floating_point/floating_point_sqrt_simple.dart
mkorbel1: it's not too many names to name signals that are generated in SV that are hard-to-understand.  it's valuable especially to name signals that are key for debug.  it's a somewhat subjective decision, but imagine a hypothetical downstream user debugging in the generated SystemVerilog

lib/src/arithmetic/fixed_sqrt.dart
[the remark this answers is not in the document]
mkorbel1: I expect you won't need to use `math.sqrt` inside the modules that are implementing them in hardware?  If you do ever need to, you can do `import 'dart:math' as math` to scope the `math` operators separately.  In usage cases of the module, from an API perspective, the scope would be unambiguous: `myFixedPointSqrt.sqrt` would never conflict with a globally visible `sqrt` function.

lib/src/arithmetic/floating_point/floating_point_sqrt.dart
mkorbel1: Ok, if we get this PR to a mergeable state except for pipelining, we should at least clearly document that the signals don't do anything yet and are plumbed for future pipelining, or pull the delta into some new branch for future work.  Ideally the users can read the doc comments to understand how the component works without looking inside to see what happens :)

lib/src/arithmetic/fixed_sqrt.dart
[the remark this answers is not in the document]
sweeksBigBlue: Fixed this issue

lib/src/arithmetic/floating_point/floating_point_sqrt.dart
sweeksBigBlue: Added docs in the simple class clarifying this isn't implemented and meant for future work

test/arithmetic/floating_point/floating_point_sqrt_test.dart
sweeksBigBlue: Agreed, removed magic numbers :)

lib/src/arithmetic/signals/fixed_point_logic.dart
sweeksBigBlue: Size was fixed to 2m.2n

lib/src/arithmetic/signals/fixed_point_logic.dart
sweeksBigBlue: Fixed to be _multiply

test/arithmetic/floating_point/floating_point_sqrt_test.dart
sweeksBigBlue: I'm not sure I know what you are asking here @desmonddak

lib/src/arithmetic/floating_point/floating_point_sqrt_simple.dart
sweeksBigBlue: I decided to add another test for the error flag on the output, and to gate the fixed point using a DeNorm flag. This is passing tests and is pushed

test/arithmetic/floating_point/floating_point_sqrt_test.dart
sweeksBigBlue: Done

lib/src/arithmetic/floating_point/floating_point_sqrt_simple.dart
sweeksBigBlue: Added name here

mkorbel1 (commented):
additional changes pushed so far look good!

lucasphillips:
> additional changes pushed so far look good!

About to push more changes from Stephen, just need to run markdown lint first.

lucasphillips:
Alright, I have pushed changes from Stephen which address all of the current comments except the second one from @desmonddak (about avoiding injecting newlines in strings).

test/arithmetic/floating_point/floating_point_sqrt_test.dart
desmonddak: You should use the Dart multi-line string to avoid having to add newlines, tabs, or extra quotes in your strings.  You can format the string exactly how you want it in your code without having to specify tabs or newlines or quotes.  Just start with three single quotes, then the lines after that are your string, then end with three single quotes.

lib/src/arithmetic/floating_point/floating_point_sqrt.dart
```diff
@@ -0,0 +1,84 @@
...
+
+  /// getter for the computed [FloatingPoint] output.
+  late final FloatingPoint sqrt = (a.clone(name: 'sqrt') as FpType)
+    ..gets(output('sqrt'));
+
+  /// getter for the [error] output.
+  late final Logic error = Logic(name: 'error')..gets(output('error'));
+
+  /// The internal error signal to pass through
+  late final Logic errorSig;
```
mkorbel1, line 54: I think this `errorSig` is not used anymore, just need to delete this?

lib/src/arithmetic/floating_point/floating_point_sqrt.dart
```diff
@@ -0,0 +1,84 @@
...
+abstract class FloatingPointSqrt<FpType extends FloatingPoint> extends Module {
+  /// Width of the output exponent field.
+  final int exponentWidth;
+
+  /// Width of the output mantissa field.
+  final int mantissaWidth;
+
+  /// The [clk] : if a non-null clock signal is passed in, a pipestage is added
+  /// to the square root to help optimize frequency.
+  /// Plummed for future work for pipelining, currently unsupported.
```
mkorbel1, line 26: nit: I think you mean "plumbed" :)

mkorbel1 (approved):
I think this looks good to me!  Great work!

desmonddak (approved):
Guys, this is beautiful work and I see you have used so many things we've asked for you to try.  A great addition to ROHD-HCL and a solid coding example.

Issue #181, opened 2025-02-25
`FixedToFloat` should accept a `FloatingPoint` for special case consideration


mkorbel1:
### Motivation

Currently, `FixedToFloat` accepts an `exponentWidth` and `mantissaWidth`, but also then calculates infinity (for example) internally based on generic floating point rules.  It would be better to use a `FloatingPoint` object, and it's `FloatingPointValue` type, to determine what "infinity" is represented as (or if it even supports infinity).

### Desired solution

Modify/add a constructor and parameterize `FixedToFloat` so that it can properly handle special cases based on the type.

Issue #157, opened 2025-01-24
Make specific `FloatingPointValue` operators return the same type


mkorbel1:
### Motivation

Operators on specific `FloatingPointValue`s (e.g. `FloatingPoint16Value`) have operators (e.g. `+`) which return generic `FloatingPointValue` objects instead of `FloatingPoint16Value`s.  It would be nice if they returned the same type.

### Desired solution

Make them return the same type.  Perhaps the `clone` could be of use to prevent code duplication?

Issue #119, opened 2024-10-31
Floating Point Value of Double Support Infinity 


soneryaldiz:
### Describe the bug

`FloatingPointValue.ofDouble` could output infinity when the input number exceeds the maximum and minimum supported values.

PR #193, opened 2025-04-22
Explicit JBit fpadder and components

desmonddak:
## Description & Motivation

Major PR with fixes for FP adder.
Multiple filesystem changes.
New components to do leading digit anticipation.
Anticipation and compound adder used in fp adder simple.
Anticipation added to fixed-to-float conversion.

## Related Issue(s)

#157 #181 #203 #119

## Testing

Tests for new components complete including both randomized wide and exhaustive narrow tests.

## Backwards-compatibility

Yes.  Moved away from conditional outputs to using a boolean and adding the conditional output. This affects the priority encoders as well as ones-complement adders that conditionally output the end-around-carry.

## Documentation

Code documentation and markdown files have been updated for new components and API changes.


desmonddak:
I updated PartialProduct to be a Module and in doing so I introduced a 2D array.  Unfortunately, yosys can't parse?

```verilog
module PartialProduct (
input logic [4:0] multiplicand,
input logic [4:0] multiplier,
output logic [8:0] row_0,
output logic [8:0] row_1,
output logic [8:0] row_2
);
logic [6:0] _shiftAmount_const_7_h1;
logic [1:0] _subset_0_1_multiples_c6;
logic encode_sign_r0;
logic [1:0] encoded_multiple_r0;
logic [1:0] encoded_multiple_r1;
logic [1:0] encoded_multiple_r2;
logic [5:0] extended_multiplier;
logic [2:0] mult_slice_r0;
logic [2:0] mult_slice_r0_xor;
logic [2:0] mult_slice_r1;
logic [2:0] mult_slice_r1_xor;
logic [2:0] mult_slice_r2;
logic [2:0] mult_slice_r2_xor;
logic [1:0][6:0] multiples;

**Comment by mkorbel1** [comment]
> Is there a reason why the confapp/pubspec.yaml points to the git ROHD-HCL instead of just '..'?
> It took me awhile to realize why I could not see components updating in confapp.
> I would think this would work as well in the installation build since the dart is there and can be compiled directly.

I think that makes sense to me, to change it to `..`

**Comment by mkorbel1** [comment]
> I updated PartialProduct to be a Module and in doing so I introduced a 2D array. Unfortunately, yosys can't parse?

Perhaps https://github.com/zachjs/sv2v may be of use?

File: doc/components/floating_point.md [hunk 1]
```diff
@@ -79,9 +79,9 @@ Again, like [FloatingPointValue](https://intel.github.io/rohd-hcl/rohd_hcl/Float
...
 
-A second [FloatingPointAdderRound](https://intel.github.io/rohd-hcl/rohd_hcl/FloatingPointAdderRound-class.html) component is available which does perform rounding.  It is based on "Delay-Optimized Implementation of IEEE Floating-Point Addition", by Peter-Michael Seidel and Guy Even, using an R-path and an N-path to process far-apart exponents and use rounding and an N-path for exponents within 2 and subtraction, which is exact.  If you pass in an optional clock, a pipe stage will be added to help optimize frequency; an optional reset and enable are can control the pipe stage.
+A second [FloatingPointAdderDualPath] component is available which does perform rounding.  It is based on "Delay-Optimized Implementation of IEEE Floating-Point Addition", by Peter-Michael Seidel and Guy Even, using an R-path and an N-path to process far-apart exponents and use rounding and an N-path for exponents within 2 and subtraction, which is exact.  If you pass in an optional clock, a pipe stage will be added to help optimize frequency; an optional reset and enable are can control the pipe stage.
```
mkorbel1, line 84: Just use ticks "`" if it's not a link

doc/components/leading_digit_anticipate.md
```diff
@@ -0,0 +1,41 @@
...
+ROHD-HCL comes with a leading-digit anticipate component which predicts the number of leading ones or zeros before the first digit change of a future sum of two inputs.  This prediction is within 1 of the actual digit change (either the `leadingDigit` or `leadingDigit` + 1 is the actual position). This can be used in parallel with an adder to avoid having to do a leading-digit detect after the addition and save delay.
+
+`LeadingDigitAnticipate` assumes a twos-complement representation of the inputs. It will provide a single set of outputs `leadingDigit`, the position of the first `1` for a positive sum and the first '0' for a negative sum, as well as `validLeadDigit` which indicates that a digit change was predicted.  
+
+Here is a sample usage:
+
+```dart
+final predictor = LeadingDigitAnticipate(a, b);
+
+final sumShift = (a + b) << predictor.leadingDigit;
```
mkorbel1, line 16: this example usage is good, but i think it's missing some context as to why you'd want to shift by `leadingDigit` amount?  i.e. maybe explain that the goal is for there to be no leading 0s?

doc/components/leading_digit_anticipate.md
```diff
@@ -0,0 +1,41 @@
...
+If you do not provide a carry, then the component outputs a pair of outputs (`leadingOneA`, `validLeadOneA`) and (`leadingOneB`, `validLeadOneB`) which you can then use to select the first set if your carry happens.
+
+The computation presume ones-complement subtraction on the inputs and if the carry from that is `1`, then that means the first operand was larger than the second, which means the first leading one computation `leadingOneA` is the correct one, otherwise use the `leadingOneB`.
+
+If you need to compute the number of leading 1s (say in a negative twos complement number), you can use a `LeadingZeroAnticipate` circuit by inverting the input.
+
+Here is a sample usage:
+
+```dart
+final predictor = LeadingZeroAnticipate(Const(0), a, aSign ^ bSign, b);
```
mkorbel1, line 36: again, some more context on this example?  why is one sign tied to 0 and the other an XOR?

lib/src/arithmetic/carry_save_mutiplier.dart
```diff
@@ -110,7 +108,7 @@ class CarrySaveMultiplier extends Multiplier {
       _pipeline.get(rCarryB),
     );
 
-    product <=
+    output('product') <=
```
mkorbel1, line 111: you can reference `product` directly instead of calling `output('product')`, right?

lib/src/arithmetic/carry_save_mutiplier.dart
```diff
@@ -53,8 +53,6 @@ class CarrySaveMultiplier extends Multiplier {
     clk = addInput('clk', clk);
```
mkorbel1, line 53: why are you overriding `Logic get product` above (can't comment directly on that line here, up on line 33) if it does the same thing as the base `Multiplier`?

lib/src/arithmetic/compound_adder.dart
```diff
@@ -123,71 +126,76 @@ class CarrySelectCompoundAdder extends CompoundAdder {
...
+          carryIn: Const(0),
+          name: 'block0_$i');
+      final fullAdder1 = adderGen(
+          a.getRange(blockStartIdx, blockStartIdx + blockWidth),
+          b.getRange(blockStartIdx, blockStartIdx + blockWidth),
+          subtractIn: subtractIn,
+          carryIn: Const(1),
+          name: 'block1_$i');
+      sum0Ary <=
+          [
```
mkorbel1, line 164: why bit-blast this instead of just using `getRange` to grab the bits?  something like 
```dart
[mux(carry0, fullAdder1.sum.getRange(1, blockWidth), fullAdder0.sum.getRange(1, blockWidth)), fullAdder0.sum[0]].swizzle().named('block_${i}_sum0Ary')
```

synthesizes the same way but simulates simpler, maybe more readable, and also makes nicer SV

lib/src/arithmetic/compound_adder.dart
```diff
@@ -224,10 +232,10 @@ class CarrySelectOnesComplementCompoundAdder extends CompoundAdder {
   /// [CarrySelectCompoundAdder.splitSelectAdderAlgorithmSingleBlock],
   CarrySelectOnesComplementCompoundAdder(super.a, super.b,
       {Adder Function(Logic, Logic, {Logic? carryIn}) adderGen =
-          ParallelPrefixAdder.new,
+          NativeAdder.new,
       Logic? subtractIn,
-      Logic? carryOut,
-      Logic? carryOutP1,
+      bool outputCarryOut = false,
```
mkorbel1, line 237: doc comment explaining what these bool arguments do?  maybe add doc comment to the outputs themselves explaining that they are only generated when those bools are set?

some other components (e.g. the FIFO) had used the the terminology of `generateNameOfPort` rather than `outputNameOfPort`.  I think I'd slightly vote for "generate" since it's not overloaded with the name of an "output", but don't feel super strongly.  Consistency would be good, at least.  What do you think?

lib/src/arithmetic/compound_adder.dart
```diff
@@ -266,19 +272,19 @@ class CarrySelectOnesComplementCompoundAdder extends CompoundAdder {
...
       final incrementer = ParallelPrefixIncr(sumPlus1);
-      sumP1 <= incrementer.out.named('sum_plus2');
+      sumP1 <=
+          mux(csadder.sumP1[-1], incrementer.out.named('sum_plus2'), sumPlus1);
     }
-    if (carryOut != null) {
+    if (outputCarryOut) {
       sum <= mux(doSubtract & csadder.sum[-1], ~csadder.sum, csadder.sum);
-      this.carryOut! <= csadder.sum[-1];
+      output('carryOut') <= csadder.sum[-1];
```
mkorbel1, line 285: prefer to use the actual getter `carryOut!` instead of `output('carryOut')`

lib/src/arithmetic/fixed_to_float.dart
```diff
@@ -29,67 +29,106 @@ class FixedToFloat extends Module {
   /// Output port [float]
   late final FloatingPoint float = _float.clone()..gets(output('float'));
 
-  /// Constructor
+  /// Constructor for fixed to floating-point conversion.
+  /// - [fixed] point num ber to convert
```
mkorbel1, line 33: typo: "num ber"

lib/src/arithmetic/fixed_to_float.dart
```diff
@@ -29,67 +29,106 @@ class FixedToFloat extends Module {
...
+  /// - [exponentWidth] desired exponent width of output
+  /// - [mantissaWidth] desired mantissa width of output
+  /// - [signed] optional, if set to false (default true)
+  /// treat input as unsigned.
   FixedToFloat(FixedPoint fixed,
       {required this.exponentWidth,
       required this.mantissaWidth,
+      bool signed = true,
+      Logic? leadingDigitPredict,
```
mkorbel1, line 42: add doc comment for what this argument does

lib/src/arithmetic/fixed_to_float.dart
```diff
@@ -29,67 +29,106 @@ class FixedToFloat extends Module {
...
       super.name = 'FixedToFloat'})
       : super(
             definitionName:
                 'Fixed${fixed.width}ToFloatE${exponentWidth}M$mantissaWidth') {
     fixed = fixed.clone()..gets(addInput('fixed', fixed, width: fixed.width));
     addOutput('float', width: _float.width) <= _float;
 
+    final localLeadingDigitPredict = (leadingDigitPredict != null)
```
mkorbel1, line 50: instead of creating a new "local" variable, why not just reassign `leadingDigitPredict = ...`?  You have no use for the original anymore during generation and reassigning will protect you from accidentally using it

lib/src/arithmetic/fixed_to_float.dart
```diff
@@ -29,67 +29,106 @@ class FixedToFloat extends Module {
...
-        .out
-        .zeroExtend(iWidth)
-        .named('jBit');
+    final maxShift = fixed.width - fixed.n + bias - 2;
+    final absValueShifted = Logic(
+        width: max(absValue.width, mantissaWidth + 2), name: 'absValueShifted');
 
+    final j = Logic(name: 'j', width: iWidth);
+    if (localLeadingDigitPredict != null) {
+      // Under construction
```
mkorbel1, line 85: is this a temporary comment?

lib/src/arithmetic/fixed_to_float.dart
```diff
@@ -29,67 +29,106 @@ class FixedToFloat extends Module {
...
+      final absValueShiftedPre = Logic(
+          width: max(absValue.width, mantissaWidth + 2),
+          name: 'absValueShiftedPre');
+
+      if (absValue.width < mantissaWidth + 2) {
+        final zeros = Const(0, width: mantissaWidth + 2 - absValue.width);
+        absValueShiftedPre <=
+            [absValue, zeros].swizzle() << localLeadingDigitPredict;
+      } else {
+        // print('here ${absValueShifted.width} $mantissaWidth');
```
mkorbel1, line 95: remove debug prints

lib/src/arithmetic/fixed_to_float.dart
```diff
@@ -107,6 +146,9 @@ class FixedToFloat extends Module {
                 .named('eShift'),
             Const(0, width: iWidth))
         .named('eRaw');
+
+    // TODO(desmonddak): this should be mantissa, but we could figure it out
```
mkorbel1, line 150: is this a functional issue or an optimization?

lib/src/arithmetic/floating_point/floating_point_adder_simple.dart
```diff
@@ -17,27 +17,31 @@ class FloatingPointAdderSimple<FpType extends FloatingPoint>
...
-  /// leading one detection (default [RecursiveModulePriorityEncoder]).
   FloatingPointAdderSimple(super.a, super.b,
       {super.clk,
       super.reset,
       super.enable,
       Adder Function(Logic a, Logic b, {Logic? carryIn}) adderGen =
           NativeAdder.new,
-      PriorityEncoder Function(Logic bitVector, {Logic? valid, String name})
-          priorityGen = RecursiveModulePriorityEncoder.new,
+      List<int> Function(int) widthGen =
```
mkorbel1, line 26: doc comment on this `widthGen`

lib/src/arithmetic/floating_point/floating_point_adder_simple.dart
```diff
@@ -17,27 +17,31 @@ class FloatingPointAdderSimple<FpType extends FloatingPoint>
...
+          CarrySelectCompoundAdder.splitSelectAdderAlgorithmSingleBlock,
       super.name = 'floatingpoint_adder_simple'})
       : super(
             definitionName: 'FloatingPointAdderSimple_'
                 'E${a.exponent.width}M${a.mantissa.width}') {
+    if (a.runtimeType != b.runtimeType) {
```
mkorbel1, line 32: this seems like a dangerous check because it doesn't consider any sort of class polymorphism.  e.g. what if someone extends an FP32 type to have some additional functionality that wouldn't break any functionality in this adder?

I think better would be to check things that definitely matter (e.g. exponent and mantissa widths) and then document the behavior clearly (e.g. "always uses input a's version of so-and-so functionality", if something like that exists).

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
```diff
@@ -67,6 +67,17 @@ class FloatingPoint extends LogicStructure {
   FloatingPointValuePopulator valuePopulator() => FloatingPointValue.populator(
       exponentWidth: exponent.width, mantissaWidth: mantissa.width);
 
+  /// Return true if the J-bit is not explicitly represented in the mantissa
```
mkorbel1, line 70: Why are these `Const` instead of `bool`?

mkorbel1 (commented):
WIP comments, sending mid-way through review to parallelize fixes :)

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
mkorbel1: also why do we have both?  can't one just always the be the opposite of the other?

lib/src/arithmetic/fixed_to_float.dart
desmonddak: Indeed.  I modified the new code to reuse variables rather than creating new ones.  In this case, we are advancing both the jbit prediction and shifting left the guess on the mantissa.  Both of these are now reused to avoid incorrect use, similar to overwriting this input with the reassignment.

My confusion was that the type changes from Logic? to Logic and I didn't recognize this is legal.

lib/src/arithmetic/fixed_to_float.dart
desmonddak: Optimization.  or'ing the rounded mantissa is happening quite late in the computation.  We can likely get an earlier indication like and'ing the absolute value, but it needs to be exhaustively proven.

desmonddak: I modified this comment and discovered that Dart linter is happy if I indent after I start a flutter TODO.  It keeps it as one single TODO instead of a first line and then a bunch of comments.

lib/src/arithmetic/floating_point/floating_point_adder_simple.dart
desmonddak: match on explicitJBit boolean.

doc/components/leading_digit_anticipate.md
desmonddak: I realized I was forcing the first input to be positive and the second to be negative to simplify.  When either can be negative, then the logic in the mux is a bit more difficult to explain.

lib/src/arithmetic/carry_save_mutiplier.dart
desmonddak: I do run into some challenges accessing the getter in a subclass, so I use the output('').  But it looks like I have found a way to eliminate these, though I think I am leveraging some aliases I need to clean up (e.g. fp_mult:  internalProduct.)

lib/src/arithmetic/carry_save_mutiplier.dart
desmonddak: Cleaned up.

lib/src/arithmetic/compound_adder.dart
desmonddak: Agreed.  generate is a better prefix.  I tried to find all my uses of this pattern and convert to generate<portName> and document.

lib/src/arithmetic/carry_save_mutiplier.dart
desmonddak: I found the problem I was facing:   the fp_simple has a getter that converts from output('sum') to a floating point.
So we can't use the getter in setting operations.

desmonddak:
I cleaned up a lot of the Configurator items.  All components should work and I use some common component configurators (adder and multiplier) as sub-knobs in larger components.  The FloatingPointMultiplier is interesting to check out as it uses both these GroupsOfKnobs as well as switching between specific implementations with one common configurator.

desmonddak:
> > I updated PartialProduct to be a Module and in doing so I introduced a 2D array. Unfortunately, yosys can't parse?
> 
> Perhaps https://github.com/zachjs/sv2v may be of use?

The new workflow uses this.  But this adds a dependency we need to document for open-source releases?

tool/gh_actions/install_sv2v.sh
```diff
@@ -0,0 +1,19 @@
...
+# Author: Desmond Kirkpatrick <desmond.a.kirkpatrick@intel.com>
+
+set -euo pipefail
+
+git clone https://github.com/zachjs/sv2v.git
+cd sv2v
+make
+cp ./bin/sv2v  /usr/local/bin
+
+# Trim if needed
```
mkorbel1, line 19: what's this "trim if needed" comment?

tool/converters/verilog_json.sh
```diff
@@ -19,7 +19,9 @@ if !(test 1 -eq $#); then
 fi;
 
 yosys_bin=/oss-cad-suite/bin/yosys
-module=`basename $1 .v`
+sv2v=/usr/local/bin/sv2v
```
mkorbel1, line 22: this is nice for the pre-generated HTMLs, but what about in the confapp schematics?

.github/workflows/general.yml
```diff
@@ -57,6 +57,9 @@ jobs:
       - name: Install CAD Suite (includes yosys)
         run: tool/gh_actions/install_opencadsuite.sh
 
+      - name: Install SystemVerilog to Verilog converter
+        run: tool/gh_actions/install_sv2v.sh
```
mkorbel1, line 61: do you want to add this to `tool/gh_codespaces/run_setup.sh` as well so the container has it too?

confapp/pubspec.yaml
```diff
@@ -12,9 +12,7 @@ dependencies:
     sdk: flutter
   rohd: ^0.6.1
   rohd_hcl:
-    git: 
-      url: https://github.com/intel/rohd-hcl
-      ref: main
+    path: ../
```
mkorbel1, line 15: while this is nice for development and hosting on rohd-hcl, i realized this makes it difficult for people to use the confapp for configuring their own designs by depending on this confapp package.  is there a reason we need to make it point to `../`?

doc/components/leading_digit_anticipate.md
```diff
@@ -0,0 +1,42 @@
...
+If you need to compute the number of leading 1s (say in a negative twos complement number), you can use a `LeadingZeroAnticipate` circuit by inverting the input.
+
+Here is a sample usage:
+
+```dart
+final predictor = LeadingZeroAnticipate(aSign, a, bSign, b);
+
+final adder = OnesComplementAdder(a, b, subtractIn: aSign ^ bSign, outputEndAroundCarry: true);
+
+final leadingZeros = mux(adder.carry | aSign, predictor.leadingOneA!, predictor.leadingOneB!);
```
mkorbel1, line 41: These examples are better, but I still think theyre a little abstract for those who don't already understand what the purpose of these are.  Maybe a couple concrete examples with actual example values would help?

lib/src/arithmetic/compound_adder.dart
```diff
@@ -35,6 +35,9 @@ abstract class CompoundAdder extends Adder {
   }
 }
 
+/// Splitting algorithm Function
+typedef Splitter = List<int> Function(int adderWidth);
```
mkorbel1, line 39: this typedef is not used anywhere, do you intend to keep it here?

lib/src/arithmetic/floating_point/floating_point_adder_simple.dart
mkorbel1: thanks for adding the doc, but what's the purpose of it still?  Why would you want to use this?  What should a function return to be useful for this?  (this applies in all doc comments talking about this argument down the stack)

lib/src/arithmetic/floating_point/floating_point_utilities.dart
```diff
@@ -49,4 +49,22 @@ abstract class FloatingPointUtilities {
 
     return (larger, smaller);
   }
+
+  /// Sort two [FloatingPoint]s and swap them if necessary so that the one
+  /// the larger exponent is the first element in the returned tuple.
+  static (FpType larger, FpType smaller)
```
mkorbel1, line 55: wondering if we should ditch this utilities class and make these into their own components with port boundaries?

lib/src/arithmetic/floating_point/floating_point_adder_simple.dart
```diff
@@ -47,82 +55,239 @@ class FloatingPointAdderSimple<FpType extends FloatingPoint>
...
+        effectiveSubtraction,
+        smallShiftedMantissa.slice(limitSize - 1, 0),
+        endAroundCarry: carry);
+
+    final lead1Prediction = predictor.leadingOne!.named('lead1Prediction');
+
+    final lead1PredictionValid =
+        predictor.validLeadOne!.named('lead1PredictionValid');
+
+    final trueSignFlopped = localFlop(trueSign).named('trueSignFlopped');
```
mkorbel1, line 155: is extra `.named` really necessary since the condFlop already names them `_flopped`?

lib/src/arithmetic/leading_digit_anticipate.dart
```diff
@@ -0,0 +1,138 @@
...
+
+    if (this.endAroundCarry == null) {
+      addOutput('leadingOneA', width: log2Ceil(fForward.width + 1)) <=
+          leadingOneA;
+      addOutput('leadingOneB', width: log2Ceil(fReverse.width + 1)) <=
+          leadingOneB;
+      addOutput('validLeadOneA') <= validLeadOneA;
+      addOutput('validLeadOneB') <= validLeadOneB;
+    } else {
+      addOutput('leadingOne', width: log2Ceil(fForward.width + 1)) <=
```
mkorbel1, line 88: It seems like this "endAroundCarry" version could almost just be a wrapper around the non-endAroundCarry version of it?  Rather than a flag with optional outputs?

lib/src/arithmetic/leading_digit_anticipate.dart
```diff
@@ -0,0 +1,138 @@
...
+          mux(this.endAroundCarry!, leadingOneA, leadingOneB);
+      addOutput('validLeadOne') <=
+          mux(this.endAroundCarry!, validLeadOneA, validLeadOneB);
+    }
+  }
+}
+
+/// Module for predicting the number of leading digits (position of first
+/// digit change) beffor the sum of the two 2s-complement numbers.
```
mkorbel1, line 97: typo: "beffor"

lib/src/arithmetic/multiply_accumulate.dart
```diff
@@ -0,0 +1,288 @@
...
+        clk: clk, reset: reset, enable: enable)
+      ..compress();
+    final adder = adderGen(compressor.add0, compressor.add1);
+    accumulate <= adder.sum.slice(a.width + b.width - 1 + 1, 0);
+  }
+}
+
+/// A MultiplyAccumulate which ignores the [c] term and applies the
+/// multiplier function
+class MultiplyOnly extends MultiplyAccumulate {
```
mkorbel1, line 241: Is this a component worth exposing in the library or just useful for testing?

lib/src/arithmetic/ones_complement_adder.dart
```diff
@@ -18,7 +18,7 @@ class OnesComplementAdder extends Adder {
   Logic get sign => output('sign');
 
   /// The end-around carry which should be added to the resulting [sum].
-  /// If the input [endAroundCarry] is not null, this value is stored there.
+  /// If the input `generateEndAroundCarry` is true, this value is stored there.
```
mkorbel1, line 21: you could make generateEndAroundCarry a `final bool generateEndAroundCarry` in the module so that you could reference it

lib/src/arithmetic/sign_magnitude_adder.dart
```diff
@@ -70,20 +87,49 @@ class SignMagnitudeAdder extends Adder {
...
+    if (generateEndAroundCarry) {
+      endAroundCarry! <=
+          (largestMagnitudeFirst ? adder.endAroundCarry! : Const(0));
     }
   }
 }
+
+/// A sign-magnitude adder implementation that uses two ones-complement
+/// adders wired in opposition to compute the magnitude and sign without
+/// using internal twos-complement addition.
```
mkorbel1, line 106: is it worth adding an explanation about why you'd use this one?  it's bigger but faster?

lib/src/component_config/config_knobs/config_knobs.dart
```diff
@@ -1,12 +1,14 @@
 // Copyright (C) 2023-2024 Intel Corporation
```
mkorbel1, line 1: update year

lib/src/component_config/components/config_fifo.dart
```diff
@@ -1,8 +1,8 @@
 // Copyright (C) 2023-2024 Intel Corporation
```
mkorbel1, line 1: year

lib/src/component_config/components/config_find.dart
```diff
@@ -1,8 +1,8 @@
 // Copyright (C) 2024 Intel Corporation
```
mkorbel1, line 1: year

lib/src/component_config/components/config_fixed_sqrt.dart
```diff
@@ -0,0 +1,36 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// config_fixed_sqrt.dart
+// Configurator for a fixed-point square root module.
+//
+// 2025 April 29
+// Author: desmond Kirkpatrick <desmond.a.kirkpatrick@intel.com>
```
mkorbel1, line 8: typo: capitalize your name

lib/src/component_config/components/config_fixed_to_float.dart
```diff
@@ -44,6 +49,8 @@ class FixedToFloatConfigurator extends Configurator {
   @override
   Module createModule() => FixedToFloat(
       FixedPoint(signed: signKnob.value, m: mKnob.value, n: nKnob.value),
-      exponentWidth: exponentWidthKnob.value,
-      mantissaWidth: mantissaWidthKnob.value);
+      FloatingPoint(
+          exponentWidth: exponentWidthKnob.value,
+          mantissaWidth: mantissaWidthKnob.value),
+      leadingDigitPredict: leadingDigitPredictionKnob.value ? Logic() : null);
```
mkorbel1, line 55: should the leading digit prediction be only 1 bit or based on the width?

lib/src/utils.dart
```diff
@@ -8,7 +8,7 @@ import 'dart:math';
 import 'package:rohd/rohd.dart';
 
 /// Computes the bit width needed to store [w] addresses.
-int log2Ceil(int w) => (log(w) / log(2)).ceil();
+int log2Ceil(int w) => (w > 0) ? (log(w) / log(2)).ceil() : 1;
```
mkorbel1, line 11: why is it good to have 1 for negative numbers?

mkorbel1 (changes requested):

test/arithmetic/floating_point/floating_point_adder_simple_test.dart
```diff
@@ -216,7 +307,11 @@ void main() {
     });
 
     test('FP: adder simple pipeline random', () async {
+      await Simulator.reset();
```
mkorbel1, line 310: better to put this in a teardown than within a test

test/arithmetic/multiplier_encoder_test.dart
```diff
@@ -431,6 +431,29 @@ void main() {
...
+
+    ppm.generateOutputs();
+    final compress = ColumnCompressorModule(ppm.rows, ppm.rowShift)..compress();
+    await ppm.build();
+    await compress.build();
+
+    File('ppg.sv').writeAsStringSync(ppm.generateSynth());
+    File('compress.sv').writeAsStringSync(compress.generateSynth());
+
+    // CompactSignExtension(ppg).signExtend();
```
mkorbel1, line 454: delete old code?

test/ready_valid_bfm_test.dart
```diff
@@ -1,8 +1,8 @@
 // Copyright (C) 2023-2024 Intel Corporation
```
mkorbel1, line 1: year

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
```diff
@@ -166,3 +184,49 @@ class FloatingPoint extends LogicStructure {
     return FloatingPoint._(signLogic, exponent, mantissa);
   }
 }
+
+/// A floating-point Logic signal with an explicit J-bit in the mantissa
+class FloatingPointExplicitJBit extends FloatingPoint {
```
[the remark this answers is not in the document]
desmonddak: Major refactoring.  I'll look at an incremental refactor by first adding the option to the constructor and seeing how easy it is to just use that condition versus overrides.  Likely it reduces a lot of overall code.

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
```diff
@@ -458,6 +472,8 @@ class FloatingPointValue implements Comparable<FloatingPointValue> {
     return true;
   }
 
+  // TODO(desmonddak): but when subnormal this fails to be accurate
```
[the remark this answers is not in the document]
desmonddak: I need to file a bug.  This is a tricky numerical definition.  I don't use the function anymore, but others may want to have it when checking for equivalence.  ULP is the unit of least precision or 'rounding  error'.

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
```diff
@@ -33,15 +33,27 @@ class FloatingPointValue implements Comparable<FloatingPointValue> {
...
   late final LogicValue mantissa;
 
   /// The [mantissa] width.
-  int get mantissaWidth => _mantissaWidth;
-  late final int _mantissaWidth;
+  int get mantissaWidth => storedMantissaWidth;
+
+  /// The stored mantissa width.
+  @protected
+  late final int storedMantissaWidth;
```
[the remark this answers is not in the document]
desmonddak: The problem is that we need this field accessible to the derived class (jbit).  But if we name it _mantissaWidth as we would normally do, it is private.  

I could learn to use 'part', or this could go away of I merge jbit into the base using a constructor option.

tool/converters/verilog_json.sh
desmonddak: Weird.  Those seem to be working.  I need to check why.  Also, that reminds me of a TODO:  d3-hwschematic had some default options passed to yosys and I don't think I followed them.

lib/src/arithmetic/floating_point/floating_point_adder_simple.dart
desmonddak: Odd.  I must have done this before fixing condFlop or was just naming happy.

lib/src/arithmetic/multiplier_components/column_compressor.dart
```diff
@@ -179,32 +183,69 @@ class ColumnCompressor {
...
+    addOutput('add0', width: maxWidth());
+    addOutput('add1', width: maxWidth());
   }
 
   /// Return the longest column length
   int longestColumn() =>
       columns.reduce((a, b) => a.length > b.length ? a : b).length;
 
+  /// Compute the maximum length of the rows
+  int maxWidth() {
```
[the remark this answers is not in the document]
desmonddak: I have an extension that uses these methods (for testing).  Unfortunately, this annotation doesn't act transitively.

lib/src/arithmetic/multiplier_components/column_compressor.dart
```diff
@@ -281,5 +322,7 @@ class ColumnCompressor {
         break;
       }
     }
+    add0 <= _extractRow(0);
```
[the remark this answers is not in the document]
desmonddak: I've fixed this by exposing this option only during testing (to be able to debug the pre-compressed product array.

lib/src/utils.dart
desmonddak: I reverted to the old code.  It must have been some bug somewhere that was passing w=0.

lib/src/arithmetic/leading_digit_anticipate.dart
desmonddak: I refactored into a base class that does a lot of the work, and then the two styles.  The challenge was that one of the classes produces MORE outputs with FEWER inputs, so its a bit complex to just have one extend the other.   So there was no clear way to order these classes.  So creating a 'Y' inheritance tree made more sense.

lib/src/arithmetic/floating_point/floating_point_utilities.dart
desmonddak: Done.

tool/gh_codespaces/run_setup.sh
```diff
@@ -23,6 +23,10 @@ tool/gh_actions/install_dependencies.sh
 # Install CAD Suite (includes yosys)
 tool/gh_actions/install_opencadsuite.sh
 
+# Install SystemVerilog to Verilog converter
+#  TODO:  Why does this fail -- devcontainer build says sv2v directory already exists!
```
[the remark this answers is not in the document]
desmonddak: Not resolved.  I'll push back in the installation, but it makes no sense as there should be no such directory.

lib/src/arithmetic/values/floating_point_values/floating_point_tf32_value.dart
```diff
@@ -36,7 +40,8 @@ class FloatingPointTF32Value extends FloatingPointValue {
   /// Creates a [FloatingPointValuePopulator], which can then be used to
   /// complete construction using population functions.
   static FloatingPointValuePopulator<FloatingPointTF32Value> populator() =>
-      FloatingPointValuePopulator(FloatingPointTF32Value.uninitialized());
+      FloatingPointValuePopulator(
+          FloatingPointTF32Value.uninitialized()..storedExplicitJBit = false);
```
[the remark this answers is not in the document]
desmonddak: Done.  I didn't recognize I could just make that the default in super.uninitialized() as well.

doc/components/floating_point.md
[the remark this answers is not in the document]
desmonddak: Added  to the floating point markdown.

doc/components/floating_point.md
```diff
@@ -20,6 +20,14 @@ Conversions from the native `double` are supported, both in rounded and unrounde
...
 
+### Explicit J-Bit
+
+In intermediate floating-point computations, it may be necessary to avoid normalization and simply store the current mantissa without shifting it left to move its leading 1 into the implicit j-bit location (no zeros before) and adjust the exponent.  We allow this by representing the j-bit explicitly in the mantissa a  leading '1', even when the floating-point is 'normal', or has a positive exponent field.  Typically, only sub-normals can have the leading j-bit stored in the mantissa.  While, in general, this can create a loss in accuracy, in some specific cases we can leverage avoiding normalization without loss of accuracy if we tailor our components to carry more precision and save the latency of normalization.
```
mkorbel1: typo: "in the mantissa *as* a leading '1'"

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
```diff
@@ -34,15 +34,28 @@ class FloatingPointValue implements Comparable<FloatingPointValue> {
...
+
+  /// The stored mantissa width.
+  @protected
   late final int _mantissaWidth;
 
+  /// The stored explicit JBit flag.
+  @protected
+  late final bool _explicitJBit;
+
+  /// Return true if the JBit is explicitly representedin the mantissa.
```
mkorbel1: typo: "representedin"

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
```diff
@@ -30,6 +30,8 @@ class FloatingPointValuePopulator<FpvType extends FloatingPointValue> {
   /// The width of the mantissa field.
   int get mantissaWidth => _unpopulated.mantissaWidth;
 
+  // bool explicitJBit() => _unpopulated.explicitJBit;
```
mkorbel1: rm old code

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
```diff
@@ -70,6 +76,33 @@ class FloatingPointValuePopulator<FpvType extends FloatingPointValue> {
       ..validate();
   }
 
+  /// Construct a [FloatingPointValue] from a
+  /// [FloatingPointValue] with a mantissa that is one smaller (implicit jbit)
+  FloatingPointValue ofFloatingPointValue(FloatingPointValue fpv) {
+    if (!explicitJBit) {
```
mkorbel1: these conditions and exceptions are confusing to me for it to be called `ofFloatingPointValue`.  Is this for conversion back and forth between explicit/implicit J-bit?  And/or from different widths?

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
```diff
@@ -383,6 +446,11 @@ class FloatingPointValuePopulator<FpvType extends FloatingPointValue> {
   /// If [normal] is true, This routine will only generate mantissas in the
   /// range of `[1,2)` and `minExponent() <= exponent <= maxExponent().`
   FpvType random(Random rv, {bool normal = false}) {
+    if (explicitJBit) {
```
mkorbel1, line 493: doesn't this generate a non-explicit-jbit version of a random floating point?

same question applies across a handful of these functions (e.g. `ofDouble`)

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
```diff
@@ -206,6 +247,11 @@ class FloatingPointValuePopulator<FpvType extends FloatingPointValue> {
   FpvType ofDouble(double inDouble,
       {FloatingPointRoundingMode roundingMode =
           FloatingPointRoundingMode.roundNearestEven}) {
+    if (explicitJBit) {
```
mkorbel1, line 294: doesn't this have implicit jbit?

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
```diff
@@ -47,26 +50,43 @@ class FloatingPoint extends LogicStructure {
...
+            explicitJBit,
             name: name);
 
   /// [FloatingPoint] internal constructor.
-  FloatingPoint._(this.sign, this.exponent, this.mantissa, {super.name})
+  FloatingPoint._(this.sign, this.exponent, this.mantissa, this._explicitJBit,
+      {super.name})
       : super([mantissa, exponent, sign]);
 
+  // TODO(desmonddak): we are temporarily allowing the override of explicitJBit.
```
mkorbel1: don't understand this TODO -- this clone doesn't have an `explicitJBit` argument exposed, but all the base classes do and then don't use it?

lib/src/arithmetic/signals/floating_point_logics/floating_point_16.dart
```diff
@@ -20,7 +20,8 @@ class FloatingPoint16 extends FloatingPoint {
             exponentWidth: FloatingPoint16Value.populator().exponentWidth);
 
   @override
-  FloatingPoint16 clone({String? name}) => FloatingPoint16(name: name);
+  FloatingPoint16 clone({String? name, bool explicitJBit = false}) =>
```
mkorbel1: remove the `explicitJBit` argument?  or did we intend to have it for some reason? (applies across a bunch/all of these types)

lib/src/arithmetic/signals/floating_point_logics/floating_point_32.dart
mkorbel1, line 1: question: should we have the option of making any of these types explicitJBit as a constructor argument or should we just leave that to users if they want to create a special type like that?  im not sure either way, wondering if you already came to a conclusion here

lib/src/arithmetic/floating_point/floating_point_adder.dart
```diff
@@ -40,44 +37,60 @@ abstract class FloatingPointAdder<FpType extends FloatingPoint> extends Module {
...
+  late final FpTypeIn b;
 
-  /// getter for the computed [FloatingPoint] output.
-  late final FloatingPoint sum = (a.clone(name: 'sum') as FpType)
+  /// getter for the computed [FpTypeOut] output.
+  late final FpTypeOut sum = (outputSum.clone(name: 'int_sum') as FpTypeOut)
     ..gets(output('sum'));
 
+  /// The conditional output FloatingPoint Logic to set
+  late final FpTypeOut outputSum;
```
mkorbel1: is this intended to be the "internal" version pre-`output('sum')`?  why isn't it driving the actual output in here then?  I see it assigned in the simple adder, but not in the dualpath?  also, this should then be `@protected`?

lib/src/arithmetic/floating_point/floating_point_adder_simple.dart
```diff
@@ -12,32 +12,65 @@ import 'package:rohd/rohd.dart';
...
+    // Would prefer to use getter here for setting, but the getter
+    // translates the type from Logic to FloatingPoint.
     output('sum') <= outputSum;
 
-    final (larger, smaller) = FloatingPointUtilities.sort((super.a, super.b));
+    final swapper = FloatingPointSortByExp(a, b);
+    final larger = swapper.outA;
+    final smaller = swapper.outB;
+    // Would like to use this to swap within this component
+    // but it doesn't work. TODO(desmonddak): Fix this.
```
mkorbel1: why doesn't it work?  did you intend to fix this in this PR?

lib/src/utils.dart
```diff
@@ -8,6 +8,7 @@ import 'dart:math';
 import 'package:rohd/rohd.dart';
 
 /// Computes the bit width needed to store [w] addresses.
+// int log2Ceil(int w) => (w > 0) ? (log(w) / log(2)).ceil() : 1;
```
mkorbel1: rm old code

lib/src/arithmetic/floating_point/floating_point_utilities.dart
```diff
@@ -18,12 +19,14 @@ abstract class FloatingPointUtilities {
     final in1 = toSwap.$1.named('swapIn1_${toSwap.$1.name}');
     final in2 = toSwap.$2.named('swapIn2_${toSwap.$2.name}');
 
-    FpType clone({String? name}) => toSwap.$1.clone(name: name) as FpType;
+    FpType clone1({String? name}) => toSwap.$1.clone(name: name) as FpType;
```
mkorbel1: we've typed the `swap` with a single `FpType` type, and the outputs can't know which input they will be receiving, so it doesn't make sense for them to need different `clone` functions, right?  i.e. this function *doesn't work as intended* if the two elements in `toSwap` can't be treated interchangeably by downstream logic

similarly below

lib/src/arithmetic/floating_point/floating_point_utilities.dart
```diff
@@ -41,12 +44,153 @@ abstract class FloatingPointUtilities {
...
+        FloatingPointUtilities.sortByExp((super.a, super.b));
+    // TODO(desmonddak): I know I can assign to a field instead, but I'm
+    // struggling getting it to work with explicit
+    output('outA') <= larger;
+    output('outB') <= smaller;
+  }
+}
+
+/// [FloatingPoint] class which wraps in a Logic for the JBit.
+class FloatingPointWithJBit extends FloatingPoint {
```
mkorbel1: what's this doing in here?

lib/src/arithmetic/multiplier_components/partial_product_generator.dart
```diff
@@ -26,7 +26,7 @@ class SignBit extends Logic {
 }
 
 /// A [PartialProductArray] is a class that holds a set of partial products
-/// for manipulation by [PartialProductGeneratorBase] and [ColumnCompressor].
+/// for manipulation by [PartialProductGeneratorBase] and column Compressor.
```
mkorbel1: typo: "Compressor"

lib/src/arithmetic/compound_adder.dart
```diff
@@ -219,15 +219,21 @@ class CarrySelectOnesComplementCompoundAdder extends CompoundAdder {
   /// - [subtractIn] is an optional Logic control for subtraction.
   /// - [subtract] is a boolean control for subtraction. It must be
   /// false(default) if a [subtractIn] Logic is provided.
+  /// - [generateCarryOut] set to true will create output[carryOut] and employ
+  /// the ones-complement optimization of not adding '1' to convert back to 2s
+  /// complement during subtraction on the [sum].
+  /// - [generateCarryOutP1] set to true will create output[carryOutP!] and
```
mkorbel1: typo: "carryOutP!", also missing some spaces in these lines

lib/src/arithmetic/ones_complement_adder.dart
mkorbel1: now that you added, you can do `[generateEndAroundCarry]` in the comment

tool/gh_codespaces/run_setup.sh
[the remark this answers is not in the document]
mkorbel1: remove the TODO then?

mkorbel1 (commented):
Looking very close!

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
desmonddak: No.  When you ask for an explicit-jbit version of a number, you call the routine with explicitJBit set.  Upon discovering this condition, these routines recurse one level and create a non-explicitJBit version with one bit smaller mantissa, and then convert it to an explicit form (which shifts the mantissa and prepends with a 1 or 0 depending on normality.

I renamed the converter ofFloatingPointValue to be toExplicitJBitFloatingPointValue, which helps understanding.

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
desmonddak: No.  You can construct a populator which has the explicitJBit set.
When you call any routine on it, it checks if this bit is true.  If so, it recurses into the same routine, but with a new populator using mantissa.width-1 and turns off the bit.  Then it converts this fpv back to an explicit form, shifting the mantissa right so the jbit is exposed.

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
desmonddak: It used to.  I was making things work with a special clone.  Now it is a 'straight clone' that the comment referred to as opposed to passing in the explicitJBit request.  It was one place I could intercept all calls before I had to add this boolean lots of places.

I removed the comment as it no longer applies.

lib/src/arithmetic/signals/floating_point_logics/floating_point_16.dart
desmonddak: Old trial code when I merged FloatingPoint and FloatingPointExplicitJBit

lib/src/arithmetic/signals/floating_point_logics/floating_point_32.dart
desmonddak: Special type.  An BF16 is an BF16, no variation.  I think a derived type would be best.  That way when call something a BF16, it is well-known exactly what it is.

lib/src/arithmetic/floating_point/floating_point_adder.dart
desmonddak: I was shadowing it in the dualpath.  I removed that declaration and used this one in the base class.  I'm following this pattern in the multiplier.  I will them consistent for now, perhaps we can clean them up with a better pattern.

For now I am calling them internalProduct and internalSum and they have a clone to produce them as outputs, all done in the base class.  This is a bit funky as these are generated from a clone of either the conditional input to specify a wider output, or from one of the inputs.  But I now realize that if I don't specify an output, but I specify two inputs that are different, I need to do something special (e.g. in the jbit, I allow a,b to be different.  If I don't specify an output, I need to set the explicitJBit field, not just clone a.  (around line 85 of floating_point_adder.dart'

lib/src/arithmetic/floating_point/floating_point_adder_simple.dart
desmonddak: We met on this issue and I have a resolution to use the swap signal internally to also swap meta-data.

lib/src/arithmetic/floating_point/floating_point_utilities.dart
desmonddak: This was the larger class to swap using the fp utilities to carry the jbit flag with the swap.  This goes away.

lib/src/arithmetic/floating_point/floating_point_utilities.dart
desmonddak: This was a weak attempt.  It now functions with the metaData form.

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
desmonddak: Renamed for clarity and updated documentation.  It is used to convert an implicit to an explicit 1 bit wider in the mantissa.

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
```diff
@@ -56,6 +59,7 @@ class FloatingPointValuePopulator<FpvType extends FloatingPointValue> {
     required LogicValue sign,
     required LogicValue exponent,
     required LogicValue mantissa,
+    bool explicitJBit = false,
```
mkorbel1: this `explicitJBit` argument is not connected to anything.  did you mean to connect it to the `_unpopulated`?

also, is is a sign of some missing testing somewhere that nothing caught this?

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
```diff
@@ -206,6 +245,11 @@ class FloatingPointValuePopulator<FpvType extends FloatingPointValue> {
   FpvType ofDouble(double inDouble,
       {FloatingPointRoundingMode roundingMode =
           FloatingPointRoundingMode.roundNearestEven}) {
+    if (explicitJBit) {
+      return toExplicitJBitFloatingPointValue(FloatingPointValue.populator(
```
mkorbel1: Shouldn't this be using `_unpopulated.clonePopulator()` instead of the base class one in case some type has its own populator?

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
```diff
@@ -310,19 +338,101 @@ class FloatingPointValue implements Comparable<FloatingPointValue> {
...
+                    pow(2.0, mantissa.width - (explicitJBit ? 1 : 0))) *
             pow(2.0, exponent.toInt() - bias);
       }
     }
     return doubleVal;
   }
 
+  /// Convert to a [FloatingPointValue] with a mantissa that is one smaller
+  /// due to the implicit J-bit.
+  FloatingPointValue toFloatingPointValue() {
```
mkorbel1: the name and doc of this is confusing now -- is this actually `toImplicitJBitFloatingPointValue` or something?  And only reduces mantissa width if it was previously explicit?  also, this is not using the populator for potential subclasses.

should this really just be deleted and just leave `canonicalize`?

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
```diff
@@ -310,19 +338,101 @@ class FloatingPointValue implements Comparable<FloatingPointValue> {
...
+          }
+        } else {
+          return populator(
+                  exponentWidth: exponentWidth,
+                  mantissaWidth: mantissaWidth,
+                  explicitJBit: explicitJBit)
+              .nan;
+        }
+      }
+      return FloatingPointValue(
```
mkorbel1: should this be using populator instead?

lib/src/arithmetic/floating_point/floating_point_adder_singlepath.dart
```diff
@@ -0,0 +1,384 @@
...
+          'not normalizing anexplicit JBit output.');
+    }
+    if ((roundingMode != FloatingPointRoundingMode.roundNearestEven) &&
+        (roundingMode != FloatingPointRoundingMode.truncate)) {
+      throw RohdHclException('FloatingPointAdderSinglePath only supports '
+          'roundNearestEven (default) and truncate).');
+    }
+    // Would prefer to use getter here for setting, but the getter
+    // translates the type from Logic to FloatingPoint.
+    output('sum') <= internalSum;
```
mkorbel1: this should be pulled out of both singnlepath and dualpath and into the base floatingpointadder, right?

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
desmonddak: uh oh.  This is strange because I was fixing bugs that varied based on explicit vs not.

desmonddak: ok, quick test tells me the flag is working on ofString().  So it may be I had tried a different way of setting explicit.  I'll track this down.

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
desmonddak: This should be toImplicitJBitFloatingPointValue.

But I clearly have some API work to do because the opposite direction is only in the populator because that is where I use it extensively to reuse the populators for implicit FP and make them work for explicit.

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
desmonddak: This is a pattern we use in each converter, right ?  (ofString, ofDouble, ofBigInts, random, etc).

lib/src/arithmetic/floating_point/floating_point_adder_singlepath.dart
desmonddak: That looks correct.  Let me try that.

desmonddak: Works.  The forces of entropy were stronger than my refactoring skills.

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
desmonddak: Ahhh I missed it in this spot.

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
desmonddak: This argument is not needed.  The FP itself contains the _explicitJBit that it uses to setup unpopulated.

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
mkorbel1: resolved to:
- make a shared _canonicalize routine
- make a canonicalizedOf populator function
- make an ofImplicit and ofExplicit populator function

don't throw if wrong type received, just copy (populate)

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
```diff
@@ -34,15 +34,28 @@ class FloatingPointValue implements Comparable<FloatingPointValue> {
...
   late final int _exponentWidth;
 
   /// The mantissa of the floating point.
   late final LogicValue mantissa;
 
   /// The [mantissa] width.
   int get mantissaWidth => _mantissaWidth;
+
+  /// The stored mantissa width.
+  @protected
```
mkorbel1: remove these `@protected`'s

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
```diff
@@ -89,7 +118,8 @@ class FloatingPointValuePopulator<FpvType extends FloatingPointValue> {
 
   /// [FloatingPointValue] constructor from a binary string representation of
   /// individual bitfields
-  FpvType ofBinaryStrings(String sign, String exponent, String mantissa) =>
+  FpvType ofBinaryStrings(String sign, String exponent, String mantissa,
+          {bool explicitJBit = false}) =>
```
mkorbel1: remove this flag

doc/components/floating_point.md
```diff
@@ -77,11 +87,9 @@ Again, like [FloatingPointValue](https://intel.github.io/rohd-hcl/rohd_hcl/Float
...
+A single path [FloatingPointAdderSinglePath](https://intel.github.io/rohd-hcl/rohd_hcl/FloatingPointAdderSinglePath-class.html) component is available which performs a straightforward processing of adding, normalizing, and rounding, but includes implicit and explicit j-bit handlingq. It takes two [FloatingPoint](https://intel.github.io/rohd-hcl/rohd_hcl/FloatingPoint-class.html) [LogicStructure](https://intel.github.io/rohd/rohd/LogicStructure-class.html)s and adds them, returning a normalized [FloatingPoint](https://intel.github.io/rohd-hcl/rohd_hcl/FloatingPoint-class.html) on the output.  An option on input is the type of ['ParallelPrefix'](https://intel.github.io/rohd-hcl/rohd_hcl/ParallelPrefix-class.html) used in the critical internal addition of the mantissas.  Another is to use a width generator for partitioning the internal compound adder to tradeoff area for reduced latency in producing the mantissa and mantissa + 1 for rounding. If you pass in an optional clock, a pipe stage will be added to help optimize frequency; an optional reset and enable are can control the pipe stage.
```
mkorbel1: typo: "handlingq"

lib/src/arithmetic/floating_point/floating_point_utilities.dart
mkorbel1: don't have two different clones (check all)

lib/src/arithmetic/floating_point/floating_point_utilities.dart
```diff
@@ -18,19 +19,21 @@ abstract class FloatingPointUtilities {
     final in1 = toSwap.$1.named('swapIn1_${toSwap.$1.name}');
```
mkorbel1: make this a general swapper, not extends FloatingPoint, and move elsewhere (utils or something), then use below for metadata

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
```diff
@@ -47,26 +50,39 @@ class FloatingPoint extends LogicStructure {
...
+      exponentWidth: exponent.width,
+      mantissaWidth: mantissa.width,
+      explicitJBit: explicitJBit);
 
+  /// Return true if the J-bit is explicitly represented in the mantissa.
+  bool get explicitJBit => _explicitJBit;
+
+  late final bool _explicitJBit;
+
+  // TODO(desmonddak): this will work incorrectly and must be fixed.
```
mkorbel1: rm todo

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
desmonddak: So the potentially cleaner way to do this is to:
.  make an ofFloatingPointValue which does the correct thing based on the input type and the populator type.
.  if they are both the same type, just clone
. Have a flag that says alwaysCanonicalize if they are the same explicit type.

desmonddak: Actually:  a new proposal:
```dart
({s,e,m}) _components(fpv, {canonicalize=true}) {}
ofFloatingPointValue(fpv, {alwaysCanonicalize=false})  --- calls _components(fpv, canonicalize: alwaysCanonicalize);
```
This way, we can convert from I to E or E to I or E to E with/without canonicalization.

desmonddak: So I resolved to use
```dart
FloatingPointValuePopulator.ofFloatingPointValue(FloatingPointValue fpv, {bool canonicalizeExplicit=false});
```
since it is one routine to do all conversions.  It calls _components, which also has a canonicalize flag.

lib/src/utils.dart
```diff
@@ -95,3 +95,19 @@ Logic condFlop(
                 resetValue: resetValue,
                 asyncReset: asyncReset)
             .named('${d.name}_flopped');
+
+/// Swap two [Logic] structures based on a conditional [swap].
+(LogicType, LogicType) logicSwap<LogicType extends Logic>(
+    Logic swap, (LogicType, LogicType) toSwap) {
```
mkorbel1: Maybe change type to `LogicType` or something since it's not floating-point specific anymore?

Suggestion: why not just call the function `swap`?

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
```diff
@@ -70,6 +72,80 @@ class FloatingPointValuePopulator<FpvType extends FloatingPointValue> {
       ..validate();
   }
 
+  /// This is a helper function to canonicalize the sign, exponent, and
+  /// mantissa values for the FloatingPointValue.
+  ({LogicValue sign, LogicValue exponent, LogicValue mantissa}) _components(
```
mkorbel1: i think this function could be `static`?

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
```diff
@@ -458,6 +521,8 @@ class FloatingPointValue implements Comparable<FloatingPointValue> {
     return true;
   }
 
+  // TODO(desmonddak): issue#203 subnormal is inaccurate.
```
mkorbel1: nit: good idea to include a link to the filed bug in the TODO

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
desmonddak: It needs to know if the current Populator has explicitJBit set.

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
desmonddak: I added the full http path?

lib/src/utils.dart
desmonddak: I had a conflict to resolve with swap, but I changed.

I don't understand the first question:  I use LogicType but you ask if I should change to LogicType.  I do this because I need a clone function.

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
mkorbel1: looks good, usually comes at the end but no big deal

https://dart.dev/tools/linter-rules/flutter_style_todos

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
mkorbel1: looks like it doesnt?  it only uses `this.exponentWidth` right now?  should it be using `fpv.exponentWidth` instead?

lib/src/utils.dart
mkorbel1: what's the conflict with `swap`?

yes i was looking at an old version or something, `LogicType` looks good now

mkorbel1: oh looks like you fixed while i was typing :)

test/arithmetic/floating_point/floating_point_adder_singlepath_test.dart
```diff
@@ -871,7 +871,7 @@ void main() {
     test(
         'FP: simple adder with mixed explicit/implicit j-bit IO '
         'widening exhaustive', () {
-      for (final outMantissaWidth in [3, 4, 5, 6, 7, 8, 9, 15]) {
+      for (final outMantissaWidth in [3, 4, 5, 6, 7, 8, 9]) {
```
mkorbel1, line 874: just curious: why remove 15?

mkorbel1 (approved):
I think this is ready, great work!

test/arithmetic/floating_point/floating_point_adder_singlepath_test.dart
desmonddak: Tests are taking a long time and 9 is wide enough to cross a significant threshold (mantissa + jbit) * 2 = 8.

PR #197, opened 2025-05-16
Fixing bug in CsrTop frontdoor reads

kimmeljo:
## Description & Motivation

Bug fix in CsrTop for frontdoor reads where the individual block read enables are tied to the top's frontdoor **write** enable. Obviously this is incorrect but technically is masked if one performs a write and then a read immediately after because the write will also trigger a read which in some cases will allow the read data to persist on the output.

## Related Issue(s)

N/A

## Testing

Slight tweak to the unit tests because they were getting fooled due to performing a write followed by a read or a read to a register that happened to be expecting an output of 0x0 (the default read value given resets).

## Backwards-compatibility

No issues with backwards compatibility - the fix is entirely contained within the implementation internals.

## Documentation

No need for doc updates.

### DIFF
--- a/lib/src/memory/csr/csr_top.dart
+++ b/lib/src/memory/csr/csr_top.dart
@@ -238,7 +238,7 @@ class CsrTop extends Module {
               maskedFrontWrAddr
                   .eq(Const(_blocks[i].baseAddr, width: addrWidth));
       _fdReads[i].en <=
-          _frontWrite.en &
+          _frontRead.en &
               maskedFrontRdAddr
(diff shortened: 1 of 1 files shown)


mkorbel1 (commented):
Can we write some unit tests to cover this gap?

kimmeljo:
> Can we write some unit tests to cover this gap?

Shoot looks like I never committed my actual unit tests...but yes will add!

mkorbel1 (approved):
Looks great!

PR #204, opened 2025-05-22
reduction with control working

desmonddak:
## Description & Motivation

There are situations where we need control at each level of reduction.

## Related Issue(s)

None.

## Testing

We provide a muxing example that tests that mux input selects come out on the correct cycle.

## Backwards-compatibility

None.

## Documentation

Yes.  The markdown is updated as is the API documentation for using the new control input.

### DIFF
--- a/lib/src/reduction_tree.dart
+++ b/lib/src/reduction_tree.dart
@@ -78,10 +78,10 @@ class ReductionTree extends Module {
   /// segment.
   /// - [sequence] is the input sequence to be reduced using the tree of
   ///   operations.
-  /// - Logic Function(List<Logic> inputs, {String name}) [operation] is the
-  ///   operation to be performed at each node. Note that [operation] can widen
-  ///   the output. The logic function must support the operation for (2 to
-  ///   [radix]) inputs.
+  /// - Logic Function(List<Logic> inputs, {int? trueDepth, String name})
+  ///   [operation] is the operation to be performed at each node. Note that
+  ///   [operation] can widen the output. The logic function must support the
+  ///   operation for (2 to [radix]) inputs.
   /// - [radix] is the width of reduction at each node in the tree (e.g.,
   ///   binary: radix=2).
   /// - [signExtend] if true, use sign-extension to widen Logic values as needed
--- a/lib/src/reduction_tree.dart
+++ b/lib/src/reduction_tree.dart
@@ -28,9 +28,13 @@ class ReductionTree extends Module {
 
   /// Operation to be performed at each node. Note that [operation] can widen
   /// the output. The logic function must support the operation for 2 and up to
-  /// [radix] inputs.
+  /// [radix] inputs. The [depth] input is the depth of the current node in the
+  /// tree to the leaves.  For sequences that are not powers of [radix], the
+  /// depth is the maximum depth to the leaves from this node in the tree.  The
+  /// [depth] can be used to index the control Logic to change behavior at each
+  /// depth of the tree.
   final Logic Function(List<Logic> inputs,
-      {int? depth, Logic? control, String name}) operation;
+      {int depth, Logic? control, String name}) operation;
 
   /// Specified width of input to each reduction node (e.g., binary: radix=2)
   final int radix;
@@ -81,8 +85,7 @@ class ReductionTree extends Module {
   /// segment.
   /// - [sequence] is the input sequence to be reduced using the tree of
   ///   operations.
-  /// - Logic Function(List<Logic> inputs, {int? depth, String name})
-  ///   [operation] is the operation to be performed at each node. Note that
+  /// - [operation] is the operation to be performed at each node. Note that
   ///   [operation] can widen the output. The logic function must support the
   ///   operation for (2 to [radix]) inputs.
   /// - [radix] is the width of reduction at each node in the tree (e.g.,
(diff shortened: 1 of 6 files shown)


lib/src/reduction_tree.dart
```diff
@@ -26,7 +29,8 @@ class ReductionTree extends Module {
   /// Operation to be performed at each node. Note that [operation] can widen
   /// the output. The logic function must support the operation for 2 and up to
   /// [radix] inputs.
-  final Logic Function(List<Logic> inputs, {String name}) operation;
+  final Logic Function(List<Logic> inputs,
+      {int? trueDepth, Logic? control, String name}) operation;
```
mkorbel1: Why is it called `trueDepth` instead of just `depth`?

should we expose flop depth and tree depth both?

how is depth computed (relative to the wide or narrow end of the tree)?  what about non-power-of-radix trees?

lib/src/reduction_tree_generator.dart
```diff
@@ -0,0 +1,208 @@
...
+  /// The control  output of the tree computation.
+  Logic? get controlOut => _controlOut;
+
+  late final Logic? _controlOut;
+
+  /// Optional [control] input to input to operation.
+  final Logic? control;
+
+  /// The combinational depth since the last flop.
+  int get depth => _computed.depth;
```
mkorbel1, line 28: This API *should* be TOTAL depth regardless of amount of flopping

lib/src/reduction_tree_generator.dart
```diff
@@ -0,0 +1,208 @@
...
+
+  /// The flop depth of the tree from the output to the leaves.
+  int get latency => _computed.flopDepth;
+
+  /// Operation to be performed at each node. Note that [operation] can widen
+  /// the output. The logic function must support the operation for 2 and up to
+  /// [radix] inputs.
+  final Logic Function(List<Logic> inputs,
+      {int? fullDepth, Logic? control, String name}) operation;
```
mkorbel1: The `fullDepth` should be changed to `currentDepth` and have nothing to do with flops.  It should be `null` if `sequence` is not a power of `radix`

lib/src/reduction_tree_generator.dart
```diff
@@ -0,0 +1,208 @@
...
+import 'dart:math';
+
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+
+/// Reduction Tree Generator that produces a recursive tree computation
+/// inline so that the [operation] can access global signals.
+class ReductionTreeGenerator {
+  /// The final output of the tree computation.
+  Logic get out => _out;
```
mkorbel1: This could just be `late final Logic out` directly

lib/src/reduction_tree_generator.dart
```diff
@@ -0,0 +1,208 @@
...
+
+    if (seqRemaining - chunk - (branchesRemaining - 1) >= 0) {
+      return pow(radix, power).toInt();
+    }
+    return pow(radix, max(power - 1, 0)).toInt();
+  }
+
+  /// Build out the recursive tree
+  ({Logic value, int depth, int fullDepth, int flopDepth, Logic? controlOut})
+      reductionTreeRecurse(List<Logic> seq) {
```
mkorbel1: this function should be private, right?

lib/src/reduction_tree.dart
```diff
@@ -17,16 +17,20 @@ class ReductionTree extends Module {
...
-  /// The flop depth of the tree from the output to the leaves.
+  /// The number of flops from the output to the leaves.
   int get latency => _computed.flopDepth;
 
   /// Operation to be performed at each node. Note that [operation] can widen
   /// the output. The logic function must support the operation for 2 and up to
   /// [radix] inputs.
-  final Logic Function(List<Logic> inputs, {String name}) operation;
+  final Logic Function(List<Logic> inputs,
+      {int? depth, Logic? control, String name}) operation;
```
mkorbel1: `depth` should not be nullable if it's never null?  Also, the doc comment for this `operation` should be made more clear as to what `depth` means specifically.

lib/src/reduction_tree.dart
```diff
@@ -52,33 +65,41 @@ class ReductionTree extends Module {
...
+  /// segments, recursively constructing [radix] child nodes to operate on each
+  /// segment.
   /// - [sequence] is the input sequence to be reduced using the tree of
-  /// operations.
-  /// - Logic Function(List<Logic> inputs, {String name}) [operation]
-  /// is the operation to be performed at each node. Note that [operation]
-  /// can widen the output. The logic function must support the operation for
-  /// (2 to [radix]) inputs.
+  ///   operations.
+  /// - Logic Function(List<Logic> inputs, {int? depth, String name})
```
mkorbel1: this type doesn't match the type of `operation` anymore` --  no need to even put the type information here, can just reference `[operation]`

lib/src/reduction_tree.dart
```diff
@@ -52,33 +65,41 @@ class ReductionTree extends Module {
...
+  ///   [operation] is the operation to be performed at each node. Note that
+  ///   [operation] can widen the output. The logic function must support the
+  ///   operation for (2 to [radix]) inputs.
   /// - [radix] is the width of reduction at each node in the tree (e.g.,
```
mkorbel1, line 91: should we add checks about radix < 2?  I guess radix 1 is ok as long as sequence is 1?  But otherwise you'll get infinite recursion?  And <=0 makes no sense?

mkorbel1 (approved):
Looks good once comments are resolved!

PR #205, opened 2025-05-27
Optional front-door access for CSRs, plus immutable configuration refactoring

mkorbel1:
## Description & Motivation

This PR introduces the ability to have optional front-door access (read, write, or neither) on CSR containers (a new base class for things like Top and Block).  Additionally, this refactors the configuration to be immutable, which makes things a bit safer.  Validation can now occur during construction, since they do not change post-construction.

## Related Issue(s)

N/A

## Testing

Existing tests, refactored them, added some new ones for no present optional accesses.

## Backwards-compatibility

Yes (relative to mainline) -- some arguments have been added and changed.

## Documentation

Yes they are included


kimmeljo (approved):
Looks great!

kimmeljo (approved):
Looks good! Totally agree on non uniquification

PR #207, opened 2025-06-08
Implement fixed-point Fast Fourier Transform

ethanuppal:
## Description & Motivation

I made this at the start of my Intel internship to learn about ROHD components. I have been polishing it up afterward to turn it into a PR.

I added:

- A bit reversal component on `LogicArray`s
- Fixed-point complex arithmetic
- Really bad FFT with $\log(n)$ butterfly stages etc

## Backwards-compatibility

## Documentation

### DIFF
--- a/lib/src/arithmetic/signals/complex_fixed_point_logic.dart
+++ b/lib/src/arithmetic/signals/complex_fixed_point_logic.dart
@@ -42,6 +42,16 @@ class ComplexFixedPoint extends Logic {
           width: _complexFixedPointWidth(signed, integerBits, fractionalBits),
         ) {}
 
+  ComplexFixedPoint.of(Logic signal,
+      {required this.signed,
+      required this.integerBits,
+      required this.fractionalBits})
+      : super(
+            width:
+                _complexFixedPointWidth(signed, integerBits, fractionalBits)) {
+    this <= signal;
+  }
+
   static ComplexFixedPoint fromPartsUnsafe(Logic realPart, Logic imaginaryPart,
       bool signed, int integerBits, int fractionalBits) {
     final result = ComplexFixedPoint(
@@ -101,8 +111,8 @@ class ComplexFixedPoint extends Logic {
         throw RohdHclException('Value is not compatible with signal.');
       }
 
-      _realPart().put(val.realPart);
-      _imaginaryPart().put(val.imaginaryPart);
+      realPart().put(val.realPart);
+      imaginaryPart().put(val.imaginaryPart);
     } else {
       throw RohdHclException('Only ComplexFixedPointValue is allowed');
     }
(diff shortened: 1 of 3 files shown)


lib/src/arithmetic/signals/complex_fixed_point_logic.dart
```diff
@@ -0,0 +1,225 @@
+// Copyright (C) 2021-2024 Intel Corporation
```
mkorbel1, line 1: Try to keep the headers consistent and accurate:
- Make sure you update copyright years
- Most files should have some header describing what the file is, and you can put your name as the author for files you create if you wish

See other files or CONTRIBUTING.md for examples

lib/src/arithmetic/signals/complex_fixed_point_logic.dart
mkorbel1, line 1: When you add new components/files that should be publicly visible, make sure to `export` them (e.g. see `lib/src/arithmetic/signals/signals.dart`, or the appropriate one per-file)

lib/src/arithmetic/signals/complex_fixed_point_logic.dart
```diff
@@ -0,0 +1,225 @@
...
+  void _verifyCompatible(ComplexFixedPoint other) {
+    if ((signed != other.signed) |
+        (integerBits != other.integerBits) |
+        (fractionalBits != other.fractionalBits)) {
+      throw RohdHclException('Inputs are not comparable.');
+    }
+  }
+
+  Logic _realPart() =>
+      getRange(0, _fixedPointWidth(signed, integerBits, fractionalBits));
```
mkorbel1, line 94: Conceptually, the `getRange` function creates a new module which extracts the requested bits from the original signal and produces an output that results from that operation.  Assigning to that output (as you did in some functions above) doesn't make sense and you'll (hopefully) get an error or at least X.

I think a more convenient way to structure a grouping of two `FixedPoint`s would be to create a `LogicStructure` which contains two `FixedPoint`s as elements.  Then they can be individually referenced, assigned, etc. without needing to extract bits each time or manage widths yourself.

lib/src/arithmetic/signals/complex_fixed_point_logic.dart
```diff
@@ -0,0 +1,225 @@
...
+
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/src/arithmetic/signals/signals.dart';
+import 'package:rohd_hcl/src/arithmetic/values/complex_fixed_point_value.dart';
+import 'package:rohd_hcl/src/exceptions.dart';
+
+class ComplexFixedPoint extends Logic {
+  final bool signed;
+
+  final int integerBits;
```
mkorbel1, line 14: for consistency (including with https://github.com/intel/rohd-hcl/pull/208), better might be to call `*Width`

lib/src/arithmetic/signals/fixed_point_logic.dart
```diff
@@ -123,6 +123,25 @@ class FixedPoint extends Logic {
...
+      final sum = Add(getRange(0, width - 1), other.getRange(0, width - 1)).sum;
+      final difference =
+          Subtract(getRange(0, width - 1), other.getRange(0, width - 1)).sum;
+
+      condition[0] <= this[-1];
+      condition[1] <= other[-1];
+
+// mux(this[-1], mux(other[-1], ), mux())
+
+      FixedPoint.of(sum, signed: signed, m: m + 1, n: n);
```
mkorbel1, line 138: looks incomplete?

lib/src/arithmetic/values/complex_fixed_point_value.dart
```diff
@@ -0,0 +1,20 @@
+// Copyright (C) 2024-2025 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+
+import 'dart:math';
+import 'package:meta/meta.dart';
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+
+@immutable
```
mkorbel1, line 9: i think you'll want to implement `==` and `hashCode` as well

lib/src/bit_reversal.dart
```diff
@@ -0,0 +1,44 @@
+// Copyright (C) 2021-2024 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+
+int bitReverse(int value, int bits) {
```
mkorbel1, line 7: you could make this private (`_bitReverse`) if this isn't something we want to expose in the library.  Alternatively, you could use `LogicValue`'s `reversed` to do the work (which also then would cover values >64 bits as well)

ethanuppal:
Closing in favor of a fully-pipelined floating-point implementation.

Issue #184, opened 2025-03-06
`FixedPoint` should be a `LogicStructure`


mkorbel1:
### Motivation

The components of a `FixedPoint` should be independently accessible (sign, m, and n).  A good way to do this is by making it a `LogicStructure`, like `FloatingPoint` is.

Also, we should probably change variable names `m` and `n` so that it is clear they represent the widths, so that `m` and `n` can be left to represent the components.

### Desired solution

- Convert `FixedPoint` into a `LogicStructure`
- Make existing `m` and `n` into `mWidth` and `nWidth`.
- Make `m` and `n` and `sign` now point to elements of the structure

PR #208, opened 2025-06-08
FixedPoint is now a LogicStructure with a Populator

desmonddak:
## Description & Motivation

We wanted FixedPoint to be a LogicStructure and to follow the new patterns for FloatingPoint in particular the populator pattern.

## Related Issue(s)

#184

## Testing

Ran existing Fixed point tests after refactoring to change both FixedPoint and FixedPointValue.

## Backwards-compatibility

No.

## Documentation

These are internals.  There may be a reference in the documentation about FixedPoint being a Logic which will need updating.

### DIFF
--- a/lib/src/arithmetic/signals/fixed_point_logic.dart
+++ b/lib/src/arithmetic/signals/fixed_point_logic.dart
@@ -7,61 +7,93 @@
 // 2024 October 24
 // Author: Soner Yaldiz <soner.yaldiz@intel.com>
 
-import 'dart:math';
+import 'package:meta/meta.dart';
 import 'package:rohd/rohd.dart';
 import 'package:rohd_hcl/rohd_hcl.dart';
 
 /// A representation of (un)signed fixed-point logic following
 /// Q notation (Qm.n format) as introduced by
 /// (Texas Instruments)[https://www.ti.com/lit/ug/spru565b/spru565b.pdf].
-class FixedPoint extends Logic {
-  /// [signed] indicates whether the representation is signed.
-  final bool signed;
+class FixedPoint extends LogicStructure {
+  /// The integer part of the fixed-point number.
+  Logic integer;
 
-  /// [m] is the number of bits reserved for the integer part.
-  final int m;
+  /// The fractional part of the fixed-point number.
+  Logic fraction;
 
-  /// [n] is the number of bits reserved for the fractional part.
-  final int n;
+  /// Return true if signed
+  bool get signed => _signed;
 
-  static int _fixedPointWidth(bool s, int a, int b) => s ? 1 + a + b : a + b;
+  late final bool _signed;
+
+  /// [mWidth] is the number of bits reserved for the integer part.
+  int get mWidth => integer.width - (signed ? 1 : 0);
+
+  /// [nWidth] is the number of bits reserved for the fractional part.
+  int get nWidth => fraction.width;
+
+  /// Utility to keep track of the Logic structure name by attaching it
+  /// to the Logic signal name in the output Verilog.
+  static String _nameJoin(String? structName, String signalName) {
+    if (structName == null) {
+      return signalName;
+    }
+    return '${structName}_$signalName';
+  }
 
   /// Constructs a [FixedPoint] signal.
   FixedPoint(
-      {required this.signed,
-      required this.m,
-      required this.n,
-      super.name,
-      super.naming})
-      : super(width: _fixedPointWidth(signed, m, n)) {
-    if ((m < 0) | (n < 0)) {
-      throw RohdHclException('m and n must be non-negative');
-    }
-    if (max(m, n) == 0) {
-      throw RohdHclException('either m or n must be greater than zero');
-    }
-  }
+      {required int mWidth,
+      required int nWidth,
+      bool signed = true,
+      String? name})
+      : this._(
+            Logic(
+                width: mWidth + (signed ? 1 : 0),
+                name: _nameJoin(name, 'integer'),
+                naming: Naming.mergeable),
+            Logic(
+                width: nWidth,
+                name: _nameJoin(name, 'fraction'),
+                naming: Naming.mergeable),
+            signed,
+            name: name);
+
+  /// [FixedPoint] internal constructor.
+  FixedPoint._(this.integer, this.fraction, this._signed, {super.name})
+      : super([fraction, integer]);
 
   /// Retrieve the [FixedPointValue] of this [FixedPoint] logical signal.
-  FixedPointValue get fixedPointValue =>
-      FixedPointValue(value: value, signed: signed, m: m, n: n);
+  FixedPointValue get fixedPointValue => valuePopulator().ofFixedPoint(this);
+
+  /// A [FixedPointValuePopulator] for values associated with this
+  /// [FloatingPoint] type.
+  @mustBeOverridden
+  FixedPointValuePopulator valuePopulator() =>
+      FixedPointValue.populator(mWidth: mWidth, nWidth: nWidth, signed: signed);
 
   /// Clone for I/O ports.
   @override
-  FixedPoint clone({String? name}) => FixedPoint(signed: signed, m: m, n: n);
+  FixedPoint clone({String? name}) =>
+      FixedPoint(signed: signed, mWidth: mWidth, nWidth: nWidth);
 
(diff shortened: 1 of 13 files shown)


lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -12,70 +12,70 @@ import 'dart:math';
...
 @immutable
 class FixedPointValue implements Comparable<FixedPointValue> {
   /// The fixed point value bit storage in two's complement.
-  late final LogicValue value;
+  late final LogicValue value = [integer, fraction].swizzle();
 
-  /// [signed] indicates whether the representation is signed.
-  final bool signed;
+  /// The integer valuue portion.
+  late final LogicValue integer;
```
mkorbel1, line 26: I like these called `integer` and `fraction` I think.  How about we call the widths `integerWidth` and `fractionWidth` as well for consistency?  Similarly for `FixedPointLogic`

lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -12,70 +12,70 @@ import 'dart:math';
...
 
-  /// [n] is the number of bits reserved for the fractional part.
-  final int n;
+  /// [nWidth] is the number of bits reserved for the fractional part.
+  late final int nWidth;
+
+  /// [signed] indicates whether the representation is signed.
+  bool get signed => _signed;
+
+  late final bool _signed;
```
mkorbel1: why a private version?  why not just `late final bool signed`?

lib/src/arithmetic/values/fixed_point_populator.dart
```diff
@@ -0,0 +1,136 @@
...
+          'FixedPointValue: $m, $n');
+    }
+    final integerValue = BigInt.from(val * pow(2, n));
+    final w = signed ? 1 + m + n : m + n;
+    final v = LogicValue.ofBigInt(integerValue, w);
+    return ofLogicValue(v);
+  }
+
+  /// Constructs [FixedPointValue] from a Dart [double] without rounding.
+  FpvType ofDoubleUnrounded(double val) {
```
mkorbel1: should we mark this as `@internal` like we did for floating?

lib/src/arithmetic/values/fixed_point_populator.dart
```diff
@@ -0,0 +1,136 @@
...
+    }
+    final integerValue = BigInt.from(val * pow(2, n + 1));
+    final w = signed ? 1 + m + n : m + n;
+    final v = LogicValue.ofBigInt(integerValue >> 1, w);
+    return ofLogicValue(v);
+  }
+
+  /// Constructs a [FixedPointValue] from another [FixedPointValue] with
+  /// by widening the integer and/or fraction widths.
+  FpvType widen(FpvType fpv) {
```
mkorbel1: thought: what if we had a flag in `ofLogicValue` to allow widening rather than a separate function?

lib/src/arithmetic/signals/fixed_point_logic.dart
```diff
@@ -7,61 +7,93 @@
...
+  /// Constructs a [FixedPoint] signal.
+  FixedPoint(
+      {required int mWidth,
+      required int nWidth,
+      bool signed = true,
+      String? name})
+      : this._(
+            Logic(
+                width: mWidth + (signed ? 1 : 0),
+                name: _nameJoin(name, 'integer'),
```
mkorbel1: is this a good pattern to prepend names of signals inside of structs with the struct name?  this sort of feels like we should more generally solve it as part of ROHD rather than expect everyone making structures to remember to prepend a struct name on the signals.

lib/src/arithmetic/signals/fixed_point_logic.dart
```diff
@@ -7,61 +7,93 @@
...
 
-  /// [m] is the number of bits reserved for the integer part.
-  final int m;
+  /// The fractional part of the fixed-point number.
+  Logic fraction;
 
-  /// [n] is the number of bits reserved for the fractional part.
-  final int n;
+  /// Return true if signed
+  bool get signed => _signed;
```
mkorbel1: why not just `final bool signed`?

lib/src/arithmetic/signals/fixed_point_logic.dart
desmonddak: It's not a great pattern:  it seems like we need something in the LogicStructure constructor to name the fields.

```dart
LogicStructure(Iterable<Logic> elements, {String? name, Iterable<String>? elementNames})
````
But the problem is that elements can be LogicStructures as well, and so we would need an equivalent naming structure to match the LogicStructure hierarchy.  Maybe that isn't too hard to do statically [name, [name, name, name], name] would be three fields with the middle being another LogicStructure with a set of three Logics.

Comment (mkorbel1) [reply lib/src/arithmetic/signals/fixed_point_logic.dart]: I was thinking more like:
- the `toString()` method reveals information about `parentStructure` for debug
- the SV generator can prepend names of `parentStructure` to generated signals

Thus you'd just be able to name things normally and everywhere you see the name it would be apparent where it was located.  Any problems with that that I'm missing?

File: lib/src/arithmetic/signals/fixed_point_logic.dart [hunk 7]
```diff
@@ -7,61 +7,92 @@
...
-      super.naming})
-      : super(width: _fixedPointWidth(signed, m, n)) {
-    if ((m < 0) | (n < 0)) {
-      throw RohdHclException('m and n must be non-negative');
-    }
-    if (max(m, n) == 0) {
-      throw RohdHclException('either m or n must be greater than zero');
-    }
-  }
+      {required int mWidth,
```
mkorbel1: we should fix these to `integerWidth` and `fractionWidth` as well

lib/src/arithmetic/signals/fixed_point_logic.dart
desmonddak: I think so:  the naming provided should be relative to the Structure and responsibility for prepending can be elsewhere.
(e.g., we shouldn't have to do nameJoin)

lib/src/arithmetic/values/fixed_point_populator.dart
desmonddak: Somehow I lost my reply:   the issue is that this takes a FXV and needs to know what the original 'dot' location is to do the extension.  Widening is relative to the 'dot', both directions.

desmonddak: I was thinking I should start naming variables with fxv instead of fpv to help when we work on conversion.

mkorbel1: > the issue is that this takes a FXV and needs to know what the original 'dot' location is to do the extension. Widening is relative to the 'dot', both directions.

ah, gotcha, thanks!

> I was thinking I should start naming variables with fxv instead of fpv to help when we work on conversion.

name change later isn't a big problem for positional arguments and type parameters i think, up to you when to do it if you want to

lib/src/component_config/components/config_fixed_sqrt.dart
```diff
@@ -13,22 +13,24 @@ import 'package:rohd_hcl/rohd_hcl.dart';
...
+        integerWidth: integerWidthKnob.value,
+        fractionWidth: fractionWidthKnob.value);
     return FixedPointSqrt(inp);
   }
 
   @override
   Map<String, ConfigKnob<dynamic>> get knobs => {
-        'm width': mWidthKnob,
-        'n width': nWidthKnob,
+        'm width': integerWidthKnob,
```
mkorbel1: nit: shall we update these keys as well, since they display on the configurator web app?

lib/src/component_config/components/config_fixed_to_float.dart
```diff
@@ -48,7 +48,10 @@ class FixedToFloatConfigurator extends Configurator {
 
   @override
   Module createModule() => FixedToFloat(
-      FixedPoint(signed: signKnob.value, m: mKnob.value, n: nKnob.value),
+      FixedPoint(
+          signed: signKnob.value,
+          integerWidth: mKnob.value,
```
mkorbel1: nit: these variables `mKnob` etc. could be rennamed too

mkorbel1 (approved):

Issue #199, opened 2025-05-20
Add basic operations directly on arithmetic types


mkorbel1:
### Motivation

It would be nice to have some basic capabilities for things like `FloatingPoint` and `FixedPoint` built-in at an `operator` level, so you can just compare two FPs (greater/less than/or equal to, plus, minus, etc.).

### Desired solution

Add operator overloading (relative to base `Logic`), and maybe more, for arithmetic types

Issue #113, opened 2024-10-23
Make `FloatingPointValue` operations scale beyond `double`


mkorbel1:
### Motivation

To support operations on `FloatingPointValue`s beyond 64 bits, math operations should be done without depending on `Dart`'s `double`

### Desired solution

Implement math operations without depending on `double` in `FloatingPointValue`

Issue #112, opened 2024-10-22
Add converters betwee `FloatingPointValue` and `FixedPointValue`


mkorbel1:
### Motivation

Not all numbers can fit in `double`, so converting back and forth from `double` doesn't cover all cases

### Desired solution

Add automation to convert between `FloatingPointValue` and `FixedPointValue`

### Additional details

https://github.com/intel/rohd-hcl/pull/99

mkorbel1:
Related issues from the hardware side: #116, #123

sweeksBigBlue:
Where exactly are you thinking of adding this logic? You want to go both directions as in Floating Point <--> Fixed Point? Are these only to be IEEE standard or do we want the conversion to be between two numbers based on sort of values provided?

mkorbel1:
I guess these could be just functions on the types directly?  Like within `FloatingPointValue`, a `toFixedPointValue`, and vice versa?

mkorbel1:
There probably need to be some arguments regarding rounding modes and precisions/widths

jcfarwe:
@sweeksBigBlue and I will claim this issue and work on it.

mkorbel1:
great!  thank you!

sweeksBigBlue:
Starting our work on this but it appears @desmonddak has done the fixed to float, so are we just going to focus on floating point to fixed for our solution @mkorbel1?

desmonddak:
This is the entire attempt by me and it is bug-riddled.  You could grab this in your own PR and free me from my addled attempt.

```dart
floating_point_value.dart: 
 /// Losslessly convert a [FloatingPointValue] to a [FixedPointValue].
  FixedPointValue toFixedPointValue() {
    // space for full shift (bias*2), mantissa, sign and jbit
    var fxdMantissa = [
      if (isNormal()) LogicValue.one else LogicValue.zero,
      mantissa
    ].swizzle().zeroExtend(bias * 2 + mantissaWidth + 2);
    fxdMantissa = sign == LogicValue.one
        ? ~fxdMantissa + 1
        : fxdMantissa; // make sure the sign is correct
    final shift = exponent.toInt() - 1;
    final shiftedFxdMantissa =
        exponent.toInt() > 0 ? fxdMantissa << shift : fxdMantissa;

    final fxpN = mantissaWidth + 2;
    final fxpM = fxdMantissa.width - fxpN - 1;
    return FixedPointValue(
[...]

PR #209, opened 2025-06-09
first stab at floating-point comparators

desmonddak:
## Description & Motivation

This is a baseline for making arithmetic types have basic operations like comparison and negate.  This provides these operations 
for FloatingPoint.

## Related Issue(s)

#199

## Testing

All added operands have basic testing.

## Backwards-compatibility

No

## Documentation

No.  This is an extension to the API for these types but a very natural one whose documentation is in the class itself.

### DIFF
--- a/lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
+++ b/lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
@@ -188,4 +188,91 @@ class FloatingPoint extends LogicStructure {
     final mantissa = Const(1, width: mantissaWidth);
     return FloatingPoint._(signLogic, exponent, mantissa, explicitJBit);
   }
+
+  /// Negate the [FloatingPoint].
+  factory FloatingPoint.negate(FloatingPoint input) {
+    final exponent = Logic(width: input.exponent.width)..gets(input.exponent);
+    final mantissa = Logic(width: input.mantissa.width)..gets(input.mantissa);
+    final sign = Logic()..gets(~input.sign);
+    return FloatingPoint._(
+      sign,
+      exponent,
+      mantissa,
+      input.explicitJBit,
+      name: input.name,
+    );
+  }
+
+  @override
+  FloatingPoint operator ~() => FloatingPoint.negate(this);
+
+  /// Equal
+  @override
+  Logic eq(dynamic other) {
+    if (other is! FloatingPoint) {
+      throw RohdHclException('Input must be floating point signal.');
+    }
+    return mux(isNaN | other.isNaN, Const(0), super.eq(other));
+  }
+
+  /// Not Equal
+  @override
+  Logic neq(dynamic other) {
+    if (other is! FloatingPoint) {
+      throw RohdHclException('Input must be floating point signal.');
+    }
+    return mux(isNaN | other.isNaN, Const(1), super.neq(other));
+  }
+
+  /// Less-than.
+  @override
+  Logic lt(dynamic other) {
+    if (other is! FloatingPoint) {
+      throw RohdHclException('Input must be floating point signal.');
+    }
+    if (explicitJBit || other.explicitJBit) {
+      throw RohdHclException(
+          'FloatingPoint with explicit J-bit cannot be compared.');
+    }
+    return mux(
+        isNaN | other.isNaN,
+        Const(0),
+        mux(this[-1], mux(other[-1], super.gt(other), Const(1)),
+            mux(other[-1], Const(0), super.lt(other))));
+  }
+
+  @override
+  Logic lte(dynamic other) {
+    if (other is! FloatingPoint) {
+      throw RohdHclException('Input must be floating point signal.');
+    }
+    if (explicitJBit || other.explicitJBit) {
+      throw RohdHclException(
+          'FloatingPoint with explicit J-bit cannot be compared.');
+    }
+    return mux(
+        isNaN | other.isNaN,
+        Const(0),
+        mux(this[-1], mux(other[-1], super.gte(other), Const(1)),
+            mux(other[-1], Const(0), super.lte(other))));
+  }
+
+  // For Greather-than operators, reverse the operands
+  /// Greater-than.
+  @override
+  Logic gt(dynamic other) {
+    if (other is! FloatingPoint) {
+      throw RohdHclException('Input must be floating point signal.');
+    }
+    return other.lt(this);
+  }
+
+  /// Greater-than-or-equal-to.
+  @override
+  Logic gte(dynamic other) {
+    if (other is! FloatingPoint) {
+      throw RohdHclException('Input must be floating point signal.');
+    }
+    return other.lte(this);
+  }
 }
--- a/test/arithmetic/floating_point/floating_point_test.dart
+++ b/test/arithmetic/floating_point/floating_point_test.dart
@@ -9,6 +9,8 @@
 //  Max Korbel <max.korbel@intel.com>
 //  Desmond A Kirkpatrick <desmond.a.kirkpatrick@intel.com
 
+import 'dart:math';
+
 import 'package:rohd/rohd.dart';
 import 'package:rohd_hcl/rohd_hcl.dart';
 import 'package:test/test.dart';
@@ -109,5 +111,73 @@ void main() {
     await Simulator.run();
 
     expect(checkRan, isTrue);
+  });
+
+  test('floating point comparison operators', () async {
+    const exponentWidth = 4;
+    const mantissaWidth = 4;
+    final fp1 = FloatingPoint(
+        exponentWidth: exponentWidth, mantissaWidth: mantissaWidth);
+
+    final val1 = FloatingPointValue.populator(
+            exponentWidth: exponentWidth, mantissaWidth: mantissaWidth)
+        .ofDouble(-1.23);
+    fp1.put(val1);
+    final val2 = FloatingPointValue.populator(
+            exponentWidth: exponentWidth, mantissaWidth: mantissaWidth)
+        .ofDouble(1.23);
+    expect((~fp1).floatingPointValue, equals(val2));
+  });
+
+  test('floating point comparison operators', () async {
+    const exponentWidth = 4;
+    const mantissaWidth = 4;
+    final fp1 = FloatingPoint(
+        exponentWidth: exponentWidth, mantissaWidth: mantissaWidth);
+    final fp2 = FloatingPoint(
+        exponentWidth: exponentWidth, mantissaWidth: mantissaWidth);
+
+    final val1 = FloatingPointValue.populator(
+            exponentWidth: exponentWidth, mantissaWidth: mantissaWidth)
+        .ofDouble(-1.23);
+    final val2 = FloatingPointValue.populator(
+            exponentWidth: exponentWidth, mantissaWidth: mantissaWidth)
+        .ofDouble(-3.45);
+
+    fp1.put(val1);
+    fp2.put(val2);
+
+    final rv = Random(71);
+
+    for (var iter = 0; iter < 500; iter++) {
+      final val1 = FloatingPointValue.populator(
+              exponentWidth: exponentWidth, mantissaWidth: mantissaWidth)
+          .random(rv);
+      final val2 = FloatingPointValue.populator(
+              exponentWidth: exponentWidth, mantissaWidth: mantissaWidth)
+          .random(rv);
+      fp1.put(val1);
+      fp2.put(val1);
+      final compareEq = fp1.eq(fp2); // This will use Logic.eq()
+      expect(compareEq.value.toBool(), isTrue);
+      final compareLte = fp1.lte(fp2);
+      expect(compareLte.value.toBool(), isTrue);
+
+      fp2.put(val2);
+      if (val1.toDouble() < val2.toDouble()) {
+        expect(fp1.lt(fp2).value.toBool(), isTrue);
+        expect(fp1.lte(fp2).value.toBool(), isTrue);
+        expect(
+            fp1.neq(fp2).value.toBool(), isTrue); // This will use Logic.neq()
+      } else if (val1.toDouble() > val2.toDouble()) {
+        expect(fp2.lt(fp1).value.toBool(), isTrue);
+        expect(fp1.gte(fp2).value.toBool(), isTrue);
+        expect(
+            fp1.neq(fp2).value.toBool(), isTrue); // This will use Logic.neq()
+      } else {
+        // rare that the two numbers would collide but just to be safe
+        expect(fp1.eq(fp2).value.toBool(), isTrue);
+      }
+    }
   });
 }


lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
```diff
@@ -188,4 +188,91 @@ class FloatingPoint extends LogicStructure {
...
+      sign,
+      exponent,
+      mantissa,
+      input.explicitJBit,
+      name: input.name,
+    );
+  }
+
+  @override
+  FloatingPoint operator ~() => FloatingPoint.negate(this);
```
mkorbel1: this does not seem correct -- this operator should be bitwise inversion right?

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
```diff
@@ -188,4 +188,91 @@ class FloatingPoint extends LogicStructure {
     final mantissa = Const(1, width: mantissaWidth);
     return FloatingPoint._(signLogic, exponent, mantissa, explicitJBit);
   }
+
+  /// Negate the [FloatingPoint].
+  factory FloatingPoint.negate(FloatingPoint input) {
```
mkorbel1: why a factory instead of just a function or getter, or even lazy operation?  e.g.
```dart
late final FloatingPoint negated = ...
```

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
```diff
@@ -188,4 +188,91 @@ class FloatingPoint extends LogicStructure {
...
+      throw RohdHclException('Input must be floating point signal.');
+    }
+    if (explicitJBit || other.explicitJBit) {
+      throw RohdHclException(
+          'FloatingPoint with explicit J-bit cannot be compared.');
+    }
+    return mux(
+        isNaN | other.isNaN,
+        Const(0),
+        mux(this[-1], mux(other[-1], super.gt(other), Const(1)),
```
mkorbel1: why use `[-1]` instead of `sign`?

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
```diff
@@ -188,4 +188,91 @@ class FloatingPoint extends LogicStructure {
...
+  @override
+  Logic lt(dynamic other) {
+    if (other is! FloatingPoint) {
+      throw RohdHclException('Input must be floating point signal.');
+    }
+    if (explicitJBit || other.explicitJBit) {
+      throw RohdHclException(
+          'FloatingPoint with explicit J-bit cannot be compared.');
+    }
+    return mux(
```
mkorbel1, line 242: what if the two `FloatingPoint`s have different exponent/mantissa widths?

test/arithmetic/floating_point/floating_point_test.dart
```diff
@@ -110,4 +112,72 @@ void main() {
...
+        expect(
+            fp1.neq(fp2).value.toBool(), isTrue); // This will use Logic.neq()
+      } else if (val1.toDouble() > val2.toDouble()) {
+        expect(fp2.lt(fp1).value.toBool(), isTrue);
+        expect(fp1.gte(fp2).value.toBool(), isTrue);
+        expect(
+            fp1.neq(fp2).value.toBool(), isTrue); // This will use Logic.neq()
+      } else {
+        // rare that the two numbers would collide but just to be safe
+        expect(fp1.eq(fp2).value.toBool(), isTrue);
```
mkorbel1: TODO: rest of the testing for the operators?  we should definitely test equality right?

desmonddak: So LogicStructure does not support operator != or operator ==.
When I try to add, it complains in FloatingPointValue about need for non-nullable type and a hashcode.

operator != is not overridable apparently.

operator -()   -- the linter cannot decide if this should have override or not!

mkorbel1: I never played with the `-` operator, interesting.

You should *not* override the `==` operator since we want that to stay as object equality.  I meant we should be sure to test that `eq` works, and also some specific corner cases (e.g. NaN, infinities, etc.) and some of the exceptions that we're checking (mismatch widths, jbit-ness, etc.)

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
```diff
@@ -188,4 +188,72 @@ class FloatingPoint extends LogicStructure {
...
+      );
+
+  /// Negate the [FloatingPoint].
+  FloatingPoint operator -() => negate();
+
+  @override
+  Logic operator >(dynamic other) => gt(other);
+  @override
+  Logic operator >=(dynamic other) => gte(other);
+  @override
```
mkorbel1: what is this floating `@override` doing here? :)

test/arithmetic/floating_point/floating_point_test.dart
```diff
@@ -110,4 +112,131 @@ void main() {
...
+        expect(fp1.neq(fp2).value.toBool(), isFalse);
+      }
+    }
+  });
+
+  // TODO(desmonddak): convert these tests to methods
+
+  BigInt toBigInt(FloatingPointValue fpv) {
```
mkorbel1: what are these to/from bigint doing in here?

desmonddak: Trial code.  I discovered I can simply use converters to and from FixedPointValue and cover issues #112 , #113.

I didn't anticipate needing to make changes to both Fixed and Float at the same time, so this was experimental code and I'll delete.

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
desmonddak: Ugh.

test/arithmetic/floating_point/floating_point_test.dart
desmonddak: I'll be adding the corner case testing.  Infinities are always equal (modulo the sign), but NaNs are never equal.

test/arithmetic/floating_point/floating_point_test.dart
desmonddak: The direction I am going:  conversions between FP and FX, with rounding logic will allow us to validate very wide FP operations.

test/arithmetic/floating_point/floating_point_test.dart
desmonddak: Should we not do this (already in the code):

floating_point_value.dart
```dart
  @override
  bool operator ==(Object other) {
    if (other is! FloatingPointValue) {
      return false;
    }
    if ((exponent.width != other.exponent.width) |
        (mantissa.width != other.mantissa.width)) {
      return false;
    }
    if (isNaN != other.isNaN) {
      return false;
    }
    if (isAnInfinity != other.isAnInfinity) {
      return false;
    }
    if (isAnInfinity) {
      return sign == other.sign;

Comment (desmonddak) [reply test/arithmetic/floating_point/floating_point_test.dart]: Another weird case:  I thought we could not override <= or <:

fixed_point_value.dart:
```dart
  @override
  bool operator ==(Object other) {
    if (other is! FloatingPointValue) {
      return false;
    }
    if ((exponent.width != other.exponent.width) |
        (mantissa.width != other.mantissa.width)) {
      return false;
    }
    if (isNaN != other.isNaN) {
      return false;
    }
    if (isAnInfinity != other.isAnInfinity) {
      return false;
    }
    if (isAnInfinity) {
      return sign == other.sign;

Comment (mkorbel1) [reply test/arithmetic/floating_point/floating_point_test.dart]: yes we should on the `*Value`s, but not on the things implementing `Logic`.  The values are sort of like immutable software primitives, whereas signals are objects that change value over time during simulation.

File: lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart [hunk 8]
```diff
@@ -78,7 +78,7 @@ class FloatingPoint extends LogicStructure {
   /// Return true if the J-bit is explicitly represented in the mantissa.
   bool get explicitJBit => _explicitJBit;
 
-  late final bool _explicitJBit;
+  final bool _explicitJBit;
```
mkorbel1, line 81: since it's final, we could actually just skip the private version and getter, just have `final bool explicitJBit`

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
```diff
@@ -190,72 +190,63 @@ class FloatingPoint extends LogicStructure {
...
+        Logic()..gets(~sign),
+        Logic(width: exponent.width)..gets(exponent),
+        Logic(width: mantissa.width)..gets(mantissa),
+        explicitJBit,
+        name: name,
+      );
 
+  /// Negate the [FloatingPoint].
+  FloatingPoint operator -() => negate();
+  // ignore the lint warning about overriding the '-' operator.
```
mkorbel1, line 203: what lint warning?  does the analyzer flag a lint or not, because it should be failing if you don't have the right ignore syntax

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
```diff
@@ -188,4 +188,77 @@ class FloatingPoint extends LogicStructure {
     final mantissa = Const(1, width: mantissaWidth);
     return FloatingPoint._(signLogic, exponent, mantissa, explicitJBit);
   }
+
+  /// Negate the [FloatingPoint].
+  FloatingPoint negate() => FloatingPoint._(
```
mkorbel1, line 193: is there anything special to consider for NaN/infinity here?

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
```diff
@@ -492,19 +492,45 @@ class FloatingPointValue implements Comparable<FloatingPointValue> {
...
   }
 
   /// Negate operation for [FloatingPointValue].
-  FloatingPointValue negate() => clonePopulator().populate(
+  FloatingPointValue negated() => clonePopulator().populate(
       sign: sign.isZero ? LogicValue.one : LogicValue.zero,
       exponent: exponent,
       mantissa: mantissa);
 
+  // Move this to FixedPointValuePopulator.ofFloatingPointValue.
```
mkorbel1: resolve this comment?

also, wouldn't we always want a fixed width `FixedPointValue` per widths of the `FloatingPointValue` components rather than the width changing based on values?

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
```diff
@@ -205,25 +205,56 @@ class FloatingPointValuePopulator<FpvType extends FloatingPointValue> {
...
           sign: LogicValue.ofBigInt(sign ? BigInt.one : BigInt.zero, 1),
           exponent: LogicValue.ofBigInt(exponent, exponentWidth),
           mantissa: LogicValue.ofBigInt(mantissa, mantissaWidth));
 
-  /// [FloatingPointValue] constructor from a set of [int]s of the binary
-  /// representation and the size of the exponent and mantissa
+  /// [FloatingPointValue] populator from a [FixedPointValue],
+  FpvType ofFixedPointValue(FixedPointValue fxv,
+      {FloatingPointRoundingMode mode =
+          FloatingPointRoundingMode.roundNearestEven}) {
```
mkorbel1: do you want to add a check that the rounding mode is not changed until we support more?

lib/src/utils.dart
```diff
@@ -10,6 +10,19 @@ import 'package:rohd/rohd.dart';
 /// Computes the bit width needed to store [w] addresses.
 int log2Ceil(int w) => (log(w) / log(2)).ceil();
 
+/// Finds the index of the first set bit in a [BigInt] number.
+/// Returns length + 1 if there are no set bits (i.e., the number is zero).
```
mkorbel1, line 14: isn't this an infinite loop if `inNumber` is 0?

wouldn't `null` be a better return value if none is found?

also, is this a utility we want to expose in the public API?  we have other things like `Find` which can also find the `n`th instance of a 1 *or* 0, plus an error signal on the hardware side.  Maybe a more complete extension to `LogicValue` would be more valuable as a public API?

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
```diff
@@ -492,19 +492,45 @@ class FloatingPointValue implements Comparable<FloatingPointValue> {
...
+    var fxdMantissa = [
+      if (isNormal()) LogicValue.one else LogicValue.zero,
+      mantissa
+    ].swizzle().zeroExtend(bias * 2 + mantissaWidth + 2);
+    fxdMantissa = sign == LogicValue.one
+        ? ~fxdMantissa + 1
+        : fxdMantissa; // make sure the sign is correct
+    final shift = exponent.toInt() - 1;
+    final shiftedFxdMantissa =
+        exponent.toInt() > 0 ? fxdMantissa << shift : fxdMantissa;
```
sweeksBigBlue, line 519: Are we shifting out the unbiased exponent here? I would think you would need to debias the exponent but maybe I'm missing something further here?

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
desmonddak: The analyzer flags a lint error, but it doesn't fail in CI.  It is a false error, the operator works.  I don't have to have this operator.  I see a blue underline in my VSCode with the message that this operator needs an `@override` annotation
image

When I add override, it still complains, probably because LogicValue doesn't have a '-' operator:
image

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
desmonddak: Oddly, no.   The standard says nothing about having a signed NaN.  Negating infinity makes sense as does negating zero.  So I could not think of anything to check here.

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
desmonddak: The width of FixedPointValue would be fixed but very wide to make FP conversion lossless.  But that width is a function of the FP width  (bias * 2 + mantissaWidth + 2):   the full range of sliding an FP number up and down the scale of exponents plus its mantissa plus a jbit plus a sign bit.

So I thought it made sense to do a .toFixedPointValue rather than doing an .ofFloatingPointValue since we need to compute the widths dependent on the FP exponent/mantissa widths (not values).

desmonddak: Maybe I didn't understand this question, but this function is dependent on FP component widths and is written to handle that -- did you see something value-based in the method?

desmonddak: Ah, now I see.  My comment was misleading and confused.  I do NOT want to build anything that is value dependent.  I was considering it when I could not figure out just how wide I had to go to cover the full range.

lib/src/utils.dart
desmonddak: I like the idea of a LogicValue extension.  We could mirror 'Find' in the Value side.  Let me take a look.

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
desmonddak: No, its really weird:  I am using the unbiased exponent of 00...001 to be fully right-shifted mantissa and then 11....10 to be the full left-shifted mantissa stored in a really wide FixedPoint integer.  We would set n/m to be around the midpoint of this super-wide LogicStructure.

lib/src/utils.dart
desmonddak: For now I added code to match my other leading1 detectors.  For some reason I have not tripped on this bug.
It makes sense to move to Find.

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
sweeksBigBlue: Oh okay, I would for sure need that explained to me 😆

desmonddak:
I converted this PR to a draft as there is a lot to fix.

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
mkorbel1: hmm weird!  i wonder if it's an analyzer bug?

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
desmonddak: I answered poorly.  yes, we are shifting the mantissa by the unbiased exponent.  That way, the tiniest numbers are all the way to the right (from mantissa-1 to 0).  Then the largest numbers are shifted as far left as we can so the mantissa sits at the left edge down to left edge - mantissa-width.  that makes the fixed-point around 2*bias + mantissa-width wide.

PR #211, opened 2025-06-10
Add error generation to `TreeOneHotToBinary`

mkorbel1:
## Description & Motivation

The `TreeOneHotToBinary` had a limitation where it could not generate an error signal, which was annoying.  This PR adds it.

## Related Issue(s)

N/A

## Testing

Added new tests

## Backwards-compatibility

No

## Documentation

No

### DIFF
--- a/lib/src/encodings/one_hot_to_binary.dart
+++ b/lib/src/encodings/one_hot_to_binary.dart
@@ -32,14 +32,11 @@ abstract class OneHotToBinary extends Module {
       {bool generateError = false, String name = 'one_hot_to_binary'}) {
     final isSmall = onehot.width <= 8;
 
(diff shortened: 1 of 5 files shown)


desmonddak (approved):
Should we be going through other components and using this generateError condition?

In the case of leadingOne, I use the fact that it returns the length of the vector when it is 0.
But generating an error also makes sense in certain situations in addition.

mkorbel1:
> Should we be going through other components and using this generateError condition?

I think it's a nice flag where appropriate, perhaps we can even make it an interface of some sort.  I added it here specifically because I needed the error signal and it was missing for wider busses (since we select the implementation based on width).

PR #212, opened 2025-06-22
Denormal as Zero / Flush to Zero feature added to floating-point adders

desmonddak:
## Description & Motivation

This PR adds an option to the FloatingPoint / FloatingPointValue classes to treat subnormals as zero.

## Testing

Exhaustive testing of daz/ftz for inputs and outputs of both dual-path and single-path adders.

## Backwards-compatibility

No.  subNormalAsZero is an optional boolean.

## Documentation

Documentation will be added later.

### DIFF
--- a/lib/src/arithmetic/floating_point/floating_point_adder.dart
+++ b/lib/src/arithmetic/floating_point/floating_point_adder.dart
@@ -52,6 +52,7 @@ abstract class FloatingPointAdder<FpTypeIn extends FloatingPoint,
   late final FpTypeOut internalSum;
 
   /// The rounding mode to use for the adder.
+  @protected
   late final FloatingPointRoundingMode roundingMode;
 
   /// Add two floating point numbers [a] and [b], returning result in [sum].
--- a/lib/src/arithmetic/floating_point/floating_point_adder_singlepath.dart
+++ b/lib/src/arithmetic/floating_point/floating_point_adder_singlepath.dart
@@ -62,9 +62,16 @@ class FloatingPointAdderSinglePath<FpTypeIn extends FloatingPoint,
           'roundNearestEven (default) and truncate).');
     }
 
+    final fa = a.subNormalAsZero
+        ? (a.clone()..gets(mux(a.isNormal, a, Const(0, width: a.width))))
+        : a;
+    final fb = b.subNormalAsZero
+        ? (b.clone()..gets(mux(b.isNormal, b, Const(0, width: b.width))))
+        : b;
+
     final aExplicit = Const(a.explicitJBit);
     final bExplicit = Const(b.explicitJBit);
-    final swapper = FloatingPointSortByExp(a, b,
+    final swapper = FloatingPointSortByExp(fa, fb,
         metaA: aExplicit, metaB: bExplicit, name: 'sorter_${a.name}_${b.name}');
     final larger = swapper.outA;
     final smaller = swapper.outB;
@@ -353,6 +360,11 @@ class FloatingPointAdderSinglePath<FpTypeIn extends FloatingPoint,
       }
       exponentRound = exponent;
     }
+    // Handle Flush to Zero subnormal case
+    mantissaRound = (internalSum.subNormalAsZero
+        ? mux(lead1Dominates | ~exponentRound.or(),
+            Const(0, width: mantissaRound.width), mantissaRound)
+        : mantissaRound);
 
     final realIsInf =
         (isInfFlopped | exponentRound.eq(infExponent)).named('realIsInf');
--- a/lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
+++ b/lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
@@ -39,6 +39,7 @@ class FloatingPoint extends LogicStructure {
       {required int exponentWidth,
       required int mantissaWidth,
       bool explicitJBit = false,
+      bool subNormalAsZero = false,
       String? name})
       : this._(
             Logic(name: _nameJoin(name, 'sign'), naming: Naming.mergeable),
@@ -51,10 +52,12 @@ class FloatingPoint extends LogicStructure {
                 name: _nameJoin(name, 'mantissa'),
                 naming: Naming.mergeable),
             explicitJBit,
+            subNormalAsZero,
             name: name);
 
   /// [FloatingPoint] internal constructor.
   FloatingPoint._(this.sign, this.exponent, this.mantissa, this._explicitJBit,
+      this._subNormalAsZero,
       {super.name})
       : super([mantissa, exponent, sign]);
 
@@ -64,6 +67,7 @@ class FloatingPoint extends LogicStructure {
         exponentWidth: exponent.width,
         mantissaWidth: mantissa.width,
         explicitJBit: explicitJBit,
+        subNormalAsZero: subNormalAsZero,
         name: name,
       );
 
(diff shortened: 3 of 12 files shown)


.github/workflows/general.yml
```diff
@@ -157,7 +157,7 @@ jobs:
         uses: flutter-actions/setup-flutter@v2
         with:
           channel: stable
-          version: 3.24.2
+          version: 3.29.3
```
mkorbel1: we should use the same version in both places?  not 3.29 and 3.27?

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
```diff
@@ -73,13 +77,17 @@ class FloatingPoint extends LogicStructure {
...
+      explicitJBit: explicitJBit,
+      subNormalAsZero: subNormalAsZero);
 
   /// Return true if the J-bit is explicitly represented in the mantissa.
   bool get explicitJBit => _explicitJBit;
-
   late final bool _explicitJBit;
 
+  /// Return true if subnormal numbers are represented as zero.
+  bool get subNormalAsZero => _subNormalAsZero;
```
mkorbel1: Should we add a simple converter function to the `FloatingPointLogic` so that you can convert to that type?  E.g. imagine you have a "normal" FP Logic and then want to pass it into an FP Adder with SAZ on the inputs?  It seems like you could just say `toSubNormalAsZero` and it would create the `mux` and output the proper type on the resulting conversion post-mux?  This could mean the muxing becomes redundant inside an adder (or other component), but synthesis could take care of it?

Also/instead, should we expand the `clone` function to take an argument to convert it to that type so that you can easily pass in an output with everything the same (widths, etc.) except just changing the SAZ?

lib/src/arithmetic/floating_point/floating_point_adder.dart
```diff
@@ -52,6 +52,7 @@ abstract class FloatingPointAdder<FpTypeIn extends FloatingPoint,
   late final FpTypeOut internalSum;
 
   /// The rounding mode to use for the adder.
+  @protected
```
mkorbel1: why protected?

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
```diff
@@ -73,13 +77,17 @@ class FloatingPoint extends LogicStructure {
...
+  late final bool _subNormalAsZero;
```
mkorbel1: why not `final bool subNormalAsZero`?  (same for `explicitJBit`)

lib/src/component_config/components/config_floating_point_adder.dart
```diff
@@ -12,6 +12,29 @@ import 'dart:collection';
...
+/// Set of knobs for subnormal treatment.
+class SubNormalAsZeroSelecKnob extends GroupOfKnobs {
+  /// Treat input 'a' as zero when subnormal.
+  final ToggleConfigKnob denormalAsZeroA = ToggleConfigKnob(value: false);
+
+  /// Treat input 'b' as zero when subnormal.
+  final ToggleConfigKnob denormalAsZeroB = ToggleConfigKnob(value: false);
+
+  /// If addition output is subnormal, flush to zero.
+  final ToggleConfigKnob flushTozero = ToggleConfigKnob(value: false);
```
mkorbel1: typo: capitalize the Z in `flushToZero`

test/arithmetic/floating_point/floating_point_conversion_test.dart
```diff
@@ -14,9 +14,13 @@ import 'package:test/test.dart';
```
mkorbel1, line 14: do converters, multipliers, etc. that don't support DAZ/FTZ need to throw an exception?

mkorbel1 (changes requested):

lib/src/arithmetic/floating_point/floating_point_adder.dart
desmonddak: I didn't realize I could simply make it final.  This is now fixed.

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
desmonddak: I think this may have been a remnant of when we were inheriting?

test/arithmetic/floating_point/floating_point_conversion_test.dart
desmonddak: Yes.  These are to be implemented but should throw for now.

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
desmonddak: I propose a method (can't come up with the name) on FPL that 'resolves' the logic to the final form, but do that INSIDE the component.  So the component says 'get me the real a', and that method does the muxing right there where the component asks for the logic.  After I went through the changes I realized it would have been better to encapsulate this muxing logic in FloatingPoint.

We have populator for FPV to that can change the type and grab just the values (ofFloatingPointValue), but we don't have that for FPL.  Do we want to overload clone or use a different method to do this type conversion?

.github/workflows/general.yml
desmonddak: Yes.  I should fix.

I tested 3.29 and it fails the analysis stage because dart:html is now fully deprecated and the changes look really painful (I tried to convert but got lost).  3.27 fixes the weird shadow bugs in the display, but without requiring dart:html to be removed (the new methodology is package:web).  It gets really weird because they changed WebASM and the Worker class, so there are tons of problems, not just renamings (e.g. a lot of things just moved to web.*  but the WebASM stuff is quite different).

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
desmonddak: Added FloatingPointLogic.resolveSubNormalAsZero() as well as factory zero().

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
```diff
@@ -81,12 +81,28 @@ class FloatingPoint extends LogicStructure {
...
-  late final bool _explicitJBit;
+  final bool explicitJBit;
 
   /// Return true if subnormal numbers are represented as zero.
-  bool get subNormalAsZero => _subNormalAsZero;
-  late final bool _subNormalAsZero;
+  final bool subNormalAsZero;
+
+  /// Convert the current [FloatingPoint] to a new [FloatingPoint] but with the
+  /// mantissa resolved if subnormal and subNormalAsZero is true.
```
mkorbel1, line 90: nit: put `[subNormalAsZero]` with brackets, and could put "`"s around true as well

lib/src/arithmetic/floating_point/floating_point_adder_singlepath.dart
```diff
@@ -353,6 +356,11 @@ class FloatingPointAdderSinglePath<FpTypeIn extends FloatingPoint,
       }
       exponentRound = exponent;
     }
+    // Handle Flush to Zero subnormal case
+    mantissaRound = (internalSum.subNormalAsZero
```
mkorbel1, line 360: what if `a` and `b` have different `subNormalAsZero`?  In this case, the output matches `a` if it's not provided.  That seems fine, but we should probably document that right?

mkorbel1 (approved):
Looks good to me, can we address those two comments before merging?

PR #214, opened 2025-06-25
Runtime configuration using dynamic

### LINKED ISSUE #185: Optional configuration or logical control signal component design pattern
### Motivation

We need a consistent way to configure components with an option that can either be a configuration or, alternatively a logical control of that feature.  For example, a signed multiplier could be either configured permanently as signed, or configured with a control signal that dynamically selects between signed and unsigned.

The challenge is that we need to provide a bool configuration option and a Logic signal that are related. 
One strategy has been to allow the Logic? signal to be active if the boolean option is false and to throw if the bool option is set true when the Logic? signal is provided (not null).  Alternatively both options could be provided as nullable.  Either can be confusing because a bool option that is nullable -- what is the configuration if the boolean option is null versus false versus true?  What is the configuration if the boolean option is true and the Logic signal is provided?  Why is it illegal for this configuration versus setting it false and providing a Logic signal.

### Desired solution

One idea is to provide a configuration class that represents this duality. 
     --> Think of a solution analogous to unions in C providing one name for two variables in the same space.

 Another is to standardize on both being provided as nullable with a naming convention to align them.  Another is to standardize on the false/non-null dichotomy.  

Then the different components that provide both fixed and dynamic configuration of a particular feature align on the same mechanism as part of the ROHD-HCL API standardization.

### Alternatives considered

_No response_

### Additional details

_No response_

desmonddak:
## Description & Motivation

This is a change to the integer multiplier API to configure signed vs unsigned operands using a dynamic instead of pairs of bool/Logic? parameters.

## Related Issue(s)

fix #185 

## Testing

Used existing regression tests.

## Backwards-compatibility

Yes.  The parameters passed to configure the integer multipliers are now single dynamic inputs rather than a boolean or a Logic?

## Documentation

Only API documentation has changed.  The markdown still needs updating.

### DIFF
--- a/lib/src/arithmetic/floating_point/floating_point_adder.dart
+++ b/lib/src/arithmetic/floating_point/floating_point_adder.dart
@@ -52,6 +52,7 @@ abstract class FloatingPointAdder<FpTypeIn extends FloatingPoint,
   late final FpTypeOut internalSum;
 
   /// The rounding mode to use for the adder.
+  @protected
   late final FloatingPointRoundingMode roundingMode;
 
   /// Add two floating point numbers [a] and [b], returning result in [sum].
--- a/lib/src/arithmetic/multiplier.dart
+++ b/lib/src/arithmetic/multiplier.dart
@@ -40,6 +40,18 @@ class Config {
     }
   }
 
+  /// Factory constructor to create a [Config] instance from a dynamic.
+  factory Config.ofDynamic(dynamic config) {
+    if (config is Logic) {
+      return RuntimeConfig(config, name: config.name);
+    } else if (config is bool) {
+      return BooleanConfig(staticConfig: config);
+    } else {
+      throw RohdHclException(
+          'Unsupported configuration type: ${config.runtimeType}');
+    }
+  }
+
   /// Return a bool representing the value of the configuration.
   @visibleForTesting
   bool get value =>
--- a/lib/src/arithmetic/multiplier.dart
+++ b/lib/src/arithmetic/multiplier.dart
@@ -132,46 +132,31 @@ abstract class Multiplier extends Module {
   @protected
   Logic get isProductSigned => output('isProductSigned');
 
-  /// Take input [a] and input [b] and return the
-  /// [product] of the multiplication result.
+  /// Take input [a] and input [b] and return the [product] of the
+  /// multiplication result.
   ///
-  /// [signedMultiplicand] parameter configures the multiplicand [a] as a signed
-  /// multiplier (default is unsigned).
+  /// [signedMultiplicandConfig] parameter configures the multiplicand [a] as a
+  /// signed multiplicand (default is unsigned) or a runtime configurable
+  /// [selectSignedMultiplicand] input.
   ///
-  /// [signedMultiplier] parameter configures the multiplier [b] as a signed
-  /// multiplier (default is unsigned).
+  /// [signedMultiplierConfig] parameter configures the multiplier [b] as a
+  /// signed multiplier (default is unsigned) or a runtime configurable
+  /// [selectSignedMultiplier] input.
   ///
-  /// Optional [selectSignedMultiplicand] allows for runtime configuration of
-  /// signed or unsigned operation, overriding the [signedMultiplicand] static
-  /// configuration.
-  ///
-  /// Optional [selectSignedMultiplier] allows for runtime configuration of
-  /// signed or unsigned operation, overriding the [signedMultiplier] static
-  /// configuration.
-  /// If [clk] is not null then a set of flops are used to make the multiply
-  /// a 2-cycle latency operation. [reset] and [enable] are optional
-  /// inputs to control these flops when [clk] is provided.
+  /// If [clk] is not null then a set of flops are used to make the multiply a
+  /// 2-cycle latency operation. [reset] and [enable] are optional inputs to
+  /// control these flops when [clk] is provided.
   Multiplier(Logic a, Logic b,
       {Logic? clk,
       Logic? reset,
       Logic? enable,
       this.signedMultiplicandConfig,
       this.signedMultiplierConfig,
-      // this.signedMultiplicand = false,
-      // this.signedMultiplier = false,
-      // Logic? selectSignedMultiplier,
       super.name = 'multiplier',
       String? definitionName})
       : super(
             definitionName:
                 definitionName ?? 'Multiplier_W${a.width}x${b.width}') {
-    // if (signedMultiplicand && (selectSignedMultiplicand != null)) {
-    //   throw RohdHclException('multiplicand sign reconfiguration requires '
-    //       'signedMultiplicand=false');
-    // }
-    // if (signedMultiplier && (selectSignedMultiplier != null)) {
-    //   throw RohdHclException('sign reconfiguration requires signed=false');
-    // }
     this.clk = (clk != null) ? addInput('clk', clk) : null;
     this.reset = (reset != null) ? addInput('reset', reset) : null;
     this.enable = (enable != null) ? addInput('enable', enable) : null;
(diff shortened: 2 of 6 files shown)


lib/src/arithmetic/floating_point/floating_point_adder.dart
```diff
@@ -52,6 +52,7 @@ abstract class FloatingPointAdder<FpTypeIn extends FloatingPoint,
   late final FpTypeOut internalSum;
 
   /// The rounding mode to use for the adder.
+  @protected
```
mkorbel1: get rid of `@protected` and remove the `late`

lib/src/arithmetic/multiplier.dart
```diff
@@ -13,6 +13,98 @@ import 'package:meta/meta.dart';
 import 'package:rohd/rohd.dart';
 import 'package:rohd_hcl/rohd_hcl.dart';
 
+/// A general configuration class for specifying parameters that are
+/// for both static or runtime configurations of a component feature.
+class Config {
```
mkorbel1: Move this configuration stuff into a new file?

lib/src/arithmetic/multiplier.dart
```diff
@@ -38,89 +130,87 @@ abstract class Multiplier extends Module {
...
+  /// control these flops when [clk] is provided.
   Multiplier(Logic a, Logic b,
       {Logic? clk,
       Logic? reset,
       Logic? enable,
-      this.signedMultiplicand = false,
-      this.signedMultiplier = false,
-      Logic? selectSignedMultiplicand,
-      Logic? selectSignedMultiplier,
+      dynamic signedMultiplicandParam,
```
mkorbel1: why suffix with "Param"?  maybe we just omit that and call it `signedMultiplicand` etc?

lib/src/arithmetic/multiplier.dart
```diff
@@ -38,89 +130,87 @@ abstract class Multiplier extends Module {
...
-  /// Take input [a] and input [b] and return the
-  /// [product] of the multiplication result.
+  /// Take input [a] and input [b] and return the [product] of the
+  /// multiplication result.
   ///
-  /// [signedMultiplicand] parameter configures the multiplicand [a] as a signed
-  /// multiplier (default is unsigned).
+  /// The optional [signedMultiplicandParam] parameter configures the
+  /// multiplicand [a] as a signed multiplicand (default is unsigned) or with a
+  /// runtime configurable [selectSignedMultiplicand] input.
```
mkorbel1: I think we should be more explicit in this documentation about the allowed types: either a `bool` or a 1-bit `Logic`, right?  Also, I think the description "runtime" is a little weird in the context of hardware -- is there a better way to describe this?  Like a `bool` is a static configuration wheras a `Logic` is a dynamic configuration controllable by other logic?

lib/src/arithmetic/multiplier.dart
```diff
@@ -38,89 +130,87 @@ abstract class Multiplier extends Module {
...
-                definitionName ?? 'Multiplier_W${a.width}x${b.width}') {
-    if (signedMultiplicand && (selectSignedMultiplicand != null)) {
-      throw RohdHclException('multiplicand sign reconfiguration requires '
-          'signedMultiplicand=false');
-    }
-    if (signedMultiplier && (selectSignedMultiplier != null)) {
-      throw RohdHclException('sign reconfiguration requires signed=false');
-    }
+            definitionName: definitionName ??
+                '${b.width}_$_signedMD(signedMultiplicandConfig)}'
```
mkorbel1: doesn't the naming here have `null` if it's null?  maybe something like 
```dart
[
  'asdf',
  if(x != null) '$x'
].nonNulls.join('_')
```

so that we can avoid null in the definition names and just omit things if they are not there?

but also, how is this not a "late variable not initialized" error at this point before you've set the config object?

lib/src/arithmetic/multiplier.dart
mkorbel1: Can we add a `@override String toString() => ...` to this class so that when you use it in naming of definitions, etc. it has a nice name?

mkorbel1: Perhaps a more descriptive name?  `Config` is very generic.

lib/src/arithmetic/multiplier.dart
```diff
@@ -13,6 +13,98 @@ import 'package:meta/meta.dart';
...
+  /// Return a bool representing the value of the configuration.
+  @visibleForTesting
+  bool get value =>
+      staticConfig ||
+      (runtimeConfig != null && runtimeConfig!.value == LogicValue.one);
+
+  /// Return the internal [Logic] signal that represents the configuration,
+  /// either static or runtime.
+  Logic getLogic(Module module) =>
+      staticConfig ? Const(1) : (tryRuntimeInput(module) ?? Const(0));
```
mkorbel1: why not `getRuntimeInput` instead of try here?  That way we don't need to call `getRuntimeInput` ahead of time before calling `getLogic`?

lib/src/arithmetic/multiplier.dart
```diff
@@ -38,89 +130,87 @@ abstract class Multiplier extends Module {
...
-        (signedMultiplicand | signedMultiplier ? Const(1) : Const(0)) |
-            ((selectSignedMultiplicand != null)
-                ? selectSignedMultiplicand
+        ((signedMultiplicandConfig != null
+                ? signedMultiplicandConfig!.getLogic(this)
                 : Const(0)) |
-            ((selectSignedMultiplier != null)
-                ? selectSignedMultiplier
-                : Const(0));
+            (signedMultiplierConfig != null
```
mkorbel1: won't this never be null?  change the type of it to not be nullable?

mkorbel1 (changes requested):

lib/src/arithmetic/multiplier.dart
```diff
@@ -38,90 +38,114 @@ abstract class Multiplier extends Module {
...
-        : null;
-    selectSignedMultiplier = (selectSignedMultiplier != null)
-        ? addInput('selectSignedMultiplier', selectSignedMultiplier)
-        : null;
+    signedMultiplicandParameter =
+        StaticOrRuntimeParameter.ofDynamic(signedMultiplicand);
+    signedMultiplierParameter =
+        StaticOrRuntimeParameter.ofDynamic(signedMultiplier);
+
+    signedMultiplicandParameter.getRuntimeInput(this);
```
mkorbel1: Don't need this anymore, since it will get auto-called when you do getLogic

lib/src/arithmetic/multiply_accumulate.dart
```diff
@@ -112,29 +123,44 @@ abstract class MultiplyAccumulate extends Module {
...
-        ? addInput('selectSignedAddend', selectSignedAddend)
-        : null;
+
+    signedMultiplicandParameter =
+        StaticOrRuntimeParameter.ofDynamic(signedMultiplicand);
+    signedMultiplierParameter =
+        StaticOrRuntimeParameter.ofDynamic(signedMultiplier);
+    signedAddendParameter = StaticOrRuntimeParameter.ofDynamic(signedAddend);
+
+    signedMultiplicandParameter.getRuntimeInput(this);
```
mkorbel1: delete these `getRuntimeInput`s as well

lib/src/static_or_runtime_parameter.dart
```diff
@@ -0,0 +1,98 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// static_or_runtime_parameter.dart
+// Configuration classes for managing parameters that can be set statically or
+// at runtime.
+//
+// 2025 June 27 Author: Desmond Kirkpatrick <desmond.a.kirkpatrick@intel.com>
```
mkorbel1: nit: newline between date and your name

PR #215, opened 2025-06-28
Prepare for 0.2.0 release

desmonddak:
## Description & Motivation

Updated CHANGELOG from PRs since last release of 0.1.0.

## Related Issue(s)

None.

## Testing

Previewed markdown in VSCode.

## Backwards-compatibility

No

## Documentation

Yes.  This is a CHANGELOG update.


mkorbel1:
Yes, good idea!  Also, we should improve (lengthen) the description I guess to get the score up

PR #217, opened 2025-06-30
AXI4 Subordinate BFM bug fix for tight consecutive write/read

kimmeljo:
## Description & Motivation

If a main sends a write on a particular channel/address and then in the cycle after WLAST for the write data sends a read on the same channel/address, the subordinate would return the old data instead of the new data.

## Related Issue(s)

N/A

## Testing

Added a unit test for this case.

## Backwards-compatibility

This change is fully backwards compatible.

## Documentation

No doc updates needed.

### DIFF
--- a/lib/src/models/axi4_bfm/axi4_main_write_driver.dart
+++ b/lib/src/models/axi4_bfm/axi4_main_write_driver.dart
@@ -128,11 +128,15 @@ class Axi4WriteMainDriver extends PendingClockedDriver<Axi4WriteRequestPacket> {
         await wIntf.wReady.nextPosedge;
       }
       Simulator.injectAction(() {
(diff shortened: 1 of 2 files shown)


mkorbel1 (changes requested):
Can you add a unit test that exercises the bug to prove the bug fix?

kimmeljo:
> Can you add a unit test that exercises the bug to prove the bug fix?

There actually already was one (SimpleWriteRead) but it wasn't exercising it before because I marked the write as complete one cycle after WLAST. I made a change to mark the write complete simultaneous to WLAST so now that test does exercise the bug.

mkorbel1 (approved):
Ok, then looks good to me!

PR #218, opened 2025-06-30
change radix to a named arg for multipliers

desmonddak:
## Description & Motivation

radix was a positional arg for CompressionTreeMultiplier and MAC.  It would be better for it to be a named arg with a default.

## Related Issue(s)

None.

## Testing

Ran suite of multiplier tests.

## Backwards-compatibility

No as multpliers are new components in ROHD-HCL.

## Documentation

Document changes coming....


doc/components/multiplier_components.md
```diff
@@ -93,7 +93,7 @@ Note that radix-4 shifts by 2 positions each row, but with only two rows and wit
 The base class of `PartialProductGenerator` is [PartialProductArray](https://intel.github.io/rohd-hcl/rohd_hcl/PartialProductArray-class.html) which is simply a `List<List<Logic>>` to represent addends and a `rowShift[row]` to represent the shifts in the partial product matrix. If customization is needed beyond sign extension options, routines are provided that allow for fixed customization of bit positions or conditional (mux based on a Logic) form in the `PartialProductArray`.
 
 ```dart
-final ppa = PartialProductArray(a,b);
+final ppa = ppg as PartialProductArray;
```
mkorbel1, line 97: where'd `ppg` come from?

desmonddak: I can construct it before.  I was trying to show that this API is part of the basic Array whereas the PPG has a lot more.
So if we want to build other kinds of partial product summation routines (e.g. dot-product), we can just think about the simple PP Array which is really just List<List<Logic>> and List<int>  (the shift of each row).

PR #222, opened 2025-07-03
pushed dynamic configuration through all multiplier componentry

desmonddak:
## Description & Motivation

Completed the use of dynamic configuration of signed operands through the multiplier componentry.

## Related Issue(s)

None.

## Testing

Ran existing multiplier tests.

## Backwards-compatibility

Yes.  Various components now have simpler APIs for sign configuration of multiplicand and multiplier (and addend for MAC).

## Documentation

API documentation fully updated.  Markdown documentation TBD.

### DIFF
--- a/lib/src/arithmetic/multiplier_components/partial_product_generator.dart
+++ b/lib/src/arithmetic/multiplier_components/partial_product_generator.dart
@@ -19,10 +19,13 @@ class SignBit extends Logic {
   bool inverted = false;
 
   /// Construct a sign bit to store
-  SignBit(Logic inl, {this.inverted = false})
-      : super(name: '${inl.name}_signbit', naming: Naming.mergeable) {
+  SignBit(Logic inl, {this.inverted = false, String? name})
+      : super(name: name ?? inl.name, naming: Naming.mergeable) {
     this <= inl;
   }
+  @override
+  SignBit clone({String? name}) =>
+      SignBit(this, inverted: inverted, name: name);
 }
 
 /// A [PartialProductArray] is a class that holds a set of partial products
@@ -294,8 +297,13 @@ abstract class PartialProductMatrix extends Module {
   /// Store the [PartialProductGeneratorBase]
   late final PartialProductGeneratorBase _array;
 
-  /// Base constructor for the matrix.
-  PartialProductMatrix({super.name = 'partial_product_matrix'});
+  /// Base constructor for the matrix
+  PartialProductMatrix(
+      {super.name = 'partial_product_matrix',
+      super.reserveName,
+      super.reserveDefinitionName,
+      String? definitionName})
+      : super(definitionName: definitionName ?? 'PartialProductMatrix');
 
   /// Generate the output vectors from the array
   void generateOutputs() {
@@ -335,7 +343,15 @@ class PartialProduct extends PartialProductMatrix {
               Logic multiplicand, Logic multiplier, RadixEncoder radixEncoder,
               {dynamic signedMultiplicand, dynamic signedMultiplier})
           genPPG = PartialProductGenerator.new,
-      super.name = 'partial_product'}) {
+      super.name = 'partial_product',
+      super.reserveName,
+      super.reserveDefinitionName,
+      String? definitionName})
+      : super(
+            definitionName: definitionName ??
+                'PartialProduct_W${multiplicand.width}x${multiplier.width}'
+                    '_${Multiplier.signedMD(signedMultiplicand)}_'
+                    '${Multiplier.signedML(signedMultiplier)}') {
     final signedMultiplicandParameter =
         StaticOrDynamicParameter.ofDynamic(signedMultiplicand);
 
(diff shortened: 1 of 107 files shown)


desmonddak:
It would be nice to generalize this to integer or boolean configurations.  A simple test example I can use is a shift register (or the rotater) which would take either a Logic or an int for a shift input.

lib/src/arithmetic/multiplier_components/partial_product_generator.dart
```diff
@@ -41,6 +41,7 @@ abstract class PartialProductArray {
   /// The actual shift in each row. This value will be modified by the
   /// sign extension routine used when folding in a sign bit from another
   /// row.
+  /// row.
```
mkorbel1, line 44: typo: accidentally added duplicate "row." here?

lib/src/arithmetic/multiplier_components/partial_product_sign_extend.dart
```diff
@@ -265,20 +248,24 @@ class CompactSignExtension extends PartialProductSignExtension {
     // Mohanty, B.K., Choubey, A. Efficient Design for Radix-8 Booth Multiplier
     // and Its Application in Lifting 2-D DWT. Circuits Syst Signal Process 36,
     // 1129–1149 (2017). https://doi.org/10.1007/s00034-016-0349-9
-    if (signedMultiplicand && (selectSignedMultiplicand != null)) {
-      throw RohdHclException('multiplicand sign reconfiguration requires '
-          'signedMultiplicand=false');
-    }
+    // if (signedMultiplicand && (selectSignedMultiplicand != null)) {
```
mkorbel1, line 251: rm old code

lib/src/arithmetic/multiplier_components/partial_product_sign_extend.dart
```diff
@@ -427,14 +415,16 @@ class StopBitsSignExtension extends PartialProductSignExtension {
   ///
   @override
   void signExtend() {
-    if (signedMultiplicand && (selectSignedMultiplicand != null)) {
-      throw RohdHclException('multiplicand sign reconfiguration requires '
-          'signedMultiplicand=false');
-    }
+    // if (signedMultiplicand && (selectSignedMultiplicand != null)) {
```
mkorbel1, line 418: rm old code

lib/src/arithmetic/multiplier_components/partial_product_sign_extend.dart
```diff
@@ -455,13 +445,15 @@ class StopBitsSignExtension extends PartialProductSignExtension {
     for (var row = 0; row < rows; row++) {
       final addend = partialProducts[row];
       final Logic sign;
-      if (selectSignedMultiplicand != null) {
-        sign = mux(selectSignedMultiplicand!, addend.last, signs[row]);
+      if (signedMultiplicandParameter.dynamicConfig != null) {
```
mkorbel1, line 448: I notice a lot of checking for `!= null`, would be nice to have a `bool get isDynamic` helper?

lib/src/static_or_dynamic_parameter.dart
```diff
@@ -0,0 +1,107 @@
...
+    if (dynamicConfig == null && staticConfig != null) {
+      this.staticConfig = staticConfig;
+    } else {
+      this.staticConfig = false;
+    }
+  }
+
+  /// Factory constructor to create a [StaticOrDynamicParameter] instance from a
+  /// dynamic.
+  factory StaticOrDynamicParameter.ofDynamic(dynamic config) {
```
mkorbel1, line 43: It seems like this the most commonly used constructor, yet the name is the longest and also a little weird (though I dont have a better suggestion, it seems appropriate) since "dynamic" is used in two meanings in the same call: a changing Logic and a dynamically typed variable.

What if we made this constructor just `StaticOrDynamicParameter` and the other one something like `StaticOrDynamicParameter.withConfigs` or something (maybe a better name)?

lib/src/arithmetic/multiplier_components/partial_product_sign_extend.dart
```diff
@@ -207,10 +190,12 @@ class BruteSignExtension extends PartialProductSignExtension {
...
+        sign = mux(signedMultiplicandParameter.dynamicConfig!, addend.last,
```
mkorbel1, line 194: what if we added a `mux` or `select` function into this `StaticOrDynamicParameter` that automatically handles the switch between using a `mux` or a ternary operator?  this would only work when it's a 1-bit value.

also, `mux` will be updated (eventually) so that if the select is a `Const`, it will resolve the result statically.  but it still might be worth adding it as an option on this infrastructure

lib/src/arithmetic/ones_complement_adder.dart
```diff
@@ -68,19 +71,10 @@ class OnesComplementAdder extends Adder {
...
     }
-    if ((subtractIn != null) & subtract) {
-      throw RohdHclException(
-          "either provide a Logic signal 'subtractIn' for runtime "
-          " configuration, or a boolean parameter 'subtract' for "
-          'generation time configuration, but not both.');
-    }
-    this.subtractIn =
-        (subtractIn != null) ? addInput('subtractIn', subtractIn) : null;
+    subtractIn = StaticOrDynamicParameter.ofDynamic(subtract).clone(this);
```
mkorbel1, line 74: why clone?

PR #223, opened 2025-07-06
fix  proper quotations of classes and literals true and false

desmonddak:
## Description & Motivation

API documentation does not properly quote classes and literals.  In a few cases, the result is what looks like HTML to pub.dev grading systems.  This PR cleans up API documentation to use quotations around such cases as well as things like `true` and
`false`.

## Related Issue(s)

None.

## Testing

Previewed a few API docs to make sure the quotations look correct.

## Backwards-compatibility

No.

## Documentation

These are documentation tweaks.


lib/src/arithmetic/arithmetic_utils.dart
```diff
@@ -16,17 +16,17 @@ import 'package:rohd_hcl/rohd_hcl.dart';
 
 /// Helper evaluation methods for printing aligned arithmetic bitvectors.
 extension LogicList on List<Logic> {
-  /// Print aligned bitvector with an optional header from List<Logic>.
+  /// Print aligned bitvector with an optional header from `List<Logic>`.
```
mkorbel1: better is `[List<Logic>]`

lib/src/arithmetic/floating_point/floating_point_converter.dart
```diff
@@ -13,7 +13,7 @@ import 'package:meta/meta.dart';
 import 'package:rohd/rohd.dart';
 import 'package:rohd_hcl/rohd_hcl.dart';
 
-/// A converter module for FloatingPoint values
+/// A converter module bwtween different [FloatingPoint] logic signals.
```
mkorbel1: typo: between

lib/src/arithmetic/parallel_prefix_operations.dart
```diff
@@ -118,9 +118,9 @@ class KoggeStone extends ParallelPrefix {
   }
 }
 
-/// BrentKung shaped ParallelPrefix tree
+/// Brent-Kung shaped ParallelPrefix tree.
```
mkorbel1: missed brackets

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
```diff
@@ -269,8 +267,8 @@ class FloatingPointValue implements Comparable<FloatingPointValue> {
     return doubleVal;
   }
 
-  /// Return a Logic true if this FloatingPointVa;ie contains a normal number,
-  /// defined as having mantissa in the range `[1,2)`.
+  /// Return a `Logic` `1` if this [FloatingPointValue] contains a normal
```
mkorbel1: but this returns a `bool`?

lib/src/fifo.dart
```diff
@@ -324,7 +324,7 @@ class FifoChecker extends Component {
 
 /// A tracker for a [Fifo] which can generate logs.
 class FifoTracker extends Tracker {
-  /// Internal tracking of occupancy in case the Fifo didn't generate it.
+  /// Internal tracking of occupancy in case the FIFO didn't generate it.
```
mkorbel1: could be `[Fifo]`?

lib/src/priority_encoder.dart
```diff
@@ -46,7 +46,7 @@ abstract class PriorityEncoder extends Module {
   }
 }
 
-/// Priority finder based on or() operations
+/// Priority finder based on or() operations.
```
mkorbel1, line 49: either "OR" or "`or()`" ?

lib/src/sort.dart
```diff
@@ -31,11 +31,11 @@ abstract class Sort extends Module {
       super.definitionName});
 }
 
-/// Compare and Swap [Logic] to the specified order.
+/// Compare and Swap `Logic` to the specified order.
```
mkorbel1: Why switch these from brackets to ticks in this file?

lib/src/arithmetic/parallel_prefix_operations.dart
desmonddak: I wanted to say [BrentKung] is an implementation of the Brent-Kung parallel prefix pattern.

lib/src/fifo.dart
desmonddak: This is an issue throughout this file.  I'll explore how to update.

lib/src/priority_encoder.dart
desmonddak: OR is better, I'll fix.

lib/src/sort.dart
desmonddak: Mistake.  Odd one.

PR #227, opened 2025-07-08
add definitionName to all components and update configurator/markdown…

desmonddak:
… to use plain schematic name

## Description & Motivation

It would be nice to be able to set the definition name of every component.  This PR allows for this by passing in an overriding
definitionName option for each component.

## Related Issue(s)

None

## Testing

Ran all tests.   HTML deployment will need live site testing.

## Backwards-compatibility

No.

## Documentation

Markdowns link to new plain HTML names.  Otherwise no documentation change is needed as these are additional, optional 
constructor arguments.

### DIFF
--- a/lib/src/memory/csr/csr_top.dart
+++ b/lib/src/memory/csr/csr_top.dart
@@ -84,15 +84,15 @@ class CsrTop extends CsrContainer {
   }
 
   /// Create the CsrBlock from a configuration
-  CsrTop({
-    required CsrTopConfig super.config,
-    required super.clk,
-    required super.reset,
-    required super.frontWrite,
-    required super.frontRead,
-    super.allowLargerRegisters,
-    this.logicalRegisterIncrement = 1,
-  }) {
+  CsrTop(
+      {required CsrTopConfig super.config,
+      required super.clk,
+      required super.reset,
+      required super.frontWrite,
+      required super.frontRead,
+      super.allowLargerRegisters,
+      this.logicalRegisterIncrement = 1,
+      super.definitionName}) {
     _validate();
 
     for (final block in config.blocks) {
--- a/lib/src/memory/register_file.dart
+++ b/lib/src/memory/register_file.dart
@@ -23,10 +23,11 @@ class RegisterFile extends Memory {
   /// [MaskedDataPortInterface]s are supported on `writePorts`, but not on
   /// `readPorts`.
   RegisterFile(super.clk, super.reset, super.writePorts, super.readPorts,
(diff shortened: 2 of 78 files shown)


lib/src/memory/csr/csr_top.dart
```diff
@@ -83,16 +83,23 @@ class CsrTop extends CsrContainer {
...
+      required super.frontRead,
+      super.allowLargerRegisters,
+      this.logicalRegisterIncrement = 1,
+      String? definitionName})
+      : super(
+            definitionName: definitionName ??
+                'CsrTop_A${config.minAddrBits()}_'
+                    'W${config.maxRegWidth()}_'
+                    'BO${config.blockOffsetWidth}_'
+                    'LR=${allowLargerRegisters}_'
```
mkorbel1: I think this `=` will get sanitized down to a `_`

lib/src/memory/register_file.dart
```diff
@@ -23,10 +23,11 @@ class RegisterFile extends Memory {
   /// [MaskedDataPortInterface]s are supported on `writePorts`, but not on
   /// `readPorts`.
   RegisterFile(super.clk, super.reset, super.writePorts, super.readPorts,
-      {this.numEntries = 8, super.name = 'rf'})
+      {this.numEntries = 8, super.name = 'rf', String? definitionName})
```
mkorbel1: While we're at it, do we want to expose `reserveName` and `reserveDefinitionName` as well?  Not sure we do, but maybe?  Thoughts? (not specific to this one, just asking generally for all components)

mkorbel1 (approved):

lib/src/memory/csr/csr_top.dart
desmonddak: oops

lib/src/memory/register_file.dart
desmonddak: Is this just a super.reserveDefinitionName as a parameter?

To force synthesis to use the provided name and not uniquify?  It makes sense to me to thread that through in all components.

mkorbel1: yes, right.  do you want to put it in this PR or we can do it later?

desmonddak: I fixed some additional missed definitionName default settings and I threaded the reserveName and reserveDefinitionName alongside in all component constructors (I don't think I missed any).

lib/src/memory/csr/csr_top.dart
mkorbel1: I still see the `=` in here?

Issue #198, opened 2025-05-19
Negate `FloatingPoint`s


mkorbel1:
### Motivation

Sometimes it's useful to be able to negate a FloatingPoint, but it's not that convenient right now

### Desired solution

Add a function `negate()` or maybe `negated` that returns -1 * the floating point number.  Should consider NaN, infinity, etc. as well.

PR #232, opened 2025-07-11
Floating Point Value operations: negation and comparison

desmonddak:
## Description & Motivation

This PR adds basic operations on floating point value types: negation and comparison.

Add feature for #198, #199

## Testing

Ran dart test along with new tests added for new operations.

## Backwards-compatibility

No.  FixedPointValue operators currently return `LogicValue` and they will now eventually `bool`, so these are deprecated and temporary methods `.ltBool`, `.lteBool`, `.gtBool` and `.gteBool` are provided as a working solution.  We are deprecating methods like `.eq` as well (e.g. use `==` instead).

## Documentation

None required as these are rounding out expected functionality from the datatype.


mkorbel1 (approved):
Looks great!

mkorbel1:
I unlinked #199 since we also want the `FixedPoint` stuff.

test/arithmetic/floating_point/floating_point_test.dart
```diff
@@ -110,4 +110,57 @@ void main() {
 
     expect(checkRan, isTrue);
   });
+
+  test('FP Comparison', () {
+    final fp1 = FloatingPoint32();
+    final fp2 = FloatingPoint32();
+
+    final val1 = FloatingPoint32Value.populator().ofDouble(1.23);
```
mkorbel1, line 120: should this have testing for subnormals, explicit jbit, negative numbers, special NaN stuff, etc?

mkorbel1 (changes requested):

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
```diff
@@ -234,4 +234,77 @@ class FloatingPoint extends LogicStructure {
...
+        Logic(width: exponent.width)..gets(exponent),
+        Logic(width: mantissa.width)..gets(mantissa),
+        explicitJBit,
+        subNormalAsZero,
+        name: name,
+      );
+
+  /// Negate the [FloatingPoint].
+  FloatingPoint operator -() => negate();
+  // ignore the lint warning about overriding the '-' operator.
```
mkorbel1: i dont see any lint warning here when i check it out.  what do you mean?

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
```diff
@@ -234,4 +234,77 @@ class FloatingPoint extends LogicStructure {
...
+
+  @override
+  Logic operator >(dynamic other) => gt(other);
+  @override
+  Logic operator >=(dynamic other) => gte(other);
+
+  /// Equal
+  @override
+  Logic eq(dynamic other) {
+    if (other is! FloatingPoint) {
```
mkorbel1, line 259: instead of copy-pasting this validation routine, can we extract it to a common validation function in this class?

you might even be able to extract that NaN comparision mux too

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
```diff
@@ -161,14 +161,37 @@ class FloatingPointValue implements Comparable<FloatingPointValue> {
...
+        (other.exponent.isZero && other.mantissa.isZero)) {
+      return 0;
+    }
+    final signCompare = -sign.compareTo(other.sign);
+
+    final canonical = canonicalize();
+    final otherCanonical = other.canonicalize();
+
+    final canonicalMantissa = canonical.explicitJBit
+        ? canonical.mantissa.reversed.getRange(1).reversed
```
mkorbel1: i think you could also do `canonical.mantissa.getRange(0, -1)` to remove the last bit, if that's what you're trying to do

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
```diff
@@ -178,34 +201,50 @@ class FloatingPointValue implements Comparable<FloatingPointValue> {
     return 0;
   }
 
+  /// Equality function for [FloatingPointValue].
+  bool eq(FloatingPointValue other) {
```
mkorbel1: what's the point of `eq` when it's the same as `==`?  In base `LogicValue` there's a difference because `eq` can return X but `==` is a bool return value.  In this case, is there a value in having both?

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
```diff
@@ -572,35 +572,362 @@ class FloatingPointValuePopulator<FpvType extends FloatingPointValue> {
     );
   }
 
-  /// Generate a random [FloatingPointValue], supplying random seed [rv].
+  void _checkMatching(String name, FpvType? fpv) {
```
mkorbel1, line 577: add doc comment?

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
```diff
@@ -572,35 +572,362 @@ class FloatingPointValuePopulator<FpvType extends FloatingPointValue> {
...
-              exponentWidth: exponentWidth, mantissaWidth: mantissaWidth - 1)
-          .random(rv, normal: normal)) as FpvType;
+  /// If [normal] is `true`, this method will only generate mantissas in the
+  /// range of `[1,2)` and `minExponent() <= exponent <= maxExponent()`,
+  /// otherwise if [subNormal] is `true`, it will only generate mantissas in the
+  /// range of `[0,1)` and `exponent == 0`. If both are `false`, it will
+  /// generate mantissas in the range of `[0,2)` and `minExponent() <= exponent
+  /// <= maxExponent()`.
+  ///
+  /// The [normal] and [subNormal] parameters are deprecated, please use
```
mkorbel1, line 607: Can we use an `enum` or something instead of having independent `bool`s with some illegal combinations?

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
```diff
@@ -572,35 +572,362 @@ class FloatingPointValuePopulator<FpvType extends FloatingPointValue> {
...
+  ///
+  /// If [genNormal] is `true`, this method will generate normal numbers, and if
+  /// [genSubNormal  ] is `true`, it will generate subnormal numbers. If both
+  /// are `false`, an exception will be thrown. Note that the range of numbers
+  /// to be generated is respected by these flags, so a range which does not
+  /// contain the requested type of number will result in a throw.
+  ///
+  /// You can constrain the range of random numbers generated by using a lower
+  /// bound specifed by [gt] or [gte] and an upper bound specified by [lt] or
+  /// [lte]. Only one of [gt] and [gte] can be specified, and only one of [lt]
```
mkorbel1: Why can't all four be specified simultaneously and internally it just computes the actual constraints as the more limiting ones?  Or maybe even better: accept a list of some sort of `FloatingPointConstraint<FpvType>` object or something which lets you specify gt, lt, gte, lte, (even eq?), and then internally it picks the most constrained ones?  Or I guess that could make it so multiple independent ranges exist.

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
```diff
@@ -572,35 +572,362 @@ class FloatingPointValuePopulator<FpvType extends FloatingPointValue> {
...
+  /// - [gt], [lt]: generate a value in the range `([gt], [lt])`
+  /// - [gte], [lt]: generate a value in the range `[[gte], [lt])`
+  /// - [gt], [lte]: generate a value in the range `([gt], [lte]]`
+  /// - [gte], [lte]: generate a value in the range `[[gte], [lte]]`
+  /// - [gt]: generate a value in the range `([gt], ∞)`
+  /// - [gte]: generate a value in the range `[[gte], ∞)`
+  /// - [lt]: generate a value in the range `(-∞, [lt])`
+  /// - [lte]: generate a value in the range `(-∞, [lte]]`
+  /// - none: generate a value in the range `(-∞, ∞)`
+  FpvType random(Random rv,
```
mkorbel1, line 634: A tricky question: what is the random distribution generated by this function?  Is it uniform over bit-vector values or uniform over magnitudes of the overall floating point number?  Which do we want (or both as an option?)?  Regardless, it should be clear in the doc comment at least.

lib/src/arithmetic/values/fixed_point_populator.dart
```diff
@@ -78,8 +78,8 @@ class FixedPointValuePopulator<FxvType extends FixedPointValue> {
         ? 1 + integerWidth + fractionWidth
         : integerWidth + fractionWidth;
     if (val.isFinite) {
-      final bigIntegerValue = BigInt.from(val * pow(2, fractionWidth));
-      final negBigIntegerValue = BigInt.from(-val * pow(2, fractionWidth));
+      final bigIntegerValue = BigInt.from(val * pow(2.0, fractionWidth));
```
mkorbel1, line 81: is this safe for large `fractionWidth`?

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
mkorbel1: Maybe two arguments: `upperBound` and `lowerBound`, each of which has a flag argument to allow for inclusive/exclusive?

lib/src/arithmetic/values/sign_magnitude_value.dart
```diff
@@ -0,0 +1,150 @@
...
+import 'dart:math';
+
+import 'package:meta/meta.dart';
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+export 'sign_magnitude_value_populator.dart';
+
+/// An immutable representation of a sign-magnitude value.
+@immutable
+class SignMagnitudeValue implements Comparable<SignMagnitudeValue> {
```
mkorbel1, line 19: is this distinct from a `FixedPointValue` with `nWidth == 0`?

lib/src/utils.dart
```diff
@@ -115,3 +120,138 @@ Logic condFlop(
...
+  /// [includeInvalidBits] to `true`. [max] can be used to set the maximum
+  /// range of the generated number and its only accept runtimeType `int` and
+  /// `BigInt`. [max] only work when [includeInvalidBits] is set to false
+  /// else an exception will be thrown. if [min] is provided, the generated
+  /// number will be in the range of [min] (inclusive) to [max] (exclusive).
+  /// [min] can be only runtimeType `int` and `BigInt`. [min] only work when
+  /// [includeInvalidBits] is set to false else an exception will be thrown
+  LogicValue nextLogicValueNew({
+    required int width,
+    dynamic min,
```
mkorbel1: for `Random.nextInt`, typically you can just do something like `Random().nextInt(max - min) + min`.  Why do we need to change the pattern for `LogicValue`?

test/arithmetic/floating_point/floating_point_test.dart
```diff
@@ -110,4 +112,104 @@ void main() {
...
+      FloatingPoint fpConstructor() => FloatingPoint(
+          exponentWidth: exponentWidth,
+          mantissaWidth: mantissaWidth,
+          explicitJBit: explicitJBit);
+      final fp1 = fpConstructor();
+      final fp2 = fpConstructor();
+
+      FloatingPointValuePopulator fpvPopulator() => fp1.valuePopulator();
+
+      for (var i = 0; i < 100; i++) {
```
mkorbel1: 100 only? :)

test/arithmetic/values/sign_magnitude_value_test.dart
```diff
@@ -0,0 +1,150 @@
...
+      final rv = Random(71);
+      for (var gtInt = -15; gtInt < 15; gtInt++) {
+        for (var ltInt = gtInt; ltInt < 15; ltInt++) {
+          final gt = populator().ofInt(gtInt);
+          final lt = populator().ofInt(ltInt);
+          for (var iter = 0; iter < 1000; iter++) {
+            // Single-sided greater-than
+            final fpvGtE = populator().random(rv, gte: gt);
+            final outGtE = fpvGtE.toInt();
+            if (outGtE < gtInt) {
```
mkorbel1: why `if`/`Exception` instead of `expect`?

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
desmonddak: Do we want `.gt=` along with `>=`?   And should these be boolean?  I thought that was what the Issue #199 was about.
Then how do we make these not surprising across the different comparison operators.

I was thinking we would provide operators for the Value types and only `.gte` style methods for `Logic`.

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
desmonddak: I guess that would be `[genNormal, genSubnormal, genAll]`.   I want to deprecate `normal` and `subNormal` because they were used already and mean something else (e.g. `normal` is `onlyNormal` really and defaults to false).

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
desmonddak: I played with these combinations but you still end up with 4.    I thought `gt|gte:,  lt|lte:` made the most sense (see the test code).  `ub:, ubInclusive:,  lb:, lbInclusive:` just seemed painful.    Right now the pain is inside the routine, outside it seems quite natural to just choose `gt:` or `gte:` as you would arithmetically.

You can get `eq` by using the same bound and using  gte: bound,  lte: bound.

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
desmonddak: Agreed.  It is uniform over the combined bitfield of exponent and mantissa.  It chooses the sign separately.
I accomplished this by creating a `SignMagnitude.random` and reused that for both fixed and floating point.

desmonddak: This approach enabled range constraints.

lib/src/arithmetic/values/fixed_point_populator.dart
desmonddak: By converting to 2.0 and making it double, it is safer than it was (remember the negative rollover when using `pow(2`.

I guess I could do this by staying in BigInt space to be safest.

lib/src/arithmetic/values/sign_magnitude_value.dart
desmonddak: Yes.  `FixedPointValue`, when `signed` is a twos-complement representation.

So `111` is -1 in fixed point, whereas `111` is negative 3 in sign-magnitude form.  This is useful for floating-point computation as well -- I intend to fix the places I use sign and magnitude to use this new datatype instead.

lib/src/utils.dart
desmonddak: When we move this to ROHD, we should make it much cleaner.  I was concerned about bitwidth management of `LogicValue`: e.g. do you need to zeroExtend after doing random, and then add and then slice back.

desmonddak: I can bypass this routine and do it in user space potentially as you suggest.  It looks like while a bit dangerous, it should work (e.g., you know that (14-8) + 8 will not need extra width to be correct.

desmonddak: Oh, but then we are pushing the int versus BigInt issues into user space as well.

desmonddak: `nextLogicValue(int max)` produces `LogicValue` of a given width, then we are adding an int or a BigInt to that as `min`, which needs to be converted to the same width `LogicValue`.  It seems that `nextLogicValue` taking a `min` argument does simplify.

desmonddak: I'll explore to see if it is simple.

test/arithmetic/values/sign_magnitude_value_test.dart
desmonddak: Sloppiness. :(

lib/src/utils.dart
desmonddak: You are right, far simpler, I just thought it was a little implicit:

```
rv.nextLogicValue(width: 4, max: limit) + LogicValue.ofBigInt(min, width:4);  // hides the fact that these can't overflow 4 bits
```

test/arithmetic/floating_point/floating_point_test.dart
desmonddak: It's much slower with constraints, so 1000 is like 15s.  I think 300 is reasonable for these narrow widths.

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
desmonddak: I looked at this again, and once we deprecate, there are only these three:  genNormal, genSubnormal, and excludeInfinity which seems clean (and easier to understand than using an enum).

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
desmonddak: I think I saw it in CI.  I'll omit and see if I still see it.

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
desmonddak: I know people are used to the python tricks on vectors.  I need a cheatsheet! :)

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
desmonddak: I can let all four be applied and choose the tighter bounds.

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
desmonddak: Removed methods and sticking with just operators for Value types.

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
desmonddak: Nope.  Must have been hallucination or a previous dart version.

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
```diff
@@ -428,6 +452,11 @@ class FloatingPointValue implements Comparable<FloatingPointValue> {
       exponent: exponent,
       mantissa: mantissa);
 
+  /// Negate the [FloatingPointValue].
+  FloatingPointValue operator -() => negate();
+  // ignore the lint warning about overriding the '-' operator.
```
mkorbel1: remove this one too?

lib/src/arithmetic/signals/floating_point_logics/floating_point_logic.dart
```diff
@@ -234,4 +234,69 @@ class FloatingPoint extends LogicStructure {
...
+  FloatingPoint operator -() => negate();
+
+  @override
+  Logic operator >(dynamic other) => gt(other);
+  @override
+  Logic operator >=(dynamic other) => gte(other);
+
+  /// Verify if comparable:  return `1` if comparable, throw exception
+  /// on mismatch.
+  Logic verifyComparable(dynamic other) {
```
mkorbel1: this should probably be private?

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
```diff
@@ -161,14 +161,39 @@ class FloatingPointValue implements Comparable<FloatingPointValue> {
...
+        (other.exponent.isZero && other.mantissa.isZero)) {
+      return 0;
+    }
+    final signCompare = -sign.compareTo(other.sign);
+
+    final canonical = canonicalize();
+    final otherCanonical = other.canonicalize();
+
+    final canonicalMantissa = canonical.explicitJBit
+        // ? canonical.mantissa.reversed.getRange(1).reversed
```
mkorbel1: remove the old commented out ones?

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
```diff
@@ -178,34 +203,33 @@ class FloatingPointValue implements Comparable<FloatingPointValue> {
     return 0;
   }
 
+  /// Inequality function for [FloatingPointValue].
+  // bool neq(FloatingPointValue other) => !eq(other);
```
mkorbel1: remove old ones not used

lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -114,29 +116,35 @@ class FixedPointValue implements Comparable<FixedPointValue> {
     }
   }
 
-  /// Equal-to operation that returns a [LogicValue].
-  LogicValue eq(FixedPointValue other) =>
-      compareTo(other) == 0 ? LogicValue.one : LogicValue.zero;
+  /// Equality function for [FixedPointValue].
+  bool eq(FixedPointValue other) => compareTo(other) == 0;
```
mkorbel1: remove the unnecessary duplicated functions in here too

lib/src/arithmetic/values/floating_point_values/floating_point_value_populator.dart
```diff
@@ -572,35 +572,389 @@ class FloatingPointValuePopulator<FpvType extends FloatingPointValue> {
...
+  /// otherwise if [subNormal] is `true`, it will only generate mantissas in the
+  /// range of `[0,1)` and `exponent == 0`. If both are `false`, it will
+  /// generate mantissas in the range of `[0,2)` and `minExponent() <= exponent
+  /// <= maxExponent()`.
+  ///
+  /// The [normal] and [subNormal] parameters are deprecated, please use
+  /// [genNormal] and [genSubNormal] instead.
+  ///
+  /// If [genNormal] is `true`, this method will generate normal numbers, and if
+  /// [genSubNormal  ] is `true`, it will generate subnormal numbers. If both
```
mkorbel1: nit: extra spaces in the brackets

lib/src/arithmetic/values/fixed_point_populator.dart
```diff
@@ -78,8 +78,9 @@ class FixedPointValuePopulator<FxvType extends FixedPointValue> {
         ? 1 + integerWidth + fractionWidth
         : integerWidth + fractionWidth;
     if (val.isFinite) {
-      final bigIntegerValue = BigInt.from(val * pow(2, fractionWidth));
-      final negBigIntegerValue = BigInt.from(-val * pow(2, fractionWidth));
+      final bigIntegerValue = BigInt.from(val) * BigInt.two.pow(fractionWidth);
```
mkorbel1: this seems safer for large widths, yes, but does it do the right math?  I don't know the answer, I'm just asking.  For example, what if val was `1.49`, then the `BigInt.from(val)` rounds to `1`.

test/arithmetic/floating_point/floating_point_test.dart
mkorbel1: why is it much slower?  it's relatively complicated, but time complexity is O(1) still right?

lib/src/arithmetic/values/sign_magnitude_value_populator.dart
```diff
@@ -0,0 +1,155 @@
...
+    if (lower > max) {
+      throw RohdHclException('randomSignMagnitude: invalid range, cannot '
+          'generate value');
+    }
+    if (limit <= zero) {
+      throw RohdHclException('randomSignMagnitude: invalid tight range, cannot '
+          'generate value');
+    }
+
+    // magnitude =
```
mkorbel1: rm old code

test/arithmetic/floating_point/floating_point_test.dart
desmonddak: Something strange is happening.  Here are the times for different loop lengths:

| iter |  time |
 |---- | ----- |
| 500 | 0:20 |
| 1000 | 1:55 |
| 2000 | 7:22|
| 4000 | 35:09|

lib/src/arithmetic/values/fixed_point_populator.dart
desmonddak: The most sensible is to let it be double before converting to BigInt because this is an ofDouble conversion.
So `BigInt.from(val * pow(2.0, fractionWidth))` is the accurate representation.

test/arithmetic/floating_point/floating_point_test.dart
desmonddak: Removed hardware generation from inner loop.

lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -114,29 +116,17 @@ class FixedPointValue implements Comparable<FixedPointValue> {
     }
   }
 
-  /// Equal-to operation that returns a [LogicValue].
-  LogicValue eq(FixedPointValue other) =>
-      compareTo(other) == 0 ? LogicValue.one : LogicValue.zero;
+  /// Less-than operator for [FixedPointValue].
```
desmonddak: What do we really want here?   I notice that LogicValue operators return a LogicValue.

But with fx and fp operators, should they return a LogicValue as well?

mkorbel1: `FixedPointValue` and `FloatingPointValue` are similar to `LogicValue`, but do not actually `implement` it.  For `LogicValue`, the use case may often involve modelling hardware, hence returning a `LogicValue` makes sense since an invalid input should result in an invalid output (perhaps partially, bitwise).  For these arithmetic types, it's less interesting for that, since it's probably either all valid or all invalid.  I think returning `bool`s makes sense for these.

mkorbel1: by the way, do these changes not consitute a backwards-incompatible change?  that's fine, just something to note.

mkorbel1 (commented):
Looking good once last couple comments are closed!

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
mkorbel1: i unresolved this one, looks like it's still there?

lib/src/arithmetic/values/fixed_point_value.dart
desmonddak: I will check.  My intent was no breaking changes on this PR.

desmonddak: There are breaking changes:  I changed the `FixedPointValue` comparisons to return bool instead of `LogicValue`.

desmonddak: Seems like the right approach would be to NOT merge this yet, but deprecate the old `FixedPointValue` methods like `eq`, `lt`, etc. and merge that. Then do a minor release of things we have. 

THEN, we can start merging some of the breaking changes in the pipeline, including this one. which would remove those  methods and change the operators to bool instead of `LogicValue`

mkorbel1: that makes sense to me!

mkorbel1 (approved):
Looks great, ready to merge!

You should update the description of the PR describing the breaking changes included.

PR #237, opened 2025-07-14
Fix signal names

desmonddak:
## Description & Motivation

This PR cleans up internal signal naming of most of the components in ROHD-HCL so that Verilog has very few oddly named signals (e.g. with '__').

This was accomplished by an external Dart program that we may incorporate that instantiated each Module (a few different configurations).  This may become part of our component checklist:  before you add a configurator, just adding these instantiations can help clarify the API and check the output SystemVerilog.

## Related Issue(s)

None

## Testing

Ran full tests.  Then ran instantiation test and grepped for '__' in output Verilog.

## Backwards-compatibility

No.

## Documentation

No updates needed.


lib/src/signed_shifter.dart
```diff
@@ -23,6 +23,16 @@ class SignedShifter extends Module {
     shift = addInput('shift', shift, width: shift.width);
 
     addOutput('shifted', width: bits.width);
-    shifted <= mux(shift[-1], bits >>> shift.abs(), bits << shift);
+    shifted <=
+        mux(shift[-1], bits >>> _abs(shift).named('shiftAbs'), bits << shift);
+  }
+
+  // TODO(desmonddak): replace with Logic.abs() when naming there is more clean.
```
mkorbel1: seems worth using the `abs` now and then just fix ROHD and get the update for free?

lib/src/arithmetic/floating_point/floating_point_converter.dart
```diff
@@ -200,34 +212,45 @@ class FloatingPointConverter<FpTypeIn extends FloatingPoint,
...
-          Const(0, width: maxExpWidth));
-
-      final tns = nextShift -
-          (se - newSe) +
-          mux(Const(source.explicitJBit), Const(-1, width: maxExpWidth),
-              Const(0, width: maxExpWidth)) +
-          mux(Const(destination.explicitJBit), Const(1, width: maxExpWidth),
-              Const(0, width: maxExpWidth));
+              biasDiff.gte(computedExpFull),
+              (source.isNormal.zeroExtend(maxExpWidth).named('srcIsNormal') +
```
mkorbel1: this is really a lot of `.named`s everywhere.  Is it worth it?

lib/src/utils.dart
```diff
@@ -114,3 +115,46 @@ Logic condFlop(
 
   return (first, second);
 }
+
+/// This class is used to check the SystemVerilog output of a module for
+/// poor naming conventions, such as names with `__` in them. It can also
+/// generate a SystemVerilog file for the module if requested.
+class CheckModuleNaming {
```
mkorbel1: Why is this a class?  Is this something we want to expose as part of the package or just an internal utility?  This file is exported.

lib/src/signed_shifter.dart
desmonddak: Yes.  There's a couple of fixes for ROHD we could explore.
I see a lot of ```~val + 1```, so a negate would be good and it currently produces  val_b_add__const ...
There's another area like `Conditional incr`  that do an add internally.

Generally adds are clumsy because they also produce the _carry and that naming is whatever is on the RHS + _carry.
So if we name the RHS well, then the _carry also comes out better.

lib/src/arithmetic/floating_point/floating_point_converter.dart
desmonddak: I was thinking of automating the check for clean SystemVerilog.  So this is zero-tolerance.  We may not want it, but these names quickly expand.  If you dont rename the zeroExtend, you get a swizzle. Then if  you add it to something, you can quickly get to swizzle_add__swizzle__add__subset....  It is great that the base system does this, don't get me wrong.  But if the metric is that this naming is in the components, then that means that people who write clean ROHD but use components get these names in their Verilog.  While we are trying to convince people about debuggability, I thought it would be good to get to zero strange names.

The current PR has *no* (as far as I have tested) names with '__'. (If you combine it with the definitionName PR).

desmonddak: I think there are better ways to write in ROHD that don't end up with so many nameds in one spot.  Like splitting them out of complex logic or assigning to Logics that are named already.

One I ran across was
```dart
count += (complex logic);
```
This resulted in a super long name __count__count__count__count ....  something like that.

It was FAR cleaner Verilog to write it as:
```dart
count = count + (complex Logic).named('increment');
```
And not using the += operator was actually beneficial to reading the code.

lib/src/utils.dart
desmonddak: It should be for testing only.  I'll label it.

Basically what this PR proposes is that for each module, we should write an instantiation test (and I discovered trace errors this way).  Just instantiate your module in its various configurations and check the names.  That would be a very quick test and would first make sure your instantiation works (no trace errors), and that the Verilog names can be checked and cleaned up.  you can see the module output and tune it to be more readable.

lib/src/arithmetic/floating_point/floating_point_converter.dart
desmonddak: Oh another thing I discovered.  There is probably a much better way to disposition Bool parameters by NOT converting them to Logic, but instead using var-style code.  This makes things far cleaner.

var  myLogic = <complex logic>;

if (parameter) {
      myLogic = <modified logic, including myLogic>
}

All this jbit code can come out more clean this way.   But for now, I was trying to get clean Verilog first.

lib/src/utils.dart
desmonddak: It is a class because it holds a global static variable to increment a number to append to a filename.  So if we create multiple variations of a module and write them out, the filenames do not collide.  

I was about to add the 'test instantiation' tests that do variations of parameterization on a component to make sure new names don't pop up under different variations.  In that case it is useful to turn on file writing as well.

lib/src/arithmetic/floating_point/floating_point_converter.dart
```diff
@@ -116,12 +116,15 @@ class FloatingPointConverter<FpTypeIn extends FloatingPoint,
...
-            Const(source.explicitJBit),
-            biasDiff +
-                source.exponent.zeroExtend(biasDiff.width) +
-                mux(~source.isNormal & Const(source.explicitJBit),
-                    Const(1, width: maxExpWidth), Const(0, width: maxExpWidth)),
-            biasDiff);
+                Const(source.explicitJBit),
+                biasDiff +
+                    source.exponent.zeroExtend(biasDiff.width) +
+                    mux(
```
mkorbel1: why not replace with `(source.isNormal & Const(source.explicitJBit)).zeroExtend(maxExpWidth)`

lib/src/arithmetic/floating_point/floating_point_converter.dart
```diff
@@ -132,13 +135,14 @@ class FloatingPointConverter<FpTypeIn extends FloatingPoint,
...
 
-      final trueShift = shift +
-          mux(
-              Const(destination.explicitJBit) &
-                  source.isNormal &
-                  ~Const(source.explicitJBit),
-              Const(-1, width: maxExpWidth),
-              Const(0, width: maxExpWidth));
+      final trueShift = (shift +
+              mux(
```
mkorbel1: why not replace with `(source.isNormal & Const(source.explicitJBit)).replicate(maxExpWidth)`

lib/src/arithmetic/floating_point/floating_point_converter.dart
mkorbel1: Adding a check for `__` or maybe `___` could be a nice check to add for the component library if we have a goal of prettier SV.  Maybe we can just run it on the same things generated by the configurator test?

lib/src/utils.dart
mkorbel1: Why do we need to write to a file?  If this is an automated check, we don't need to write to a file to check the `String`, and if it's a manual thing, we can just print it out ourselves to the console or to a file.

If we don't want to expose it on the API, for rohd-hcl, we should mark it with `@internal` or `@visibleForTesting` or move it to another file that's not `export`ed.  And if it must be a class, should it not be `abstract`?

lib/src/arithmetic/floating_point/floating_point_converter.dart
```diff
@@ -204,33 +211,35 @@ class FloatingPointConverter<FpTypeIn extends FloatingPoint,
...
-          mux(Const(source.explicitJBit), Const(-1, width: maxExpWidth),
-              Const(0, width: maxExpWidth)) +
-          mux(Const(destination.explicitJBit), Const(1, width: maxExpWidth),
-              Const(0, width: maxExpWidth));
+              biasDiff.gte(newSe),
+              source.isNormal.zeroExtend(maxExpWidth) + (biasDiff - newSe),
+              Const(0, width: maxExpWidth))
+          .named('nextShift');
+
+      final jBitAdjust = (mux(Const(source.explicitJBit),
```
mkorbel1: since this is all `Const`s, why not just evaluate it in software and throw a constant in?  why use `mux` instead of Dart ternary operators?

lib/src/arithmetic/fixed_to_float.dart
```diff
@@ -46,8 +46,13 @@ class FixedToFloat extends Module {
...
             definitionName: definitionName ??
                 'Fixed${fixed.width}ToFloatE${outFloat.exponent.width}'
-                    'M${outFloat.mantissa.width}') {
-    fixed = fixed.clone()..gets(addInput('fixed', fixed, width: fixed.width));
+                    'M$outFloat.mantissa.width') {
+    fixed = fixed.clone(name: 'fixed')
+      ..gets(addInput('fixed', fixed, width: fixed.width));
+
+    // TODO(desmonddak): fixed shows up as a swizzle
+    final fixedAsLogic = Logic(width: fixed.width, name: 'fixedAsLogic');
```
mkorbel1: you can use `fixed.packed`

lib/src/utils.dart
```diff
@@ -115,3 +116,33 @@ Logic condFlop(
...
+          .toSynthFileContents()
+          .join('\n');
+
+  /// Check for poor signal names which are those that have a `__` in them,
+  /// e.g., one that comprises multiple input signal names and an operation.
+  final poorSignalName = RegExp(r'\w+__[^(\d+)$]\w+');
+  final matches = poorSignalName.allMatches(systemVerilogString);
+
+  if (matches.isNotEmpty) {
+    stdout.writeln('${mod.definitionName} Poor Logic names:');
```
mkorbel1: nit: why not just `print`?

lib/src/arithmetic/fixed_to_float.dart
```diff
@@ -45,9 +45,13 @@ class FixedToFloat extends Module {
...
       : super(
             definitionName: definitionName ??
-                'Fixed${fixed.width}ToFloatE${outFloat.exponent.width}'
+                'Fixed${fixed.width}ToFloat_E${outFloat.exponent.width}'
                     'M${outFloat.mantissa.width}') {
-    fixed = fixed.clone()..gets(addInput('fixed', fixed, width: fixed.width));
+    fixed = fixed.clone(name: 'fixed')
+      ..gets(addInput('fixed', fixed, width: fixed.width));
+
+    // TODO(desmonddak): fixed shows up as a swizzle
```
mkorbel1: question: what is this TODO for?  I think this probably isn't something easily fixable without adding support for `struct` generation or full-struct-assignment stuff in ROHD

desmonddak: It was a reminder to check the internal name once we have the struct fixes in.  I think I can remove this as we will be looking at structure names once we have struct assignment.

lib/src/utils.dart
desmonddak: I'm only using this routine for SV naming checking with a loop I wrote elsewhere.  I think we should get systematic name checking in place later.

desmonddak: I'm removing this utility for now.

mkorbel1 (approved):
Looks good!

PR #242, opened 2025-07-17
Floating Point to Fixed and Fixed to Floating Point Conversion

sweeksBigBlue:
This pull request is a base line implementation of [PR 112](https://github.com/intel/rohd-hcl/issues/112).

In this PR we make a generalized converter for Fixed to Float and Float to Fixed method implementations. This should make it easier in the future for designers to implement transitions between the two elements and expand hardware design capabilities.


sweeksBigBlue:
@desmonddak  and @mkorbel1 we are finally ready to review this!

lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -178,6 +178,73 @@ class FixedPointValue implements Comparable<FixedPointValue> {
...
   }
 
+  /// Converts a [FixedPointValue] to a [FloatingPointValue].
+  FloatingPointValue toFloat() {
+    if (!value.isValid) {
+      throw RohdHclException('Inputs must be valid.');
+    }
+
+    // Qm.n need to conver so we have in some form (1.<MANTISSA>)
+    const minmialExponentWidth = 4;
```
mkorbel1, line 194: why is the minimum 4?

doc/components/fixed_point.md
```diff
@@ -4,7 +4,9 @@ Fixed-point binary representation of numbers is useful several applications incl
...
+A [FixedPointValue](https://intel.github.io/rohd-hcl/rohd_hcl/FixedPointValue-class.html) represents a signed or unsigned fixed-point value following the Q notation (Qm.n format) as introduced by [Texas Instruments](https://www.ti.com/lit/ug/spru565b/spru565b.pdf). It comprises an optional sign, integer part and/or a fractional part. [FixedPointValue](https://intel.github.io/rohd-hcl/rohd_hcl/FixedPointValue-class.html)s can be constructed from individual fields or from a Dart [double](https://api.dart.dev/stable/3.6.0/dart-core/double-class.html), converted to Dart [double](https://api.dart.dev/stable/3.6.0/dart-core/double-class.html), can be compared and can be operated on (+, -, *, /). [FixedPointValue] can now call toFloat method to switch from a fixed
```
mkorbel1: Don't need to say "can now", just describe it in the state after this PR is merged :)

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
```diff
@@ -461,4 +461,40 @@ class FloatingPointValue implements Comparable<FloatingPointValue> {
           sign.bitString, exponent.bitString, '${'0' * (mantissa.width - 1)}1');
     }
   }
+
+  /// Losslessly convert a [FloatingPointValue] to a [FixedPointValue].
+  /// TODO(jcfarwe):
```
mkorbel1: Did you mean to include these TODO's in this PR or do them?

Also, TODOs should be after `//` rather than `///` (since they are not part of the doc comment)

test/arithmetic/values/fixed_point_value_test.dart
```diff
@@ -289,4 +289,31 @@ void main() {
       }
     }
   });
+
```
mkorbel1, line 294: what about testing with vs without `explicitJBit`?

lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -178,7 +178,74 @@ class FixedPointValue implements Comparable<FixedPointValue> {
     return isNegative() ? -value : value;
   }
 
-  /// Addition operation that returns a [FixedPointValue].
```
mkorbel1, line 181: Your merge removed the brackets around `FixedPointValue`, should add it back

test/arithmetic/values/floating_point_value_test.dart
```diff
@@ -631,4 +631,79 @@ void main() {
       }
     });
   });
+  test('FPV: toFixedPointValue', () async {
+    //
+    // TODO(jcfarwe): Find all corner cases, automate test cases.
```
mkorbel1: lingering TODO?

test/arithmetic/values/floating_point_value_test.dart
```diff
@@ -631,4 +631,79 @@ void main() {
       }
     });
   });
+  test('FPV: toFixedPointValue', () async {
```
mkorbel1: maybe add an exhaustive run as well for at least some width(s)?

lib/src/arithmetic/values/fixed_point_value.dart
sweeksBigBlue: If it is less than 4 you get a weird exponent behavior and also the data translation just becomes kind of useless. So I decided on FP4 as the minimum.

test/arithmetic/values/fixed_point_value_test.dart
sweeksBigBlue: Do you mind expanding on that? Idk what that is but happy to add it to my tests.

lib/src/arithmetic/values/fixed_point_value.dart
sweeksBigBlue: Fixed!

test/arithmetic/values/fixed_point_value_test.dart
mkorbel1: https://github.com/intel/rohd-hcl/blob/main/doc/components/floating_point.md#explicit-j-bit

I'll test the documentation on you: if this doc doesn't explain it well enough, we should improve it!

doc/components/fixed_point.md
sweeksBigBlue: Removed the "can now" from it, let me know if this looks good.

test/arithmetic/values/fixed_point_value_test.dart
sweeksBigBlue: Great question! I think maybe this would be an excellent second issue I can follow-up on in a different PR maybe? Honest answer is, even with reading that I'd need time to digest that and maybe that should be an expanded method to this base one? Or maybe I'm over complicating that.

lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -178,7 +178,74 @@ class FixedPointValue implements Comparable<FixedPointValue> {
...
+  /// Converts a [FixedPointValue] to a [FloatingPointValue].
+  FloatingPointValue toFloat() {
```
desmonddak: Should this be toFloatingPointValue()?

sweeksBigBlue: Great point, I can fix that if you think that is more appropriate

test/arithmetic/values/fixed_point_value_test.dart
```diff
@@ -291,4 +291,31 @@ void main() {
...
+      final fxv = FixedPointValue.populator(
+              integerWidth: m, fractionWidth: n, signed: true)
+          .ofLogicValue(LogicValue.ofInt(i, width));
+      final fpv = fxv.toFloat();
+      expect(fpv.toDouble(), fxv.toDouble(),
+          reason: 'toFloatingPointValue failed for $i');
+    }
+  });
+
+  test('FixedPointValue: toFloatingPointValue unsigned', () {
```
desmonddak, line 311: You could add the signage as a loop in previous test and have one test.

Testing for m = 0, and n = 0 as two important corners.
```dart
for (final (m,n) in [(10,12), (0, 22), (22,0)] {
```

sweeksBigBlue: So I had this thought originally, so this is probably a picking nits thing for me, but I like tests to have a single focus so I personally prefer a signed versus unsigned testing approach. I'm okay changing those if you.

For the 0 and 0 test what should we expect? Is that for loop in reference to corner cases you want tested?

lib/src/arithmetic/values/fixed_point_value.dart
mkorbel1: what weird exponent behavior?  4 feels like a magic number

lib/src/arithmetic/values/fixed_point_value.dart
sweeksBigBlue: Updated name, it is more clear and obvious good catch.

test/arithmetic/values/fixed_point_value_test.dart
mkorbel1: I think it should be a relatively straight-forward change, basically even normal things look like sub-normals with an explicit `1`.  @desmonddak  what do you think?

lib/src/arithmetic/values/fixed_point_value.dart
sweeksBigBlue: That is fair, so if you do less than 4 that the exponent doesn't get translated correctly in the conversion behavior because it doesn't have the bits to expand. Might be easier if I showed it to you.

But overall, you end up without the wiggle room to translate it. Might not be an issue since I did the exponent with testing. I can run and see. But that is the reason I chose 4, to give enough width of translating to the exponent with bias.

test/arithmetic/values/fixed_point_value_test.dart
sweeksBigBlue: Okay, this one I will probably need some work with on it. I believe mine has the explicit 1, but just hadn't heard of this before so I just need some help learning if this is what we want to do :)

test/arithmetic/values/floating_point_value_test.dart
```diff
@@ -631,4 +631,79 @@ void main() {
...
+      [0, 8, 1, 0x000000, 23], // -1.0
+      [0, 8, 1, 0x400000, 23], // -1.5
+      [0, 5, 0, 0x000, 10], // 1.0, 16-bit float
+      [-127, 8, 0, 0x000001, 23], // 1e-45
+      [-127, 8, 0, 0x000000, 23], // 0
+      [-127, 8, 1, 0x000000, 23], // -0
+    ];
+
+    for (final testCase in testCases) {
+      final bias = (pow(2, testCase[1] - 1) - 1).toInt();
```
desmonddak: Removing the sizes and using the FPV populator might be more readable for this test.
It's a bit confusing whether the sizes are for the FPV or the FXV.

See compound_adder_test.dart:262 for an example of list structuring and using Boolean literals.  It's not great but it helps to structure the list to help the reader parse (the structure has separated inputs and outputs, for example).

test/arithmetic/values/fixed_point_value_test.dart
desmonddak: FPV supports an explicitJBit parameter which means the JBit is not implicit, but rather stored in the mantissa.
So you just need to add this variant in your testing to make sure this works (and if not, fix the converter to check for explicitJBit.
The way to approach this is that subnormal numbers have an explicit JBit, but normals do not.  The flag would tell you that your normal number needs to have the leading '1' stored as the first bit of the mantissa and that there is no implicit leading 1.

lib/src/arithmetic/values/fixed_point_value.dart
sweeksBigBlue: I think I'm on a way to solve this. I need to make sure that the exponentWidth, once added to the bias, isn't all 1's. Just figuring out how to check that (don't have a good idea yet).

lib/src/arithmetic/values/fixed_point_value.dart
```diff
@@ -178,7 +178,74 @@ class FixedPointValue implements Comparable<FixedPointValue> {
...
+
+    // if we have a zero value, we can return a zero floating point value
+    if (fixedNum.isZero) {
+      return FloatingPointValue(
+          sign: LogicValue.zero,
+          exponent: LogicValue.zero,
+          mantissa: LogicValue.zero);
+    }
+
+    // FOR loop to find first one to figure out correct exponent value and width
```
desmonddak: On comments: this can confuse people.   Find the first `1` .would be better, or 'Count the leading zeros in order to correct the exponent value.' may be better still.  No need to say 'FOR loop'.  :j)

If you like to use one, then use it like this `one` and `zero`.

Remember to end on a period (Dart conventions).

lib/src/arithmetic/values/fixed_point_value.dart
sweeksBigBlue: Yeah, I can't get it working without the minimal check for the exponent width to a solid value. Else I'm just chasing corner cases so maybe this part is just FUBAR code on my part. Sorry guys.

test/arithmetic/values/fixed_point_value_test.dart
sweeksBigBlue: I think I follow. I'm still a bit confused because I think I implicitly imply a j-bit? So you are saying we need to add that to the mantissa for subnormals? So I have to check somehow if the number is a normal or not?

lib/src/arithmetic/values/fixed_point_value.dart
desmonddak: Can you just put in a failing test and comment out the expect for us to see?

sweeksBigBlue: @desmonddak I remembered what it was now, and I do have an example. It is essentially the -1 problem. Because we have to handle the negative to positive representation, the minimum width is actually mathematically 3 (I chose four to represent the smallest FP standard I know of which is FP4). So if you drop below 3, you limit greatly your max representation and will end up with erroneous NaNs/INF. To avoid this, you need a ta minimum of an exponent width of 3. Happy to demo that for you Monday as I plan on working on some of your other suggestions this weekend. But essentially, if you want a truly generalizable value, you will need a minimum width of 3 for your exponent.

test/arithmetic/values/fixed_point_value_test.dart
desmonddak: Sorry for the late reply:  it should be quite simple; a small change in how you adjust the mantissa in each direction.
So going from fixed value to normal float value:  you would traditionally compute the mantissa for the floating point value and chop off the leading '1' and store the rest of the mantissa into the FPV.

But going to an explicit j-bit form, you would keep that leading 1 and put it into the mantissa.  The FPV feels effectively one bit shorter because you need to store the leading '1.  Note that you would do this just as you do with subnormals.  In other words, if you fixed point was small enough to require a subnormal representation in FPV, you would also keep that leading '1'.

Going from an explicitJBit normal float to fixed, you would recognize that you don't need to prefix the mantissa with a '1' representing the j-bit.  The leading '1' is already in the mantissa.  This is just like it is for subnormal.  If you are converting a subnormal float to fixed, you do not prefix the mantissa with a '1'.

To test, you need to provide an explicitJBit form of floatingPointValue to translate to fixed and conversely, indicate that you want an explicitJBit form of floatingPointValue when converting from fixed.

Explicit JBIT is experimental:  we recognize that we do not have to normalize in some situations as it doesn't produce any more accuracy (say adding two narrow mantissa FPs to get a wider mantissa FP).  Rather than normalizing, we can leave it in explicit j-bit form for the next operation and save a normalization step, and let normalization happen after the next operation.

lib/src/arithmetic/values/floating_point_values/floating_point_value.dart
jcfarwe: Resolved

test/arithmetic/values/floating_point_value_test.dart
jcfarwe: Resolved

test/arithmetic/values/floating_point_value_test.dart
jcfarwe: Added exhaustive test for various exponent and mantissa widths

test/arithmetic/values/fixed_point_value_test.dart
jcfarwe: Something like this;
  test('FixedPointValue: toFloatingPointValue signed', () {
    const m = 10;
    const n = 12;
    const width = m + n + 1; // 1 for sign bit
    for (var i = 0; i < pow(2, width); i++) {
      final testVal = (i / pow(2, width)).toInt();
      final fxv = FixedPointValue.populator(
              integerWidth: m, fractionWidth: n, signed: true)
          .ofLogicValue(LogicValue.ofInt(testVal, width));
      FloatingPointValue fpv;
      if (testVal > 0) {
        fpv = fxv.toFloatingPointValue(false); // explicitJBit
      } else {
        fpv = fxv.toFloatingPointValue(true);
      }
      expect(fpv.toDouble(), fxv.toDouble(),
          reason: 'toFloatingPointValue failed for $testVal');
    }
  });
?

lib/src/arithmetic/values/fixed_point_value.dart
sweeksBigBlue: Fixed comments, good call :)

test/arithmetic/values/fixed_point_value_test.dart
sweeksBigBlue: Added edge cases to fixed point tests, all except `[0,0]`. Waiting for feedback on expected behavior.

test/arithmetic/values/fixed_point_value_test.dart
sweeksBigBlue: @jcfarwe has pushed tests for J-bit, ready for review @desmonddak and @mkorbel1

test/arithmetic/values/floating_point_value_test.dart
```diff
@@ -631,4 +631,151 @@ void main() {
...
+      final mantWidths = [10, 12, 17, 23];
+
+      for (final expWidth in expWidths) {
+        for (final mantWidth in mantWidths) {
+          final minExp = -1 << (expWidth - 1);
+          final maxExp = (1 << (expWidth - 1)) - 1;
+          for (var testExp = minExp; testExp < maxExp; testExp++) {
+            final bias = (pow(2, expWidth - 1) - 1).toInt();
+            final exp = LogicValue.ofInt(testExp, expWidth);
+            final sign = LogicValue.ofInt(0, 1);
```
desmonddak: Do you also want to put 'sign in a loop:
```dart
for (final sign in [LogicValue.zero, Logicvalue.one]) {
```

jcfarwe: Resolved

test/arithmetic/values/fixed_point_value_test.dart
desmonddak: Tests look good.  Can you merge the signed/unsigned versions into one?

desmonddak: Can you shorten the widths to speed up the tests?  22 bits is quite wide.

I tried the following which ran in about 1s:
```dart
    for (final (m, n) in [(5, 7), (7, 5), (0, 7), (7, 0)]) {
```

test/arithmetic/values/fixed_point_value_test.dart
desmonddak: I had the experience of missing a case when I used the word exhaustive.  One idea is to put these different focus tests in a group that adds up to being exhaustive.

test/arithmetic/values/fixed_point_value_test.dart
```diff
@@ -291,4 +291,36 @@ void main() {
...
+          final fpv = fxv.toFloatingPointValue(explicitJBit: explicitJBit);
+          expect(fpv.toDouble(), fxv.toDouble(),
+              reason: 'toFloatingPointValue failed for $i');
+        }
+      }
+    }
+  });
+
+  test('FixedPointValue: toFloatingPointValue unsigned', () {
+    for (final (m, n) in [(10, 12), (0, 22), (22, 0)]) {
```
desmonddak, line 312: Definitely narrow these widths...

test/arithmetic/values/floating_point_value_test.dart
```diff
@@ -631,4 +631,152 @@ void main() {
       }
     });
   });
```
desmonddak, line 633: We like to see a blank line between tests or groups.

test/arithmetic/values/floating_point_value_test.dart
```diff
@@ -631,4 +631,152 @@ void main() {
...
+      final nLen = exp.isNegative ? mantSize - exp : mantSize;
+      final mLen = finalMantissa.width - nLen - 1; // one bit for sign
+      return FixedPointValue.populator(
+        integerWidth: mLen,
+        fractionWidth: nLen,
+        signed: true,
+      ).ofLogicValue(finalMantissa);
+    }
+
+    test('FPV: toFixedPointValue exhaustive', () async {
```
desmonddak, line 664: Something is fishy about this test.  It only fills the upper 20 bits or so of the integer portion, yet the test is labeled 'exhaustive'.    

Here is what I see when I print `fxv`:
```
(0 0010000000000000000101101100000000000000000000000000000000000000000000000000000000000000000000 00000000000000000000000)
(1 1101111111111111111010010100000000000000000000000000000000000000000000000000000000000000000000 00000000000000000000000)
```

test/arithmetic/values/fixed_point_value_test.dart
desmonddak: I wasn't sure what feedback you wanted:  m=0, n=0 should probably throw.

PR #243, opened 2025-07-25
dot-product-component

desmonddak:
## Description & Motivation

This is the addition of a dot-product component that builds up a set of partial products in a single compression tree to compute a dot product all at once instead of having separate adders for each multiplication.
One possible extension is to add pipelining before the final add, but this may not create a very balanced pipeline for wide dot-products.

## Related Issue(s)

None.

## Testing

Exhaustive testing of a single dot-product, including signage control

## Backwards-compatibility

No.

## Documentation

Yes.  The component was added to the README.md and details appended to multiplier.md.


lib/src/arithmetic/dotproduct.dart
```diff
@@ -0,0 +1,112 @@
...
+  /// false, or unsigned) or dynamically with a 1-bit [Logic] input. Passing
+  /// something other null, bool, or [Logic] will result in a throw.
+  ///
+  ///
+  /// The optional [signedMultiplier] parameter configures the [multipliers]
+  /// statically using a bool to indicate a signed multiplier (default is false,
+  /// or unsigned) or dynamically with a 1-bit [Logic] input.  Passing something
+  /// other null, bool, or [Logic] will result in a throw.
+  DotProduct(List<Logic> multiplicands, List<Logic> multipliers,
+      {int radix = 4,
```
mkorbel1: Might be good to mention in the doc comment the purpose of `radix` and `adderGen`

lib/src/arithmetic/dotproduct.dart
```diff
@@ -0,0 +1,112 @@
...
+      super.reserveDefinitionName = false,
+      String? definitionName})
+      : super(
+            definitionName:
+                definitionName ?? 'DotProduct_W{$multipliers[0].width}_') {
+    if (multipliers.length != multiplicands.length) {
+      throw RohdHclException(
+          'Number of multipliers and multiplicands must be equal.');
+    }
+    for (var i = 1; i < multipliers.length; i++) {
```
mkorbel1: Alternatively, you could do something like:
```dart
final numDifferentWidths =  [...multipliers, ...multiplicands].map((e) => e.width).toSet().length;
```
though this would give a potentially less helpful error message than what you have.  Either way is fine, just shared FYI

lib/src/arithmetic/dotproduct.dart
```diff
@@ -0,0 +1,112 @@
...
+      }
+    }
+    // Enforce square products.
+    for (var i = 0; i < multipliers.length; i++) {
+      if (multipliers[i].width != multiplicands[i].width) {
+        throw RohdHclException('Multiplier and multiplicand at index $i '
+            'must have the same width.');
+      }
+    }
+    if (radix != 2 && radix != 4 && radix != 8 && radix != 16) {
```
mkorbel1: suggestion: a little more readable might be 
```dart
const legalRadices = {2, 4, 8, 16};
if(!legalRadices.contains(radix)) {
  throw RohdHclException('Radix must be in $legalRadices');
}
```

lib/src/arithmetic/dotproduct.dart
```diff
@@ -0,0 +1,112 @@
...
+      throw RohdHclException('Radix must be 2, 4, 8, or 16.');
+    }
+
+    final signedMultiplicandParameter =
+        StaticOrRuntimeParameter.ofDynamic(signedMultiplicand);
+    final signedMultiplierParameter =
+        StaticOrRuntimeParameter.ofDynamic(signedMultiplier);
+
+    multiplicands = [
```
mkorbel1: suggestion: a couple cool tricks you might try

```dart
    multiplicands = [
      for (final (i, multiplicand) in multiplicands.indexed)
        addInput(
            'multiplicand_$i', multiplicand,
            width: multiplicand.width)
    ];
```

```dart
    multiplicands = multiplicands.mapIndexed((i, multiplicand) =>
        addInput(
            'multiplicand_$i', multiplicand,
            width: multiplicand.width)).toList()
```

lib/src/arithmetic/dotproduct.dart
```diff
@@ -0,0 +1,112 @@
...
+      ppg.rowShift.addAll(ppgNext.rowShift);
+      return ppg;
+    });
+    final vec = [
+      for (var row = 0; row < ppg.rows; row++)
+        ppg.partialProducts[row].rswizzle()
+    ];
+    final columnCompressor = ColumnCompressor(vec, ppg.rowShift);
+    final adder = adderGen(columnCompressor.add0, columnCompressor.add1);
+    final sum = adder.sum.slice(adder.sum.width - 3, 0);
```
mkorbel1, line 146: question: why slice off the upper 3 bits?  also, what is the width of the final `product` relative to the inputs?  does it expand or stay the same?  might be worth documenting.

lib/src/arithmetic/multiplier.dart
```diff
@@ -74,18 +74,15 @@ abstract class Multiplier extends Module {
   /// multiplication result.
   ///
   /// The optional [signedMultiplicand] parameter configures the multiplicand
-  /// [a] statically using a [bool] as a signed multiplicand (default is
-  /// `false`, or unsigned) or dynamically with a 1-bit [Logic]
-  /// [selectSignedMultiplicand] input. You can pass either a [bool] (for static
-  /// configuration) or a [Logic] (dynamically configuring the type handled)
-  /// with a signal to this parameter, otherwise this constructor will throw.
+  /// [a] statically using a bool to indicate a signed multiplicand (default is
```
mkorbel1: suggestion: use `s or []s around bool.

mkorbel1 (commented):
This is awesome!  Few minor notes/suggestions/questions

lib/src/arithmetic/dotproduct.dart
desmonddak: I tried to learn some more functional style and used something that still generates the position.  Not really very intuitive, but I agree that looping feels redundant.

lib/src/arithmetic/dotproduct.dart
desmonddak: I find this elsewhere.  I put a static array in the limiting component and reference that.

lib/src/arithmetic/dotproduct.dart
desmonddak: Better.

lib/src/arithmetic/multiplier.dart
desmonddak: I found a bunch of these.  Also they are in another PR (the breaking one on multiplierComponents).  I hope to remember to clean these up.

lib/src/arithmetic/dotproduct.dart
```diff
@@ -0,0 +1,126 @@
...
+      super.reserveDefinitionName = false,
+      String? definitionName})
+      : super(
+            definitionName:
+                definitionName ?? 'DotProduct_W{$multipliers[0].width}_') {
+    if (multipliers.length != multiplicands.length) {
+      throw RohdHclException(
+          'Number of multipliers and multiplicands must be equal.');
+    }
+    final candWidthMiss = multiplicands
```
mkorbel1, line 63: I don't understand how this `candWidthMiss` uses `length` to determine the index of the first (or any) index where the widths mismatched?

lib/src/arithmetic/dotproduct.dart
mkorbel1: In that case, can we just leave it to the sub-component to throw the error message rather than check it again inside `DotProduct`?

lib/src/arithmetic/dotproduct.dart
desmonddak: `candWidthMiss` produces a vector of `true`s whenever the `multiplicand` width matches the `multiplier` width.  If that vector is less than the `multiplicands` length, then it is a failure.  The `where` trims the list to just `true`s.  

Unfortunately it counts the mismatches rather than stopping at the first, so I'll adjust the message for now.

I'll add a test to ensure it throws.

test/arithmetic/dotproduct_test.dart
```diff
@@ -0,0 +1,298 @@
...
+      Logic(width: 4, name: 'b_0'),
+      Logic(width: 4, name: 'b_1'),
+      Logic(width: 4, name: 'b_2'),
+      Logic(width: 4, name: 'b_3')
+    ];
+
+    CompressionTreeDotProduct(multipliers, multipliers);
+    try {
+      CompressionTreeDotProduct(multiplicands, multipliers);
+      expect(true, throwsA(isA<RohdHclException>()));
```
mkorbel1: i think you just want `fail()` here

mkorbel1 (approved):
Looks great!  Just that one minor tweak in the test

lib/src/arithmetic/dotproduct.dart
mkorbel1: ah yeah your change makes sense to me now :)

mkorbel1 (approved):
Looks great!

PR #245, opened 2025-08-07
update Interface classes to include the required clone() method.

desmonddak:
## Description & Motivation

Updated our interface classes in use in ROHD-HCL to deprecate any former static clone routines and add in the 
clone() overrides on instance classes that inherit from Interface.

## Related Issue(s)

None

## Testing

Ran dart test.

## Backwards-compatibility

No

## Documentation

Not required.

### DIFF
--- a/lib/src/arithmetic/divider.dart
+++ b/lib/src/arithmetic/divider.dart
@@ -92,6 +92,7 @@ class MultiCycleDividerInterface extends PairInterface {
         ]);
 
   /// A match constructor for the divider interface.
+  @Deprecated('Use clone() instead.')
   MultiCycleDividerInterface.match(MultiCycleDividerInterface other)
       : this(dataWidth: other.dataWidth);
 
--- a/lib/src/arithmetic/divider.dart
+++ b/lib/src/arithmetic/divider.dart
@@ -94,6 +94,11 @@ class MultiCycleDividerInterface extends PairInterface {
   /// A match constructor for the divider interface.
   MultiCycleDividerInterface.match(MultiCycleDividerInterface other)
       : this(dataWidth: other.dataWidth);
+
+  /// Clones this [MultiCycleDividerInterface].
+  @override
+  MultiCycleDividerInterface clone() =>
+      MultiCycleDividerInterface(dataWidth: dataWidth);
 }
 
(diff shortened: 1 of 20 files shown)


lib/src/arithmetic/divider.dart
```diff
@@ -94,6 +94,11 @@ class MultiCycleDividerInterface extends PairInterface {
   /// A match constructor for the divider interface.
   MultiCycleDividerInterface.match(MultiCycleDividerInterface other)
       : this(dataWidth: other.dataWidth);
+
+  /// Clones this [MultiCycleDividerInterface].
+  @override
```
mkorbel1, line 100: deprecate the `match`

lib/src/arithmetic/multiplier_components/partial_product_generator.dart
```diff
@@ -23,6 +23,8 @@ class SignBit extends Logic {
       : super(name: '${inl.name}_signbit', naming: Naming.mergeable) {
     this <= inl;
   }
+  @override
+  SignBit clone({String? name = 'clone'}) => SignBit(this, inverted: inverted);
```
mkorbel1: don't provide a default for the name, and also the name should get passed through as an argument to the main constructor so that it can properly be renamed.  if no name is provided when cloning, you should keep the existing name.

lib/src/clock_gating.dart
```diff
@@ -27,6 +28,10 @@ class ClockGateControlInterface extends PairInterface {
   /// generated.
   final bool isPresent;
 
+  /// Capture the additional ports that are part of this interface, if any.
+  @protected
```
mkorbel1: i think this doesn't need to be `@protected` nor `late`

lib/src/arithmetic/divider.dart
desmonddak: We call this on line 143.  What should I replace this with?

desmonddak: I presume this is just a clone.

mkorbel1: yea `match` and `clone` were two needlessly different conventions that showed up in places, its the same

PR #246, opened 2025-08-08
convert a boolean signExtend to a runtime or static parameter

desmonddak:
## Description & Motivation

Converted `signExtend` boolean to a StaticOrRuntimeParameter.

## Testing

Ran existing boolean tests.  Should extend to at least one Logic runtime test.

## Backwards-compatibility

No as existing calls using boolean will be automatically converted internally.

## Documentation

No.  Will look to see if we need to adjust the markdown regarding the sign extension parameter.

### DIFF
--- a/lib/src/reduction_tree.dart
+++ b/lib/src/reduction_tree.dart
@@ -9,6 +9,7 @@
 
 import 'dart:math';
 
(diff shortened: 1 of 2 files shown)


mkorbel1 (approved):
Looks good to me!

Issue #170, opened 2025-02-06
Count pattern occurrence


mkorbel1:
### Motivation

Similar to the [Count bit occurrence](https://github.com/intel/rohd-hcl/blob/main/doc/components/count.md), but instead, find an entire multi-bit pattern

### Desired solution

A component that can count instances of a fixed-width pattern

izzianyramli:
Hello, can we work on this issue?

PR #247, opened 2025-08-11
Add Count Pattern Occurrence

izzianyramli:
## Description & Motivation

Similar to the [Count bit occurrence](https://github.com/intel/rohd-hcl/blob/main/doc/components/count.md), but instead, find an entire multi-bit pattern

## Related Issue(s)

Count pattern occurrence https://github.com/intel/rohd-hcl/issues/170

## Testing

Test are defined in test/count_pattern_test.dart with multiple cases

## Backwards-compatibility

No

## Documentation

Yes, documentation on how to use it, and some examples are added in doc/components/count_pattern.md

### DIFF
--- a/test/count_pattern_test.dart
+++ b/test/count_pattern_test.dart
@@ -17,13 +17,15 @@ void main() {
     test('present once', () {
       final bus = Const(bin('00111110'), width: 8);
       final pattern = Const(bin('110'), width: 3);
-      final countPattern = CountPattern(bus, pattern);
+      final countPattern = CountPattern(bus, pattern, generateError: true);
+      expect(countPattern.error!.value.toInt(), equals(0));
       expect(countPattern.count.value.toInt(), equals(1));
     });
     test('present more than once', () {
       final bus = Const(bin('00110110'), width: 8);
       final pattern = Const(bin('01'), width: 2);
-      final countPattern = CountPattern(bus, pattern);
+      final countPattern = CountPattern(bus, pattern, generateError: true);
+      expect(countPattern.error!.value.toInt(), equals(0));
       expect(countPattern.count.value.toInt(), equals(2));
     });
     test('not present', () {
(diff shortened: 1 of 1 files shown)


lib/src/count_pattern.dart
```diff
@@ -0,0 +1,81 @@
...
+  /// [error] is a getter for error in CountPattern and is generated when
+  /// [generateError] is `true`.
+  /// When pattern is not found, it will result in error `1`.
+  Logic? get error => tryOutput('error');
+
+  /// Count the number of occurence of a fixed-width pattern in a bus.
+  ///
+  /// Takes in [bus] of type [Logic].
+  /// [pattern] is the pattern to be counted in the bus.
+  /// If [fromStart] is `true`, the count starts from the beginning of the bus.
```
mkorbel1, line 43: What is the meaning of counting from the start or the end?  Shouldn't the resulting count be the same?

lib/src/count_pattern.dart
```diff
@@ -0,0 +1,81 @@
...
+      final valCheck = bus.getRange(minBit, maxBit).eq(pattern);
+      count += valCheck.zeroExtend(count.width);
+    }
+
+    _output = addOutput('count', width: count.width);
+    _output <= count;
+
+    if (generateError) {
+      // If pattern is not found (count equals to 0), return error
+      addOutput('error');
```
mkorbel1, line 77: why is it an error if count is 0?

lib/src/count_pattern.dart
```diff
@@ -0,0 +1,81 @@
...
+        minBit = bus.width - i - pattern.width;
+        maxBit = bus.width - i;
+      }
+
+      // Check if pattern matches, add to count
+      final valCheck = bus.getRange(minBit, maxBit).eq(pattern);
+      count += valCheck.zeroExtend(count.width);
+    }
+
+    _output = addOutput('count', width: count.width);
```
mkorbel1, line 72: what's the purpose of the internal `_output` vairable?  Why not just assign `this.count <= count` or something?

lib/src/count_pattern.dart
```diff
@@ -0,0 +1,81 @@
...
+
+  /// [error] is a getter for error in CountPattern and is generated when
+  /// [generateError] is `true`.
+  /// When pattern is not found, it will result in error `1`.
+  Logic? get error => tryOutput('error');
+
+  /// Count the number of occurence of a fixed-width pattern in a bus.
+  ///
+  /// Takes in [bus] of type [Logic].
+  /// [pattern] is the pattern to be counted in the bus.
```
mkorbel1, line 42: it may be worth explicitly stating that if the pattern width is greater than the width of the bus, the count will always be 0

PR #250, opened 2025-08-23
Fix package resolution

desmonddak:
## Description & Motivation

tool/gh_action/verify_formatting.sh results in warnings because dart packages are not up to date for confapp/.

## Related Issue(s)

None.

## Testing

Ran tool/gh_action/verify_formatting.sh

## Backwards-compatibility

No.

## Documentation

None required.


tool/gh_actions/install_dependencies.sh
```diff
@@ -12,3 +12,6 @@
 set -euo pipefail
 
 dart pub get
+
+cd confapp
+flutter pub get
```
mkorbel1, line 17: we already have this other file:
tool/gh_actions/install_dependencies_generator_site.sh

should we just update/use this?

desmonddak: This fix does not work as flutter is not in this container.
At issue is that `dart pub get` has a lot of warnings (no failure) which I was trying to fix.  But we can't call flutter here.
Since confapp has flutter dependencies we get warnings.

desmonddak: I may just drop this PR.

PR #251, opened 2025-08-28
Refactoring AXI to support protocol variants (AXI-Lite, AXI-S, ACE, ACE-Lite)

kimmeljo:
## Description & Motivation

Adding support for other common AMBA interfaces that are mostly derivatives of AXI. In some cases (ACE), this involves enhancing the BFM.

## Related Issue(s)

N/A

## Testing

Adding unit tests using both existing and new BFMs.

## Backwards-compatibility

The changes to AXI itself should be fully backwards compatible.

## Documentation

Yes, included.

### DIFF
--- a/lib/src/interfaces/amba4/ace4_lite.dart
+++ b/lib/src/interfaces/amba4/ace4_lite.dart
@@ -1,0 +1,141 @@
+// Copyright (C) 2024-2025 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// ace4_lite.dart
+// Definitions for the ACE-Lite extension on the AXI interface.
+//
+// 2025 August
+// Author: Josh Kimmel <joshua1.kimmel@intel.com>
+
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+
+/// ACE-Lite read interface.
+///
+/// This is mostly the same as AXI-4 but with some coherency additions.
+class Ace4LiteReadInterface extends Axi4BaseReadInterface {
+  /// Width of the coherency domain signal.
+  final int domainWidth;
+
+  /// Should the ARBAR signal be present.
+  final bool useBar;
+
+  /// Coherency domain.
+  ///
+  /// Width is equal to [domainWidth].
+  Logic? get arDomain => tryPort('ARDOMAIN');
+
+  /// Coherency barrier.
+  ///
+  /// Width is always 1.
+  Logic? get arBar => tryPort('ARBAR');
+
+  /// Constructor.
+  ///
+  /// Should match regular AXI4 but with DOMAIN and BAR.
+  Ace4LiteReadInterface({
+    super.idWidth = 4,
+    super.addrWidth = 32,
+    super.lenWidth = 8,
+    super.dataWidth = 64,
+    super.aruserWidth = 32,
+    super.ruserWidth = 32,
+    super.useLock = true,
+    super.useLast = true,
+    this.domainWidth = 1,
+    this.useBar = true,
+  }) : super(
+          sizeWidth: 3,
+          burstWidth: 2,
+          cacheWidth: 4,
+          protWidth: 3,
+          qosWidth: 4,
+          regionWidth: 4,
+          rrespWidth: 2,
+        ) {
+    setPorts([
+      if (domainWidth > 0) Logic.port('ARDOMAIN', domainWidth),
+      if (useBar) Logic.port('ARBAR'),
+    ], [
+      Axi4Direction.fromMain,
+    ]);
+  }
+
+  /// Copy constructor.
+  Ace4LiteReadInterface clone() => Ace4LiteReadInterface(
+      idWidth: idWidth,
+      addrWidth: addrWidth,
+      lenWidth: lenWidth,
+      dataWidth: dataWidth,
+      aruserWidth: aruserWidth,
+      ruserWidth: ruserWidth,
+      useLock: useLock,
+      useLast: useLast,
+      domainWidth: domainWidth,
+      useBar: useBar);
+}
+
+/// ACE-Lite write interface.
+///
+/// This is mostly the same as AXI-4 but with some coherency additions.
+class Ace4LiteWriteInterface extends Axi4BaseWriteInterface {
+  /// Width of the coherency domain signal.
+  final int domainWidth;
+
+  /// Should the ARBAR signal be present.
+  final bool useBar;
+
+  /// Coherency domain.
+  ///
+  /// Width is equal to [domainWidth].
+  Logic? get arDomain => tryPort('AWDOMAIN');
+
+  /// Coherency barrier.
+  ///
+  /// Width is always 1.
+  Logic? get arBar => tryPort('AWBAR');
+
+  /// Constructor.
+  ///
+  /// Should match regular AXI4 but with DOMAIN and BAR.
+  Ace4LiteWriteInterface({
+    super.idWidth = 4,
+    super.addrWidth = 32,
+    super.lenWidth = 8,
+    super.dataWidth = 64,
+    super.awuserWidth = 32,
+    super.wuserWidth = 32,
+    super.buserWidth = 16,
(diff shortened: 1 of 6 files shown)


lib/src/interfaces/amba4/axi4_s.dart
```diff
@@ -0,0 +1,134 @@
...
+// axi4_s.dart
+// Definitions for the AXI-S interface.
+//
+// 2025 August
+// Author: Josh Kimmel <joshua1.kimmel@intel.com>
+
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+
+/// A standard AXI4 read interface.
```
mkorbel1, line 13: nit: AXI4-Stream in the comment

lib/src/interfaces/amba4/ace4_lite.dart
```diff
@@ -0,0 +1,141 @@
...
+// 2025 August
+// Author: Josh Kimmel <joshua1.kimmel@intel.com>
+
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+
+/// ACE-Lite read interface.
+///
+/// This is mostly the same as AXI-4 but with some coherency additions.
+class Ace4LiteReadInterface extends Axi4BaseReadInterface {
```
mkorbel1: why doesnt this extend Axi4 instead of Axi4Base?

mkorbel1 (commented):
Looking great!

Question to think about: do we want to consider using `PairInterface` and make address/data channels as subInterfaces?  Might make BFM etc. easier.

doc/components/amba/lti.md
```diff
@@ -0,0 +1,41 @@
+# LTI
+
+ROHD-HCL contains a bunch of collateral for LTI. The collateral includes interface definitions, monitors, drivers, trackers, compliance checkers, BFMs, and HW implementations.
+
+For cross-referencing against the official LTI specification:
+
+- Any LTI collateral should be aligned with LTI specification `D`.
```
mkorbel1: for this and axi, might be good to just include links to these specs as well?

lib/src/interfaces/amba4/axi4.dart
```diff
@@ -0,0 +1,880 @@
...
+//
+// 2025 January
+// Author: Josh Kimmel <joshua1.kimmel@intel.com>
+
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/src/exceptions.dart';
+
+/// A grouping of signals on the AXI
+/// interfaces based on direction.
+enum Axi4Direction {
```
mkorbel1: It looks like this isn't used anymore, remove?

lib/src/interfaces/amba5/axi5_base.dart
```diff
@@ -0,0 +1,1580 @@
...
+  Logic? get rp => tryPort('${prefix}RP');
+
+  /// Transfer using a shared credit.
+  ///
+  /// Width is always 1.
+  Logic? get sharedCrd => tryPort('${prefix}SHAREDCRD');
+
+  /// Give one credit on the given resource plane.
+  ///
+  /// Width is always 1.
```
mkorbel1, line 72: spec says this should be `numRp` width, not 1?

lib/src/interfaces/amba5/axi5_base.dart
```diff
@@ -0,0 +1,1580 @@
...
+    this.sharedCredits = false,
+    this.numRp = 0,
+  }) {
+    setPorts([
+      if (!useCrediting) Logic.port('${prefix}READY'),
+      if (useCrediting) Logic.port('${prefix}CRDT'),
+      if (useCrediting && sharedCredits) Logic.port('${prefix}CRDTSH')
+    ], [
+      if (main) PairDirection.fromConsumer,
+      if (!main) PairDirection.fromProvider,
```
mkorbel1: suggestion: generally, you could just use `else` instead of `if(!main)`

mkorbel1 (commented):
WIP reviewing

doc/components/amba/lti.md
kimmeljo: Yeah good call!

lib/src/interfaces/amba5/axi5_base.dart
kimmeljo: Ah yup that's right!

lib/src/interfaces/amba5/axi5_base.dart
kimmeljo: Can you give a specific example of the syntax? I wanted to do something like that but it didn't work (syntax error to simply replace `if(!main)` with `else`)

desmonddak:
``` dart
      if (main) PairDirection.fromConsumer else PairDirection.fromProvider,
```
or
```dart
      main ? PairDirection.fromProvider : PairDirection.fromConsumer,
```

test/amba/amba5/axi5_bfm_test.dart
desmonddak, line 1: This test is quite noisy with lots of INFO about Stream with ID 0 has completed - dropping.  INFO messages.

lib/src/interfaces/amba4/ace4_lite.dart
desmonddak, line 1: Files that were created in 2025 should have a 2025 copyright (not 2024-2025)

test/amba/amba4/axi4_bfm_test.dart
desmonddak, line 1: This test has [SEVERE] compliance messages (a lot of them).

kimmeljo: There is a particular flavor of unit test in there that is meant to test if the compliance checkers are working - i.e., actually flagging errors that they should flag. These tests are "supposed" to fire off SEVERE messages if they are working - that is why they don't actually "fail" per se.

I cleared this methodology with Max a while back.

test/amba/amba5/axi5_bfm_test.dart
kimmeljo: Ah we should probably up the log level to "WARNING"

lib/src/interfaces/amba5/axi5_base.dart
kimmeljo: Will give this a try - thanks!

lib/src/interfaces/amba4/ace4_lite.dart
kimmeljo: To be honest, in some cases, I'm not actually sure if files were created in 2024 or 2025. I guess I can cross reference the git history for every file but does this really matter enough to spend that time? Will keep that in mind moving forward.

test/amba/amba4/axi4_bfm_test.dart
desmonddak: OK, but that means if I run dart test, I cannot see clearly where things failed when they do.  Can we modify to do an expect on the 'fail' instead of printing it, also for the INFO?  I'm pretty sure we are going to come back and realize this is too much noise for future use.  Do you need the messages?  Do we have a logging methodology for generating log files instead of standard out?

lib/src/interfaces/amba4/ace4_lite.dart
desmonddak: OK.

test/amba/amba4/axi4_bfm_test.dart
kimmeljo: How's this - for those tests only, I've turned the log level to OFF. So there are no longer any prints. But the test still "passes" in that the compliance checker flagged something that resulted in a test failure.

desmonddak: This helps a lot.  test/amba/amba5/lti_bfm_test.dart simple run still issues INFO output.  It's nice to have none as we can better see which test is running (it spins in place on the command line) and if a failure crops up, we get a new set of lines, so it is a nice UI without any noise if everything is working.

kimmeljo: Ah, yes it does! Just pushed a fix.

PR #253, opened 2025-09-04
Resettable entries

mkorbel1:
## Description & Motivation

Adds a `mixin` so that structures with indexed entries can be reset all together to the same value or to individual values via constants, signals, lists, and maps.

Also, fixed a bug where power-of-two depth FIFOs `occupancy` would be too small.

## Related Issue(s)

N/A

## Testing

Added extensive testing

## Backwards-compatibility

No

## Documentation

API docs are updated


lib/src/shift_register.dart
```diff
@@ -107,7 +87,7 @@ class ShiftRegister extends Module {
       final stageI = addOutput(_stageName(i), width: width);
       conds.add(stageI < dataStage);
 
-      resetValues?[stageI] = resetValue is List ? resetValue[i] : resetValue;
+      resetValues?[stageI] = resetValueList![i];
```
desmonddak, line 90: How do we know this nullable is safe here?

lib/src/resettable_entries.dart
```diff
@@ -0,0 +1,95 @@
...
+  ///
+  /// Accepted types include:
+  /// - [Logic]: all reset values are the same, based on that signal.
+  /// - Other [LogicValue.of]-compatible types: all reset values are the same,
+  ///   based on that static value.
+  /// - [List]: each entry can have a different reset value, corresponding to
+  ///   the index in the list.
+  /// - [Map<int, dynamic>]: each entry can have a different reset value,
+  ///   specified by the key-value (key is index) pairs in the map. Unspecified
+  ///   entries will get the default of `0`.
```
desmonddak, line 28: Why not 'x'?

test/register_file_test.dart
```diff
@@ -0,0 +1,49 @@
+// Copyright (C) 2023-2025 Intel Corporation
```
desmonddak, line 1: WHy 2023?

lib/src/shift_register.dart
mkorbel1: since the `?` on `resetValues` will short-circuit the `=`, and there's no case where `resetValues` is non-null but `resetValueList` is `null`.

lib/src/resettable_entries.dart
mkorbel1: After a `reset`, I think the typical expectation is a known state.  E.g. a shift-register full of 0's seems like a nice thing to be able to rely on.

test/register_file_test.dart
mkorbel1: Because I copied these tests from another file for reorganization, they were actually written in ~2023 I think.

Issue #256, opened 2025-09-11
Update `RegisterFile` to use a `LogicArray` internally


mkorbel1:
### Motivation

Currently it just has a bunch of unique storage signals, which is very verbose when generated into SV.  Using a `LogicArray` will make cleaner SV and potentially help synthesis tools.

### Desired solution

Modify `RegisterFile` to use `LogicArray` for internal storage.

PR #259, opened 2025-09-19
new components: cam and replacement

desmonddak:
## Description & Motivation

These are two components needed to build various kinds of cache memory.
First is a CAM which is a contents-addressable memory.  It addresses memory by using a tag instead of an address.
It primarily returns a 'hit' but also the address of where it is found (we can work to make this optional).
The CAM also provides a standard write interface for populating the tags as data, using the line address.

Second is a Replacement module which is abstract and one kind of policy implemented; pseudoLRU.  This will eventually be an optional Module passed into a Cache that can be used to modify a Cache with different way replacement policies.

## Related Issue(s)

None.

## Testing

Basic testing of the CAM to allow for simultaneous tag writes then tag lookups.

## Backwards-compatibility

No.

## Documentation

No.


lib/src/memory/replacement_policy.dart
```diff
@@ -0,0 +1,270 @@
+// Copyright (C) 2024-2025 Intel Corporation
```
mkorbel1: nit: copyright years should start in 2025 presumably for new files?

lib/src/memory/replacement_policy.dart
```diff
@@ -0,0 +1,270 @@
...
+// Author: Desmond Kirkpatrick <desmond.a.kirkpatrick@intel.com>
+
+import 'dart:math';
+
+import 'package:meta/meta.dart';
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+
+/// An interface to a replacement policy that tracks [way] accesses, either
+/// a hita hit on a [way] or
```
mkorbel1: typo: "a hita hit"

lib/src/memory/replacement_policy.dart
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    _buildLogic();
+  }
+  @mustBeOverridden
+  void _buildLogic();
+}
+
+/// A tree-based pseudo-LRU replacement policy.
+class PseudoLRUReplacement extends ReplacementPolicy {
```
mkorbel1: organize file(s) so that we aim for 1 class per file?  if someone adds a new replacement policy in the future, it would be nice for them to feel like adding a new file is appropriate instead of appending to this file.

lib/src/memory/replacement_policy.dart
```diff
@@ -0,0 +1,270 @@
...
+                v.getRange(mid + 1).named(
+                    'miss_${v.name}_${v.width - 1}_${mid + 1}',
+                    naming: Naming.mergeable),
+                base: mid + 1 + base,
+                sz: lsz));
+  }
+
+  /// Access a given way and mark the LRU path in the tree with 0s.
+  ///
+  // At an node, 0 means LRU is right.
```
mkorbel1: typo: missing a `/` in `///`

lib/src/memory/replacement_policy.dart
```diff
@@ -0,0 +1,270 @@
...
+      super.reserveName,
+      super.reserveDefinitionName,
+      String? definitionName})
+      : super(
+            definitionName: definitionName ??
+                'psuedo_lru_replacement_'
+                    'H${hits.length}_A${allocs.length}_WAYS=$ways');
+
+  /// Declare a miss and ask for the least-recently-used way.
+  Logic allocPLRU(Logic v, {int base = 0, int sz = 0}) {
```
mkorbel1: these internal functions should be private (or at least protected), right?

lib/src/memory/replacement_policy.dart
```diff
@@ -0,0 +1,270 @@
...
+  }
+
+  @override
+  void _buildLogic() {
+    // Storage for pseudo-LRU bits that represent the tree.
+    final treePLRUIn =
+        Logic(name: 'plru_in', naming: Naming.mergeable, width: ways - 1);
+    final treePLRU =
+        Logic(name: 'plru', naming: Naming.mergeable, width: ways - 1);
+    Sequential(clk, [
```
mkorbel1: this could be replaced easily with just a `treePLRU <= flop(clk, reset:reset, treePLRUIn)` i think

test/memory/cache_test.dart
```diff
@@ -0,0 +1,388 @@
...
+      await clk.nextPosedge;
+      // write data to address addr
+      const first = 0x20;
+      wrPort.en.inject(1);
+      wrPort.valid.inject(1);
+      wrPort.addr.inject(first);
+      wrPort.data.inject(9);
+      await clk.nextPosedge;
+      wrPort.en.inject(0);
+      await clk.nextPosedge;
```
mkorbel1: you can use `clk.waitCycles` for multiple waits

test/memory/cache_test.dart
```diff
@@ -0,0 +1,388 @@
...
+      await clk.nextPosedge;
+      await clk.nextPosedge;
+      const second = 0x40;
+      wrPort.addr.inject(second);
```
mkorbel1: maybe we should build a little ROHD-VF BFM to send transations instead of doing per-port driving, but not necessarily needed in this PR.  this could be a pathway to a more general testbench, sending random stimulus into the cache and automatically checking that the expected data is returned.

lib/src/memory/cache.dart
```diff
@@ -0,0 +1,487 @@
...
+      this.replacement = PseudoLRUReplacement.new,
+      super.name = 'Cache',
+      super.reserveName,
+      super.reserveDefinitionName,
+      String? definitionName})
+      : dataWidth = (writes.isNotEmpty)
+            ? writes[0].dataWidth
+            : (reads.isNotEmpty)
+                ? reads[0].dataWidth
+                : 0,
```
mkorbel1: is dataWidth=0 legal/supported/tested?

lib/src/memory/cache.dart
```diff
@@ -0,0 +1,487 @@
...
+        reads, tagWidth, lineAddrWidth,
+        prefix: 'match_rd');
+    final validTagRFAlloc = _genValidTagRFInterfaces(
+        writes, tagWidth, lineAddrWidth,
+        prefix: 'alloc');
+
+    // The Tag `RegisterFile`.
+    for (var way = 0; way < ways; way++) {
+      RegisterFile(clk, reset, validTagRFAlloc[way],
+          validTagRFMatchWr[way]..addAll(validTagRFMatchRd[way]),
```
mkorbel1: this line is a little confusing -- why are you adding all the matchrd interfaces to the matchwr interfaces, and why are you doing it during construction of the RF?  it's unclear to me why, but if thats what you want to do, it would make more sense to add tem above before constructing the RF so its a clear thing you're doing instead of a side effect of creating the RF?

lib/src/memory/cache/cache.dart
```diff
@@ -0,0 +1,487 @@
+// Copyright (C) 2024-2025 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// cache.dart
```
mkorbel1, line 4: reminder: writing a nice guide in `doc/` would be good

lib/src/memory/cache.dart
```diff
@@ -0,0 +1,487 @@
...
+        policyAllocPorts[line][wrPortIdx].access <=
+            wrPort.en &
+                wrPort.valid &
+                writeValidPortMiss[wrPortIdx] &
+                getLine(wrPort.addr).eq(Const(line, width: lineAddrWidth));
+      }
+    }
+
+    // Process allocates (misses) and invalidates.
+    for (var wrPortIdx = 0; wrPortIdx < numWrites; wrPortIdx++) {
```
mkorbel1: I see a lot of repeated looping across `wrPortIdx`.  Is there an opportunity to refactor code so that there's classes that aggregate a large amount of "per-wrPortIdx" logic within them rather than tracking multiple independent lists indexed the same way?

lib/src/memory/cache.dart
```diff
@@ -0,0 +1,487 @@
...
+                  for (var way = 0; way < ways; way++)
+                    If.block([
+                      Iff(
+                          // Write with allocate.
+                          wrPort.valid &
+                              writeValidPortMiss[wrPortIdx] &
+                              Const(way, width: log2Ceil(ways))
+                                  .eq(policyAllocPorts[line][wrPortIdx].way),
+                          [
+                            validTagRFAlloc[way][wrPortIdx].en < wrPort.en,
```
mkorbel1: the contents of these if and elseif are the same?  can we make a function or something to reduce copy/paste?

lib/src/memory/cam.dart
```diff
@@ -0,0 +1,142 @@
...
+// Author: Desmond Kirkpatrick <desmond.a.kirkpatrick@intel.com>
+
+import 'package:collection/collection.dart';
+import 'package:meta/meta.dart';
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+
+/// An interface to a Content-Addressable Memory (CAM) that allows querying
+/// for tags and returns the index of the matching tag if found.
+class TagInterface extends Interface<DataPortGroup> {
```
mkorbel1, line 17: Could this be a `PairInterface` instead or are there both read and write cases with this interface?

lib/src/memory/cam.dart
```diff
@@ -0,0 +1,142 @@
...
+  /// The width of data in the memory.
+  final int tagWidth;
+
+  /// The width of addresses in the memory.
+  final int idWidth;
+
+  /// The "enable" bit for this interface, enabling a request.
+  Logic get en => port('en');
```
mkorbel1: why do we need an enable bit?

lib/src/memory/cam.dart
```diff
@@ -0,0 +1,142 @@
...
+
+  void _buildLogic() {
+    // create local storage bank
+    _storageBank = List<Logic>.generate(
+        numEntries, (i) => Logic(name: 'tag_$i', width: dataWidth));
+
+    // A Cam lookup returns the index.
+    for (final lookupPort in lookupPorts) {
+      Combinational([
+        lookupPort.hit < Const(0),
```
mkorbel1: i think these initializers are unnecessary since you already define a default in the Case

lib/src/memory/cam.dart
```diff
@@ -0,0 +1,142 @@
...
+    }
+    _buildLogic();
+  }
+
+  /// Flop-based storage of all memory.
+  late final List<Logic> _storageBank;
+
+  void _buildLogic() {
+    // create local storage bank
+    _storageBank = List<Logic>.generate(
```
mkorbel1, line 132: we could make this a `LogicArray` instead (planned on fixing the RF also https://github.com/intel/rohd-hcl/issues/256, https://github.com/intel/rohd-hcl/pull/258)

lib/rohd_hcl.dart
```diff
@@ -12,20 +12,22 @@ export 'src/encodings/encodings.dart';
 export 'src/error_checking/error_checking.dart';
 export 'src/exceptions.dart';
 export 'src/extrema.dart';
-export 'src/fifo.dart';
 export 'src/find.dart';
 export 'src/find_pattern.dart';
 export 'src/gaskets/spi/spi_gaskets.dart';
 export 'src/interfaces/interfaces.dart';
+export 'src/memory/cam.dart';
```
mkorbel1: you could make a `src/memory/memory.dart` that exports all the memory-related stuff

lib/src/memory/cache.dart
```diff
@@ -0,0 +1,487 @@
...
+/// primary operations:
+/// - Reading from a cache can result in a hit or miss. The only state change is
+///   that on a hit, the [replacement] policy is updated.  This is similar to a
+///   memory, except that a valid bit is returned with the read data.
+/// - Writing to a cache with a valid bit set results in a write into the cache
+///   memory, potentially allocating a line in a new way if the data was not
+///   present. Externally, this just looks like a memory write.
+/// - Writing to a cache without the valid bit set results in an invalidate of
+///   the matching line if present.
+abstract class Cache extends Module {
```
mkorbel1: how can you tell when something is getting evicted, what got evicted?

lib/src/memory/replacement_policy.dart
desmonddak: I am proposing:
```
  @visibleForTesting
  @protected
  ```
  That way I can write combinational tests for these core functions without having to go through the module interface.

File: lib/src/memory/cam.dart (hunk 15 above)
Comment (desmonddak) [reply lib/src/memory/cam.dart]: To tell it to do the cam match.  You are providing a `tag` in, and the `en` tells it to do a match and return the `idx` or data back.  This is the same as a `DataPortInterface` providing an `en` and `addr`.  Just replace those with `en` and `tag`.
You may not want it doing a match if you simply have the `tag` field set on the read port.

File: lib/src/memory/cam.dart (hunk 17 above)
Comment (desmonddak) [reply lib/src/memory/cam.dart:132]: I was waiting on that PR to see how to do it.  And I was debating also about `LogicArray` versus `LogicType` for the RF so we could store structured data....

File: lib/src/memory/cache.dart (hunk 19 above)
Comment (desmonddak) [reply lib/src/memory/cache.dart]: Today it is opaque.  The `Cache` uses the `ReplacementPolicy` to decide and then it writes to that location.
I have to think about write-back caches:  you want to know that data is dirty and being evicted, so you need to know in advance to read it and push it out.  It may be possible to have external access to the `ReplacementPolicy` to know something is about to be evicted (it is multi-ported).

File: lib/src/memory/cam.dart (hunk 14 above)
Comment (desmonddak) [reply lib/src/memory/cam.dart:17]: It has read and write directionalities.  Read sends tag received data.  Write sends tag and data.

File: lib/src/memory/cache.dart (hunk 10 above)
Comment (desmonddak) [reply lib/src/memory/cache.dart]: These are `match` interfaces.  We can match either read or write.  So think of this as just the 'tag' interface.
The other interface is the 'miss' interface called alloc.  So read and write ports both get tag access interfaces to the Tag RF.  Only write gets an alloc interface.

I kept playing with names and settled on `validTagRFMatchWr` and `validTagRFMatchRd` to indicate these are both matches, not read/write interfaces.  Painful to come up with how to make them parallel.  I'll do it explicitly outside.

File: lib/src/memory/cache.dart (hunk 13 above)
Comment (desmonddak) [reply lib/src/memory/cache.dart]: They aren't the same.  One writes to the Alloc side (a write miss) and the other is doing an invalidate.  I struggled to find a way to simplify this combinational.  It's truly tough to do it with just pure signal assignment without a fairly deep muxing structure.  But I admit this block is hard to read.

File: lib/src/memory/cache.dart (hunk 9 above)
Comment (desmonddak) [reply lib/src/memory/cache.dart]: Good catch. I also need to make sure certain numbers of ports are enforced (at least one read/write), and widths match.

File: lib/src/memory/replacement_policy.dart (hunk 5 above)
Comment (desmonddak) [reply lib/src/memory/replacement_policy.dart]: I just checked in an example of this.  I have several more.  It is far easier to exhaustively test the replacement scheme directly than through the ports.

File: lib/src/memory/cache.dart [hunk 20]
```diff
@@ -0,0 +1,483 @@
...
+        reads, tagWidth, lineAddrWidth,
+        prefix: 'match_rd');
+    final validTagRFAlloc = _genValidTagRFInterfaces(
+        writes, tagWidth, lineAddrWidth,
+        prefix: 'alloc');
+
+    // The Tag `RegisterFile`.
+    for (var way = 0; way < ways; way++) {
+      // Combine the read and write match ports for this way.
+      final validTagRFMatch = validTagRFMatchWr[way]
```
mkorbel1: this is actually modifying the `validTagRFMatchWr`, which from your other comment sounds like it's not your intent?  maybe you want this?
```dart
[
    ...validTagRFMatchWr[way],
    ...validTagRFMatchRd[way]
]
```

lib/src/memory/cache.dart
mkorbel1: oh i see yes they are not the same, sorry

lib/src/memory/cam.dart
mkorbel1: but i only see it used in one way, and the write ports are `DataPortInterface`s.  Is there another place you're planning to use them?

lib/src/memory/cam.dart
mkorbel1: the enable is needed for a write, but if you're doing a read isn't it the same if we just always read whatever address is given?  what do you mean "doing a match"?  is this for power savings?  the combo logic is there either way

lib/src/memory/cam.dart
mkorbel1: unresolving since i still see the initializers in the current state of the branch?

lib/src/memory/cam.dart
mkorbel1: that's fine we can come back to it

lib/src/memory/cache.dart
mkorbel1: yeah somehow it seems important to know that information right?  maybe we can omit the feature in this first version, but its also worth thinking about whether it's an easy API to add later (without breaking backwards compatibility)

desmonddak: I thought about this more.  I think we can add an eviction interface (same as Read) which is #reads + #writes wide.
Upon a miss, we can fill that addr/data with the evicted data.  Upon a write/miss, we can setup the write to that location to occur concurrently with the eviction read.  for 1R/1W cache, this would mean 2R/1W memories to support.

desmonddak: Perhaps I can add a nullable or empty eviction list.  `List<ValidDataPortInterface>` which is either passed in as empty or with length=#reads+#writes. I think that would work to support a write-back style cache.

lib/src/memory/cam.dart
desmonddak: Good point and simpler is better.

lib/src/arithmetic/floating_point/floating_point_adder.dart
```diff
@@ -85,23 +84,23 @@ abstract class FloatingPointAdder<FpTypeIn extends FloatingPoint,
     this.clk = (clk != null) ? addInput('clk', clk) : null;
     this.reset = (reset != null) ? addInput('reset', reset) : null;
     this.enable = (enable != null) ? addInput('enable', enable) : null;
-    this.a = (a.clone(name: 'a') as FpTypeIn)
-      ..gets(addInput('a', a, width: a.width));
-    this.b = (b.clone(name: 'b') as FpTypeIn)
-      ..gets(addInput('b', b, width: b.width));
+    this.a = (a.clone(name: 'a') as FpTypeIn)..gets(addTypedInput('a', a));
```
mkorbel1: you dont need these `clone`s anymore

doc/Module.md
```diff
@@ -0,0 +1,156 @@
...
+        await clk.waitCycles(3);
+        expect(output.value, equals(Const(6, width: output.width)));
+
+      await Simulator.endSimulation();
+    });
+  });
+}
+```
+
+Prefer using `waitCycles` instead of `nextPosEdge` and use `inject` instead of
```
mkorbel1: typo: `nextPosedge`

doc/Module.md
```diff
@@ -0,0 +1,156 @@
...
+`put` when working with sequential tests.
+
+When testing a combinational path, and you `inject` inputs after a positive
+clock edge, use `nextNegEdge` to look at the value mid-way through the clock
```
mkorbel1: better: use `previousValue` after waiting for a clock edge

lib/src/arithmetic/signals/fixed_point_logic.dart
```diff
@@ -64,7 +64,10 @@ class FixedPoint extends LogicStructure {
   /// Clone for I/O ports.
   @override
   FixedPoint clone({String? name}) => FixedPoint(
-      signed: signed, integerWidth: integerWidth, fractionWidth: fractionWidth);
+      signed: signed,
+      integerWidth: integerWidth,
+      fractionWidth: fractionWidth,
+      name: name);
```
mkorbel1: probably want `name ?? this.name`

lib/src/interfaces/ready_valid_interface.dart
```diff
@@ -0,0 +1,83 @@
...
+  /// Use the underlying data's name for a stable port name instead of
+  /// interpolating the object itself.
+  late final Logic accepted = (ready & valid).named('${data.name}_accepted');
+
+  /// Creates a [ReadyValidInterface] with the given [data].
+  ReadyValidInterface(this.data)
+      : super(
+          portsFromProvider: [data, Logic.port('valid')],
+          portsFromConsumer: [Logic.port('ready')],
+          // connectFieldsAsPorts: true,
```
mkorbel1: rm old comment

lib/src/interfaces/ready_valid_interface.dart
```diff
@@ -0,0 +1,83 @@
+import 'package:rohd/rohd.dart';
```
mkorbel1, line 1: missing header

lib/src/memory/cache/cached_request_response_channel.dart
```diff
@@ -0,0 +1,327 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// cached_request_response_channel.dart
+// Cached request/response channel with address-based caching.
+//
+// 2025 October 26
+// Authors: Desmond Kirkpatrick <desmond.a.kirkpatrick@intel.com>
+//         GitHub Copilot <github-copilot@github.com>
```
mkorbel1: dont list copilot as an author, you're the author using copilot

lib/src/memory/csr/csr.dart
```diff
@@ -103,23 +103,24 @@ class Csr extends LogicStructure {
     required this.config,
     required this.rsvdIndices,
     required List<Logic> fields,
-  }) : super(fields, name: config.name);
-
+    String? logicName,
+  }) : super(fields, name: logicName ?? config.name);
+
+  // Historically the CSR disallowed renaming because the architectural
```
mkorbel1, line 109: @kimmeljo any concerns?

test/memory/cache/cached_request_response_channel_test.dart
mkorbel1, line 1: these test files are getting pretty long.  is there an opportunity to refactor a bit to reduce repetition by making some reusable infra?

test/memory/cache_test.dart
desmonddak: I was thinking we could extend the idea of the ready-valid BFM to be the test equivalents of loop-back when we test transmitter/receivers:   A specific one for ready_and_valid, ready_then_valid, and valid-then-ready, to be at the front and back of a ready/valid component.  The `ReadyValidFifo` could be an example test-case.  I haven't used ROHD-VF but it seems I should learn that first and make sure this extends that.  But the big issue is avoiding combinational loops as the ready-line is often fully combinational through these and you have to use a skid-buffer (1-deep fifo) to break them.  As I mentioned before, detecting combinational loops statically (not exactly fv) is an interesting capability to help designers not have inexpicable X's when they compose components.

test/memory/cache/cached_request_response_channel_test.dart
desmonddak: The last checkin https://github.com/intel/rohd-hcl/pull/259/commits/104e6aa8cf548472857d27cb3af2e6d9be64b020

you will see a major refactoring of the cache and channel testing.

test/memory/cache_test.dart
mkorbel1: Theres no loops in the BFMs so the same one should work for all of them

lib/src/interfaces/request_response_channel_base.dart
kimmeljo, line 1: What's the difference between this file and lib/src/interfaces/request_response_channel.dart?

lib/src/memory/cache/cached_request_response_channel.dart
kimmeljo, line 1: Two suggestions here for the future (both of which we've discussed):

(1) It might be useful to create a "flavor" of this module that assumes strict ordering of responses relative to requests. Then, instead of a CAM (fully associative cache) to track ordering, we can simplify the HW a bit by using a FIFO. This has the added benefit at the API level of not requiring a unique ID on incoming requests.

(2) It might be useful to create a "flavor" of this module that has a register file instead of a full blown cache on the data side. This would more closely mirror the notion of hash table lookup in HW. The API might have to be different though because it would require explicit invalidation/eviction from the outside.

lib/src/memory/cache/direct_mapped_cache.dart
kimmeljo, line 1: More of a general question, but does the RegisterFile abstraction allow for use of things like SRAM? I can imagine in some cases we'll need an SRAM at least for the data portion. Maybe we just punt on this for now?

lib/src/memory/cache/set_associative_cache.dart
kimmeljo, line 1: More of a nitpick, but isn't set associative the general term while fully associative/direct mapped are just special cases (ways=n, 1 respectively). Maybe worth consolidating into a single implementation?

lib/src/memory/csr/csr.dart
kimmeljo: I can't recall if the instance config can override the name from the architectural config of a CSR. If it can, then I don't see why this change would be necessary as the logic name == instance name.

@desmonddak was there a specific reason you needed this?

lib/src/memory/ready_valid_fifo.dart
kimmeljo, line 1: This is awesome. I vaguely recall in the "regular" FIFO implementation that writeEnable when full and readEnable when empty can cause bad things to happen. I see you've handled that here. Is it worth handling that in the "regular" implementation too? Or am I misremembering?

test/interfaces/apb_bfm_test.dart
kimmeljo, line 1: Note: I am moving these tests in my axi branch to a different location. So maybe undo this change and I will deal with it on my side? Unless you really need it...

lib/src/interfaces/request_response_channel_base.dart
desmonddak: Good catch.  This file is supposed to be the structures and the abstract base class (API).
The other file is supposed to be just a wired passthrough implementation.  I accidentally duplicated in the second file.

lib/src/memory/cache/direct_mapped_cache.dart
desmonddak: This is an important question.  Right now our RegisterFile abstraction supports asynchronous reads and I don't believe SRAM timing will allow that.  So this will impact microarchitecture choices because of the additional cycle.
I think TAG side requires multiple ports because the read and write ports need to do a tag match, so I think you are right: we will likely stay with flops on the TAG side, but we need to think about the data side being an actual SRAM.

lib/src/memory/cache/set_associative_cache.dart
desmonddak: Yes, this is absolutely true.  I can try to consolidate, first at the API level, and then perhaps at the implementation level. ROHD is powerful enough to probably make this simple (like single 'way' array probably eliminates some logic signals).

lib/src/memory/csr/csr.dart
desmonddak: Now I cannot remember why I needed this.  Let me undo and figure out what demanded the name.

lib/src/memory/ready_valid_fifo.dart
desmonddak: I cannot remember if we built the bypasses there as well.  
The reason it could be worth handling is things like skidbuffers (single entry FIFOs) that need to be depth 2 if you don't handle these bypass cases.  There can be a lot of those that have to have a spare entry.  
We need to check this (and make sure there is a clear test for these edge cases to ensure they never break.

test/interfaces/apb_bfm_test.dart
desmonddak: I will undo this. I do not think I need it.

PR #270, opened 2025-11-26
APB Completer HW

kimmeljo:
## Description & Motivation

An implementation of an abstract APB completer along with a concrete implementation that hooks into a CsrBlock for register reads/writes. 

## Related Issue(s)

N/A

## Testing

Unit tests added for the concrete implementation.

## Backwards-compatibility

Fully backwards compatible.

## Documentation

Added to APB documentation.

### DIFF
--- a/lib/src/models/amba/apb_bfm/apb_compliance_checker.dart
+++ b/lib/src/models/amba/apb_bfm/apb_compliance_checker.dart
@@ -102,14 +102,14 @@ class ApbComplianceChecker extends Component {
           }
 
           // collect "last" items for next check
-          lastWrite = intf.write.previousValue!;
-          lastAddr = intf.addr.previousValue!;
+          lastWrite = intf.write.previousValue;
+          lastAddr = intf.addr.previousValue;
           lastSel = currSels;
-          lastWriteData = intf.wData.previousValue!;
-          lastStrb = intf.strb.previousValue!;
-          lastProt = intf.prot.previousValue!;
-          lastAuser = intf.aUser?.previousValue!;
-          lastWuser = intf.wUser?.previousValue!;
+          lastWriteData = intf.wData.previousValue;
+          lastStrb = intf.strb.previousValue;
+          lastProt = intf.prot.previousValue;
+          lastAuser = intf.aUser?.previousValue;
+          lastWuser = intf.wUser?.previousValue;
         }
       }
 


mkorbel1:
Looks like this PR includes changes from #251, so we'll either merge that one first or you can separate it out

lib/src/gaskets/amba5/apb_completer.dart
desmonddak, line 1: Needs a Copyright header

lib/src/gaskets/amba5/amba5_gaskets.dart
```diff
@@ -0,0 +1,4 @@
+// Copyright (C) 2024-2025 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+
```
desmonddak, line 3: Our pattern has been to include a date line and an author line in the header.

doc/components/amba/apb.md
```diff
@@ -1,4 +1,14 @@
-# APB BFM
+# APB
```
desmonddak, line 1: Since I am quite unfamiliar with Arm Bus protocols, I don't know if some of these belong in the VF versus HCL libraries.
Maybe some description of what is here and what is there could help.
Also, some hierarchical description of the componentry  (like a small list with indentations) could help the user grok all the componentry needed on the component side versus the VF side without going into reading the spec)  I don't think the list is too long, maybe a dozen or so lines.

kimmeljo: In general my understanding is that ROHD-VF is meant for foundational verification components (i.e., the base Monitor, Driver, etc. classes). I.e., we should not have anything implementation/use case specific in there - like an AMBA component. Unless I'm misunderstanding the concern?

lib/src/gaskets/amba5/amba5_gaskets.dart
kimmeljo: I'm looking at other files that only do exports and we don't seem to have an author line in those....

kimmeljo: Ex:
lib/rohd_hcl.dart

doc/components/amba/apb.md
kimmeljo: To the second comment, the list of components is pretty clearly (at least in my opinion) laid out in the APB BFM section. Do you not agree?

lib/src/gaskets/amba5/amba5_gaskets.dart
mkorbel1: Yeah that sounds right to me, since there's not much authorship to do in these kinds of files

doc/components/amba/apb.md
mkorbel1: > my understanding is that ROHD-VF is meant for foundational verification components (i.e., the base Monitor, Driver, etc. classes). I.e., we should not have anything implementation/use case specific in there - like an AMBA component

yes that's right

lib/src/gaskets/amba5/apb_completer.dart
```diff
@@ -0,0 +1,199 @@
...
+
+  /// Indicator that data from APB can be consumed downstream.
+  /// This can be used as an input for the consuming logic.
+  @protected
+  late final Logic downstreamValid;
+
+  /// Indicator that data from downstream is consumable on APB
+  /// This must be properly driven in any child class.
+  @protected
+  late final Logic upstreamValid;
```
mkorbel1, line 43: should this be just `Logic get upstreamValid` or something to force child-classes to implement it?

lib/src/gaskets/amba5/apb_completer.dart
```diff
@@ -0,0 +1,199 @@
...
+        ],
+      ),
+    ]);
+
+    _build();
+  }
+
+  /// User hook to deal with downstream.
+  /// Must be implemented by the user.
+  void buildCustomLogic();
```
mkorbel1: this function seems kind of weird to me since nothing is calling it in the base class.  isn't it just implicit to the user that they must implement some logic somewhere?  i'm not sure what this abstract function adds

lib/src/gaskets/amba5/apb_completer.dart
```diff
@@ -0,0 +1,199 @@
...
+                    width: fsm.currentState.width)) |
+                fsm.currentState.eq(Const(ApbCompleterState.access.index,
+                    width: fsm.currentState.width))) &
+            upstreamValid;
+  }
+}
+
+/// APB Completer that is meant to be used against CSRs
+/// as defined in ROHD-HCL.
+class ApbCsrCompleter extends ApbCompleter {
```
mkorbel1: separate file?

lib/src/gaskets/amba5/apb_completer.dart
```diff
@@ -0,0 +1,199 @@
...
+  /// How many APB clock cycles before we should indicate
+  /// data is complete.
+  late final int apbClkLatency;
```
mkorbel1: nit: kind of a weird variable name, why not just `responseLatency` or something?  It's always APB and always counted in clocks, right?

lib/src/models/amba/apb_bfm/apb_compliance_checker.dart
```diff
@@ -45,7 +45,7 @@ class ApbComplianceChecker extends Component {
     LogicValue? lastAuser;
     LogicValue? lastWuser;
 
-    intf.clk.posedge.listen((event) {
+    intf.clk.negedge.listen((event) {
```
mkorbel1: why negedge?  is that what the spec says or is this maybe a band-aid over things that should have been `previousValue`?

test/apb_test.dart
```diff
@@ -7,21 +7,26 @@
...
+import 'dart:io';
+
 import 'package:rohd/rohd.dart';
 import 'package:rohd_hcl/rohd_hcl.dart';
+import 'package:rohd_vf/rohd_vf.dart';
 import 'package:test/test.dart';
 
-class ApbCompleter extends Module {
-  ApbCompleter(ApbInterface intf) {
+class ApbCompleterTest extends Module {
```
mkorbel1: nit: a `Test` is a ROHD-VF class, maybe call it `ApbCompleterTestModule` or something?

kimmeljo: Yeah good point!

lib/src/models/amba/apb_bfm/apb_compliance_checker.dart
kimmeljo: You're right - it certainly was! Refactored to use `previousValue` everywhere

lib/src/gaskets/amba5/apb_completer.dart
kimmeljo: Yeah, fair point. Especially with the CSR completer as an example I agree that users can probably figure out what they need to do without explicit parent class constraints.

lib/src/gaskets/amba5/apb_completer.dart
kimmeljo: I'm trying to think how best to do it. I want to always initialize it to be a `Logic` within the module, but the child class must be responsible for driving that `Logic`. Not sure if we can enforce that even with a getter?

PR #271, opened 2025-11-26
Up/Down Counter

kimmeljo:
## Description & Motivation

A convenience a constructor of Counter that increments and decrements by 1. This is useful for several more complex flows such as crediting.

## Related Issue(s)

N/A

## Testing

Added a unit test for this new counter constructor.

## Backwards-compatibility

Fully backwards compatible.

## Documentation

Added to summation documentation.

### DIFF
--- a/lib/src/summation/counter.dart
+++ b/lib/src/summation/counter.dart
@@ -1,4 +1,4 @@
-// Copyright (C) 2024 Intel Corporation
+// Copyright (C) 2024-2025 Intel Corporation
 // SPDX-License-Identifier: BSD-3-Clause
 //
 // counter.dart
@@ -166,7 +166,7 @@ class Counter extends SummationBase {
   /// to count [by] along with much of the
   /// other available configuration in the default [Counter] constructor.
   /// And allows for both incrementing and decrementing the count.
-  Counter.updn({
+  Counter.upDown({
     required Logic clk,
     required Logic reset,
     required Logic enableInc,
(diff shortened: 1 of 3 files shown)


lib/src/summation/counter.dart
```diff
@@ -162,6 +162,45 @@ class Counter extends SummationBase {
             saturates: saturates,
             name: name);
 
+  /// A simplified constructor for [Counter] that accepts a single fixed amount
+  /// to count [by] along with much of the
+  /// other available configuration in the default [Counter] constructor.
+  /// And allows for both incrementing and decrementing the count.
+  Counter.updn({
```
mkorbel1: Dart conventions suggest we should call this `Counter.upDown`
https://dart.dev/effective-dart/style#do-name-other-identifiers-using-lowercamelcase

It's a little longer, but might be worth the consistency benefits.  Thoughts?

lib/src/summation/counter.dart
```diff
@@ -162,6 +162,45 @@ class Counter extends SummationBase {
...
+    bool saturates = false,
+    bool asyncReset = false,
+    int resetValue = 0,
+    String name = 'counter',
+  }) : this([
+          SumInterface(
+            width: width,
+            fixedAmount: by,
+            hasEnable: true,
+          )..enable?.gets(enableInc),
```
mkorbel1: since these `hasEnable=true`, they should be `..enable!` instead of `?`

lib/src/summation/counter.dart
mkorbel1, line 1: update copyright year on files modified (i think the test also)

mkorbel1 (changes requested):
Great addition, thank you

lib/src/summation/counter.dart
kimmeljo: I'm good with it!

PR #273, opened 2025-12-13
Add setData testbench hook for register file

RossComputerGuy:
## Description & Motivation

Adds an easy testbench function for initializing data in the register file.

## Backwards-compatibility

## Documentation


mkorbel1 (approved):
This looks like a reasonable addition, thanks for the contribution!

Issue #146, opened 2024-12-12
Wrapper for ready/valid indication on fixed-latency hardware


mkorbel1:
### Motivation

If you know some hardware has fixed-latency, it would be nice to be able to wrap it with some automation that signals information about when it is ready on the input and valid on the output.

For example:
- Pipelined: show a valid on the output, optionally maybe an ID that was passed in with the input?
- Multi-cycle: show a ready on the input, valid on the output

### Desired solution

Some easy automation wrapper for multi-cycle ready/valid indications

PR #274, opened 2025-12-14
Add AssertiveDataPortInterface

RossComputerGuy:
## Description & Motivation

Adds a valid output to `DataPortInterface`. This makes it possible to know whether the read or write has completed successfully.

Adds a done output to `DataPortInterface`. This makes it possible to know whether the read or write has completed.

## Related Issue(s)

Seems like #146 is somewhat related since this allows for polling if `valid` is set and knowing that marks things are valid.

## Testing

Ensure `test/memory_model.dart` passes

## Backwards-compatibility

## Documentation

### DIFF
--- a/lib/src/memory/memory.dart
+++ b/lib/src/memory/memory.dart
@@ -17,7 +17,33 @@ enum DataPortGroup {
   control,
 
   /// For data signals to/from memory.
-  data
+  data,
+
+  /// For data signal validation.
+  validation,
+}
+
+/// A [DataPortInterface] that signals the data validation and readiness.
+class AssertiveDataPortInterface extends DataPortInterface {
+  /// On a read port, this indicates that the data is ready to be read.
+  /// On a write port, this indicates that the write has completed.
+  Logic get ready => port('ready');
+
+  /// A signal that indicates the transaction is successful.
+  Logic get valid => port('valid');
+
+  /// Constructs a [DataPortInterface] with ready and valid.
+  AssertiveDataPortInterface(super.dataWidth, super.addrWidth) {
+    setPorts([
+      Logic.port('ready'),
+      Logic.port('valid'),
+    ], [
+      DataPortGroup.validation
+    ]);
+  }
+
+  @override
+  DataPortInterface clone() => AssertiveDataPortInterface(dataWidth, addrWidth);
 }
 
 /// A [DataPortInterface] that supports byte-enabled strobing.
--- a/lib/src/memory/memory.dart
+++ b/lib/src/memory/memory.dart
@@ -208,14 +208,14 @@ abstract class Memory extends Module {
       rdPorts.add(readPorts[i].clone()
         ..connectIO(this, readPorts[i],
             inputTags: {DataPortGroup.control},
-            outputTags: {DataPortGroup.data},
+            outputTags: {DataPortGroup.data, DataPortGroup.validation},
             uniquify: (original) => 'rd_${original}_$i'));
     }
     for (var i = 0; i < numWrites; i++) {
       wrPorts.add(writePorts[i].clone()
         ..connectIO(this, writePorts[i],
             inputTags: {DataPortGroup.control, DataPortGroup.data},
-            outputTags: {},
+            outputTags: {DataPortGroup.validation},
             uniquify: (original) => 'wr_${original}_$i'));
     }
   }
--- a/lib/src/models/memory_model.dart
+++ b/lib/src/models/memory_model.dart
@@ -69,10 +69,17 @@ class MemoryModel extends Memory {
         if (!wrPort.en.previousValue!.isValid && !storage.isEmpty) {
           // storage doesnt have access to `en`, so check ourselves
           storage.invalidWrite();
+
+          if (wrPort is AssertiveDataPortInterface) {
+            wrPort.ready.put(LogicValue.one);
+            wrPort.valid.put(LogicValue.zero);
+          }
           return;
         }
 
-        if (wrPort.en.previousValue == LogicValue.one) {
+        final wrEnHigh = wrPort.en.previousValue == LogicValue.one;
+
+        if (wrEnHigh) {
           final addrValue = wrPort.addr.previousValue!;
 
           if (wrPort is MaskedDataPortInterface) {
@@ -92,6 +99,11 @@ class MemoryModel extends Memory {
             storage.writeData(addrValue, wrPort.data.previousValue!);
           }
         }
+
+        if (wrPort is AssertiveDataPortInterface) {
+          wrPort.ready.put(LogicValue.one);
+          wrPort.valid.put(wrEnHigh ? LogicValue.one : LogicValue.zero);
+        }
       }
 
       for (final rdPort in rdPorts) {
@@ -101,7 +113,8 @@ class MemoryModel extends Memory {
               !rdPort.en.previousValue!.toBool() ||
               !rdPort.addr.previousValue!.isValid) {
             unawaited(_updateRead(
-                rdPort, LogicValue.filled(rdPort.dataWidth, LogicValue.x)));
+                rdPort, LogicValue.filled(rdPort.dataWidth, LogicValue.x),
+                valid: false));
           } else {
             unawaited(_updateRead(
                 rdPort, storage.readData(rdPort.addr.previousValue!)));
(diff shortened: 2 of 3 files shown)


lib/src/models/memory_model.dart
```diff
@@ -69,10 +69,12 @@ class MemoryModel extends Memory {
         if (!wrPort.en.previousValue!.isValid && !storage.isEmpty) {
           // storage doesnt have access to `en`, so check ourselves
           storage.invalidWrite();
+          wrPort.valid.put(0);
+          wrPort.done.put(1);
           return;
         }
 
-        if (wrPort.en.previousValue == LogicValue.one) {
+        if (wrPort.en.value == LogicValue.one) {
```
mkorbel1: this change looks like it introduces a bug, since it should be sampling the value pre-clk edge

lib/src/memory/memory.dart
```diff
@@ -62,6 +65,12 @@ class DataPortInterface extends Interface<DataPortGroup> {
   /// The data sent or received with the associated request.
   Logic get data => port('data');
 
+  /// Whether the data was successfully read or written
```
mkorbel1: It looks like these two signals are valid at the same time (latency) as `data`, is that right?  It would probably be worth indicating that in the documentation

lib/src/memory/memory.dart
```diff
@@ -77,6 +86,13 @@ class DataPortInterface extends Interface<DataPortGroup> {
     ], [
       DataPortGroup.data
     ]);
+
+    setPorts([
```
mkorbel1: thought: do we want to make these integrity signals optional/nullable via a flag and hardware components can dynamically choose whether to generated related logic based on their presence?

lib/src/memory/register_file.dart
```diff
@@ -53,44 +61,83 @@ class RegisterFile extends Memory with ResettableEntries {
...
+        ],
+        orElse: [
+          for (var entry = 0; entry < numEntries; entry++)
+            ...wrPorts.map(
+              (wrPort) =>
+                  // set storage bank if write enable and pointer matches
+                  If(
+                wrPort.en & wrPort.addr.eq(entry),
+                then: [
+                  wrPort.valid < 1,
```
mkorbel1: I wonder if including valid and done as part of the RF would impact synthesis' ability to identify these as RFs?  Also, I wonder if this generates more logic than is necessary (i.e., can we just do an address range check and a shift-register instead of a per-entry case statement, and would that synthesize differently)?

lib/src/memory/memory.dart
```diff
@@ -62,6 +65,12 @@ class DataPortInterface extends Interface<DataPortGroup> {
   /// The data sent or received with the associated request.
   Logic get data => port('data');
 
+  /// Whether the data was successfully read or written
+  Logic get valid => port('valid');
```
mkorbel1: thought: does calling this `valid` overload the terminology related to "ready/valid" protocols in a confusing way?  @desmonddak what do you think?

test/memory_test.dart
mkorbel1, line 1: it would probably be good to add some testing for when valid and done should be `false`

lib/src/memory/csr/csr_block.dart
```diff
@@ -252,11 +255,20 @@ class CsrBlock extends CsrContainer {
       );
     }
 
+    if (frontWritePresent) {
+      final wrHit = wrHits.isEmpty ? Const(0) : wrHits.reduce((a, b) => a | b);
```
mkorbel1: nit: you could use `wrHits.swizzle().or()`

mkorbel1 (commented):
Thank you for the thoughtful contribution!  A few questions and comments

lib/src/memory/memory.dart
RossComputerGuy: Maybe, though would that just be not connecting those up then? If they're not connected, I think ROHD would just not generate those.

lib/src/memory/memory.dart
RossComputerGuy: I think valid is the most sensible name since it lets you know whether the request was able to complete. In terms of reading, that means `data` is valid. In terms of writing, it means the address which was written to is valid.

lib/src/memory/register_file.dart
RossComputerGuy: Afaict, it doesn't impact yosys's ability to generate RF's. Yosys doesn't even recognize RF's as memories. Also, there's a chance that yosys would just optimize it better than how we have it. Though, idk if we even need an address range check. We might be able to get away with just having `wrPort.valid < wrPort.en`.

lib/src/models/memory_model.dart
RossComputerGuy: Reverted it back, I think I left it when I was debugging something.

lib/src/memory/memory.dart
RossComputerGuy: Yes, because done & valid assert that it was successful while done & !valid means it was unsuccessful. So `data` in a read is valid when `done & valid` is true.

lib/src/memory/csr/csr_block.dart
RossComputerGuy: Done

test/memory_test.dart
RossComputerGuy: Been trying to but the sim is being picky.

lib/src/memory/memory.dart
kimmeljo: One other consideration here is backwards compatibility. Making them nullable allows for existing implementations using `DataPortInterface` not to have to do anything should they not want to.

kimmeljo (commented):
Per the discussion on the ports being required versus optional, for backward compatibility reasons, I'd make the case that they should be optional. This can (probably will) simplify the change set for this PR.

desmonddak:
The dilemma we face is that we are going to start using the `DataPortInterface` in many different ways for components and we may need to architect it so that there are derived from a base `Interface` class representing the simplest possible memory accessor.  Right now, we are overloading it with byte output control, for example, whereas the simplest interface probably should not have that.

Furthermore, we are extending it with a valid bit to call it `ValidDataPortInterface`  in https://github.com/intel/rohd-hcl/pull/259/files#diff-1db666ec943a7d409d08eda2bc00935cd71160c8ae7ae14b471fe5458fc5dde7   (see the first 100 lines of the cache.dart file).

In that same PR, we are also extending the `Fifo` to be ready/valid. 

If we add valid/done to the base memory, we may later find it difficult to create a broad array of memory components (CAM/TCAM/RF/FIFO/SRAM/WriteBuffer) with the various configurations they need.  So I want to be careful incrementally extending `DataPortInterface`  as we probably should architect an `Interface` hierarchy of data interfaces that will give us this more general capability to our data interfaces across this spectrum of components (to be built eventually).  Already, I think we should factor out the byte enable somehow and only configure it for components that support byte-enable. 

I think the fork in the road is:  do we add nullable `Logic` and parameters to this `DataPortInterface` or do we inherit from a simpler `Interface` class of address+data and build up the various kinds of interfaces for memory components, as well as a hierarchy of memory components that inherit from `Memory` or `RF` or `SRAM` and have more and more features.

For now, I would propose a nullable interface to keep this PR simple, but I am a bit concerned about the concept of 'done' when we want to build a read/valid protocol set of components and 'done' is introducing new semantics potentially.

desmonddak:
I think in the intervening interval, we merged a PR with significant filesystem reorganization, which means you currently have merge conflicts.  I think you should consider a new interface that inherits (which may make it easier for the future I mentioned, which was to consider a tree of data-oriented interfaces for memory-like devices.  Or you can proceed with a Logic? add to the existing.  The interface is a contract, so if we push too much into the bottom, then there are more exceptions to document for rudimentary components (like a flop-based RF would probably have few features supported).

RossComputerGuy:
Sounds good, sorry for the delay

RossComputerGuy:
I've added `AssertiveDataPortInterface`, currently trying to figure out how to integrate it in with some devices. In my case right now, it really would be useful with SPI devices.

mkorbel1:
This change seems less disruptive to me, sorry for the delay on review.  @desmonddak, would value your inputs on this PR since you're spending more time than I on related topics and implementations.

Regarding signal naming -- what about just matching the naming you're using in the comments?  I.e. when it is complete, `completed`, and when it is successful, `successful`?  That avoids any potential name collisions with ready/valid handshakes.

One note, is it would be good to include some sort of test to protect the implementation, even a very simple one.

RossComputerGuy:
> Regarding signal naming -- what about just matching the naming you're using in the comments? I.e. when it is complete, `completed`, and when it is successful, `successful`? That avoids any potential name collisions with ready/valid handshakes.

I don't see how there'd be name collisions with ready/valid handshakes because this is creating one. When `connectIO` is used, it should rename things in a safe way.

RossComputerGuy:
I've made `MemoryModel` work with `AssertiveDataPortInterface` now. Tests pass for it.

mkorbel1:
Re: name collisions, I meant in terminology and usage rather than programmatically.  I.e. if there are `DataPortInterface`s of various types, with multiple different semantics and meanings for `ready` and `valid` it could be confusing?

PR #275, opened 2025-12-19
DTI Messages and HW

kimmeljo:
## Description & Motivation

This adds `LogicStructure`s for AMBA compliant DTI_TBU messages. This also adds a generic DTI controller module that is capable of sending and receiving DTI messages in a compliant fashion over a pair of AXI-S interfaces.

As part of the controller module, we also add AXI-S compliant link layer controller modules for TX and RX on the physical interface.

## Related Issue(s)

N/A

## Testing

Unit testing add for all HW modules described above.

## Backwards-compatibility

There is one minor backward incompatible change to the AXI-S interface definition. There was a bug where the `TUSER` signal was not taking the correct width if present.

## Documentation

Documentation will be added as part of this change set under doc/components/amba/dti.md.


lib/src/interfaces/ready_valid_interface.dart
```diff
@@ -0,0 +1,83 @@
+import 'package:rohd/rohd.dart';
```
desmonddak: Need Copyright headers.  Copilot does a decent job if you say 'put our usual Copyright header on files without'.

kimmeljo: See my comment below about `ready_valid_interface.dart`. I think it's better for that to come from a different PR but I needed it for my work. Anything else that's missing a copyright header I will add.

lib/src/protocols/amba5/axi_s/axi_s_interface_tx.dart
```diff
@@ -0,0 +1,335 @@
...
+            actions: [],
+          ),
+          // WORKING: move to IDLE when we are done sending.
+          // but only if there isn't another send immediately following
+          State(AxiStreamBeatState.working, events: {
+            acceptNextSend & ~msgToSendValid: AxiStreamBeatState.idle,
+          }, actions: []),
+        ],
+      );
+      sendIdle <= sendState.currentState.eq(AxiStreamBeatState.idle.index);
```
mkorbel1, line 188: suggestion: rather than querying `currentState`, move driving of related signals for different states into the `actions`

lib/src/protocols/amba5/axi_s/axi_s_interface_tx.dart
```diff
@@ -0,0 +1,335 @@
...
+// 2025 December
+// Author: Josh Kimmel <joshua1.kimmel@intel.com>
+
+import 'dart:math';
+
+import 'package:rohd/rohd.dart';
+import 'package:rohd_hcl/rohd_hcl.dart';
+
+/// A generic module to send messages over AXI-S.
+class AxiStreamInterfaceTx extends Module {
```
mkorbel1, line 17: question: why have the word `Interface` in the class name?  why not just `AxiSTx` or something?

lib/src/protocols/amba5/axi_s/axi_s_interface_tx.dart
```diff
@@ -0,0 +1,335 @@
...
+  /// Clock and reset.
+  late final Axi5SystemInterface sys;
+
+  /// Outbound messages.
+  late final Axi5StreamInterface stream;
+
+  /// Is the provided [msgToSend] valid.
+  ///
+  /// Only drive on the interface if so.
+  late final Logic msgToSendValid;
```
mkorbel1, line 30: should these `Logic`s be `@protected`?  Are the supposed to be used/visible outside of the class?

lib/src/protocols/amba5/axi_s/axi_s_interface_tx.dart
```diff
@@ -0,0 +1,335 @@
...
+      // determine how many beats we need
+      final numBeats = (msgToSend.width / stream.dataWidth).ceil();
+      final sendIdle = Logic(name: 'sendIdle');
+
+      // must count as we're sending the message across multiple beats
+      // but restart the count once we're done sending
+      final countEnable = stream.valid & (stream.ready ?? Const(1));
+      final acceptNextSend =
+          sendIdle | (countEnable & (stream.last ?? Const(1)));
+      final beatCounter = Counter.simple(
```
mkorbel1, line 152: without having read the spec in detail, is there an opportunity to use the `Serializer` (etc.) components to reduce complexity/increase reuse?

lib/src/protocols/amba5/axi_s/axi_s_interface_tx.dart
kimmeljo: No particularly strong reasoning here. If you think it's better without, I have no problem changing!

lib/src/protocols/amba5/axi_s/axi_s_interface_tx.dart
kimmeljo: They shouldn't be visible externally so good call!

lib/src/protocols/amba5/axi_s/axi_s_interface_tx.dart
kimmeljo: Ah yes it sure does look like `Serializer`  (and `Deserializer` for rx) would work nicely here - let me try refactoring

kimmeljo: Turns out there is a limitation that would prevent me from using these standard components. I filed an issue here:
https://github.com/intel/rohd-hcl/issues/277

I might just try to crank the issue out myself and then apply it.

Issue #166, opened 2025-02-03
`Serializer` computes `count` width 0 when only 1 element in `deserialized`


mkorbel1:
### Describe the bug

If a single element is passed into a `Serializer` (i.e. `deserialized` is a 1D array with 1 element), then the width of `count` is determined to be 0, which then breaks the assignment to the output of the `Counter` since it always has at least 1 bit.

### To Reproduce

Create a `Serializer` with a 1D array with 1 element in it passed in.

### Expected behavior

The `Serializer` should always have 1-bit output on `count` probably, or at least it should build properly and drive outputs in a way that makes sense.

### Actual behavior

>   Signal count has the wrong width. Expected 0 but found 1. 

### Additional: pubspec.yaml

```yaml

```

joserochh:
Working on this issue!

PR #284, opened 2026-03-20
Minimum width on serializer's counter - Fix for issue 166

joserochh:
## Description & Motivation

Ensuring a minimum width on serializer's counter.

## Related Issue(s)

#166 

## Testing

The test creates a Serializer with a 1-element LogicArray([1], 8) and asserts count.width >= 1 (the invariant fixed by max(1, ...)).
Then verifies the module builds and drives serialized = 42 with done = true on the first active clock cycle.

## Backwards-compatibility

No

## Documentation

No

### DIFF
--- a/lib/src/count.dart
+++ b/lib/src/count.dart
@@ -7,7 +7,6 @@
 // 2023 July 11
 // Author: Rahul Gautham Putcha <rahul.gautham.putcha@intel.com>
 
-import 'dart:math';
 
 import 'package:rohd/rohd.dart';
 import 'package:rohd_hcl/src/utils.dart';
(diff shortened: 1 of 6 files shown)


desmonddak:
Looks great!  Will merge soon.

joserochh:
> Looks great! Will merge soon.

I found other cases where this log2ceil is used, without this fix, or with this very same fix already. So maybe we can have a new function that ensures minimum width, but not sure how to call it.

Maybe widthFor(val), or bitsNeeded like chisel.  Or less creative, log2ceilMin1

mkorbel1 (approved):
Looks great, thank you!

PR #287, opened 2026-04-24
Multi-cycle divider improvements: FSM enum, 1's complement, and quotient-only mode

### LINKED ISSUE #139: Multi-cycle divider improvements
### Motivation

Some of the computations are using 2's complement where a 1's complement (inversion) based approach may be more efficient.
This should be explored to see if you can save all those carry operations.

More corner cases for the divider would be good to have in the test suite.
A test using a narrow divider might cover all corners.
We could enhance the MultCycleDividerState to be an enum and perhaps use the ROHD-HCL FSM construct.

### Desired solution

Create a narrow divider and see if we can cover all corners.
If not, create a list of corner cases and cover them with short tests.

Try converting computations to use 1's complement and avoid the carries.

### Alternatives considered

_No response_

### Additional details

_No response_

joserochh:
## Description & Motivation

This PR refactors and extends the MultiCycleDivider with two key changes:

1. FSM Refactor (divider.dart): Replaces the manual ad-hoc FSM (using _MultiCycleDividerState static constants, raw Combinational/Sequential blocks, and a currentState/nextState pair) with ROHD's FiniteStateMachine<MultiCycleDividerStates>. The state constants are promoted to a proper enum MultiCycleDividerStates. This reduces boilerplate, improves readability, and aligns with idiomatic ROHD patterns.

2. Quotient-only mode (computeRemainder flag): Adds a computeRemainder parameter to MultiCycleDivider. When false, the divider uses a faster O(n) binary long-division algorithm (one quotient bit per clock cycle) instead of the full O(n²) greedy algorithm, and the remainder output is always 0.

3. New OnesComplementDivider (ones_complement_divider.dart): Adds a new sibling divider that uses one's complement signed arithmetic instead of two's complement. It shares the same MultiCycleDividerInterface and MultiCycleDividerStates FSM but replaces the sign-negate operation (~x + 1) with a simpler bitwise invert (~x), reducing the critical path in the convert state. It also supports the computeRemainder flag with both O(n²) and O(n) modes.

## Related Issue(s)

#139

## Testing

* Existing tests for MultiCycleDivider were extended to run in both computeRemainder: true and computeRemainder: false modes (all test groups are now parameterised over both modes).
* New VF test infrastructure added for OnesComplementDivider: dedicated OnesComplementInputMonitor, OnesComplementScoreboard, OnesComplementAgent, OnesComplementEnv, and OnesComplementTest classes.
* New test sequences added: OnesCompBasicSequence, OnesCompEvilSequence (including negative-zero divisor edge cases), OnesCompVolumeSequence (1000 random vectors), and OnesCompSignedCornerSequence (all four sign quadrants + both divide-by-zero forms).
* Corner-case sequence (MultiCycleDividerCornerSequence) added for the two's complement divider covering signed boundaries, overflow (MIN_INT ÷ −1), and divide-by-zero.
* A fix to currIndex width (logDataWidth + 1 bits) ensures the loop-exit condition works correctly for all data widths.

## Backwards-compatibility

No

## Documentation

Doc changes included in this PR.

### DIFF
--- a/doc/components/divider.md
+++ b/doc/components/divider.md
@@ -1,10 +1,22 @@
 # Divider
 
-ROHD HCL provides an integer divider module to get the quotient and the remainder of dividend and divisor operands. The divider implementation is not pipelined and has a minimum latency of 3 cycles. The maximum latency is dependent on the width of the operands (upper bound of `O(WIDTH**2)`). Note that latency increases exponentially as the absolute difference between the dividend and the divisor increases (worst case: largest possible dividend and divisor of 1).
+ROHD HCL provides two multi-cycle integer divider modules that share a common interface:
+
+* `MultiCycleDivider` — uses **two's complement** signed arithmetic.
+* `OnesComplementDivider` — uses **one's complement** signed arithmetic.
+
+Both modules compute an integer quotient and, optionally, a remainder. The dividers are not pipelined and have a minimum latency of 3 cycles. The maximum latency depends on the width of the operands and the selected mode:
+
+| Mode | Algorithm | Latency upper bound |
+|------|-----------|---------------------|
+| `computeRemainder: true` (default) | O(n²) greedy | `O(WIDTH²)` |
+| `computeRemainder: false` | O(n) binary long-division | `O(WIDTH)` |
+
+In both modes, latency increases as the absolute difference between dividend and divisor increases (worst case: largest possible dividend, divisor of 1).
 
 ## Interface
 
-The inputs to the divider module are:
+Both modules share `MultiCycleDividerInterface`. The inputs are:
 
 * `clock` => clock for synchronous logic
 * `reset` => reset for synchronous logic (active high, synchronous to `clock`)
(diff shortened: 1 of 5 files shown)


doc/components/divider.md
```diff
@@ -1,10 +1,22 @@
 # Divider
 
-ROHD HCL provides an integer divider module to get the quotient and the remainder of dividend and divisor operands. The divider implementation is not pipelined and has a minimum latency of 3 cycles. The maximum latency is dependent on the width of the operands (upper bound of `O(WIDTH**2)`). Note that latency increases exponentially as the absolute difference between the dividend and the divisor increases (worst case: largest possible dividend and divisor of 1).
+ROHD HCL provides two multi-cycle integer divider modules that share a common interface:
+
+* `MultiCycleDivider` — uses **two's complement** signed arithmetic.
+* `OnesComplementDivider` — uses **one's complement** signed arithmetic.
```
desmonddak, line 6: I wonder about the organization -- both dividers could share a common class that has all the state management, and this could be called `MultiCycleDivider`, and then we could have a `TwosComplementDivider` and a `OnesComplementDivider` extend and use some customization of the logic.  Right now, `OnesComplementDivider` uses the states of the `MultiCycleDivider`, but everything else is different, and it isn't clear why.

One good reason to figure this out is that we may end up with other 1 versus 2s complement arithmetics and we might be able to borrow some of the key differences (say during accumulation you need to do slightly different things).

joserochh: Moved to common base with two extensions. Base class in on a different file while divider.dart keeps Two's Complement Divider, which is the default divider implementation.

PR #293, opened 2026-06-30
Bug fix in loadMemString for unaligned sections

kimmeljo:
## Description & Motivation

`loadMemString` has a bug when there is a discontinuity at a start address that is not aligned with the word size where it is possible for data to get squashed.

## Related Issue(s)

N/A

## Testing

Confirmed in user environment and will add a unit test.

## Backwards-compatibility

Backwards compatible as this is a bug fix.

## Documentation

N/A


mkorbel1 (changes requested):
Looks great once you add a unit test, thanks for catching and fixing!

mkorbel1 (approved):
Looks great to me, thanks for fixing!

PR #301, opened 2026-09-08
docs: add DataPortInterface usage example #192

### LINKED ISSUE #192: [Documentation] Add Usage examples to online github docs
### Motivation

I was trying to understand how to use the DataPortInterface class using https://intel.github.io/rohd-hcl/rohd_hcl/DataPortInterface-class.html and could not figure out how I was supposed to use it.. A simple usage example of the class would go a long way for each of the classes.

### Desired solution

Class documentation includes a simple usage example the user get started using the class. 

### Alternatives considered

Searching the repo for the class and figure out the usage that way. 

### Additional details

_No response_

AzazelSensei:
## Description & Motivation

The generated class page for `DataPortInterface` did not show how to wire it. I added a short dartdoc example that constructs read/write ports, passes them to `RegisterFile`, and drives `en`/`addr`/`data`.

## Related Issue(s)

Fixes #192

## Testing

Snippet matches the simple write path in `test/memory/memory_test.dart`. Docs-only, no behavior change.

## Backwards-compatibility

No.

## Documentation

Yes — class dartdoc on `DataPortInterface`, which is what https://intel.github.io/rohd-hcl/rohd_hcl/DataPortInterface-class.html is built from.

### DIFF
--- a/lib/src/memory/memory.dart
+++ b/lib/src/memory/memory.dart
@@ -45,7 +45,20 @@ class MaskedDataPortInterface extends DataPortInterface {
 /// An interface to a simple memory that only needs enable, address, and data.
 ///
 /// Can be used for either read or write direction by grouping signals using
-/// [DataPortGroup].
+/// [DataPortGroup]. Drive [en], [addr], and [data] on a write port. Drive [en]
+/// and [addr] on a read port; [data] is returned by the memory.
+///
+/// ```dart
+/// final clk = SimpleClockGenerator(10).clk;
+/// final reset = Logic();
+/// final wrPort = DataPortInterface(32, 5);
+/// final rdPort = DataPortInterface(32, 5);
+/// RegisterFile(clk, reset, [wrPort], [rdPort], numEntries: 20);
+///
+/// wrPort.en.put(1);
+/// wrPort.addr.put(3);
+/// wrPort.data.put(0xdeadbeef);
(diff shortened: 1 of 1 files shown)


lib/src/memory/memory.dart
```diff
@@ -45,7 +45,20 @@ class MaskedDataPortInterface extends DataPortInterface {
 /// An interface to a simple memory that only needs enable, address, and data.
 ///
 /// Can be used for either read or write direction by grouping signals using
-/// [DataPortGroup].
+/// [DataPortGroup]. Drive [en], [addr], and [data] on a write port. Drive [en]
+/// and [addr] on a read port; [data] is returned by the memory.
+///
+/// ```dart
```
mkorbel1, line 67: This example is good for a testbench, but not for real hardware.  I think it's worth including (additionally at least) an example of using the interface in a hardware context (not just `.put`s from a testbench).  Some people may be encountering ROHD `Interface`s one of the first times with these memory components.

mkorbel1 (changes requested):
Thank you for contributing!

lib/src/memory/memory.dart
AzazelSensei: Added a hardware example that drives the ports with `<=`, and left the `.put` snippet as a testbench example.

PR #302, opened 2026-09-10
Floating

### LINKED ISSUE #112: Add converters betwee `FloatingPointValue` and `FixedPointValue`

### LINKED ISSUE #113: Make `FloatingPointValue` operations scale beyond `double`

### LINKED ISSUE #115: Maximum Exponent for E4M3
### Describe the bug

The maximum exponent for E4M3 is `bias + 1` to increase the range unlike other floating point types.

### To Reproduce

_No response_

### Expected behavior

_No response_

### Actual behavior

_No response_

### Additional: Dart SDK info

_No response_

### Additional: pubspec.yaml

_No response_

### Additional: Context

_No response_

### LINKED ISSUE #173: Unify rounding modes
### Motivation

It would be nice to have the same options for rounding for both fixed and float operations by using our enum for rounding modes.
Right now we have ofDouble and ofDoubleUnrounded for the floating point type, and just ofDouble for the fixed-point. 

### Desired solution

 Floating and fixed-point should have a parallel API that passes in the type of rounding.

Also, the other rounding modes need implementation.  Right now we have truncation and RNE supported in various places, both on the Logic side as well as the value side.  These should all be combed out so they are parallel and uniform to lead to a least-surprise API.

We are working on a refactoring of the FPV API to remove some of the boilerplate and factory constructors, but this feature could be added after that is complete to help further simplify the API on the value side.

### Alternatives considered

_No response_

### Additional details

_No response_

### LINKED ISSUE #190: RoundRNE is not rounding to even, both logic and value side
### Describe the bug

While the FP components match our conversion from double with rounding, the rounding algorithm is not truly RNE.   RNE needs to consider the mantissa width; it looks like it is working correctly for odd lengths but not even.

### To Reproduce

The bug won't reproduce on 32-bit because the RoundRNE is correct for odd mantissa lengths.  So it needs to be validated by inspection and modifying both the value side and the logic side together.

### Expected behavior

_No response_

### Actual behavior

_No response_

### Additional: Dart SDK info

_No response_

### Additional: pubspec.yaml

```yaml

```

### Additional: Context

_No response_

### LINKED ISSUE #191: Add all rounding modes
### Motivation

We only support truncation and rounding-nearest-even in our logic and value side of  floating-point.
We have enumerated the different rounding modes but not implemented them.  If we have more components like RoundRNE and we factor out the value-side equivalents, we will have a full bevy of rounders available for our FP components.

We need to refactor our components to also use these rounders, which requires setting up the mantissas for the rounders to perform.

### Desired solution

Create value-side rounders and refactor the ofDouble routine to use them.
Create logic-side rounders and refactor the adders and multipliers and converters to use them.
Write tests that, at the very least, compare the value side and logic side implementations using exhaustive tests.

### Alternatives considered

_No response_

### Additional details

_No response_

### LINKED ISSUE #194: FloatingPointMultiplier to narrower FP type
### Motivation

Currently, the fp multiplier can expand to a wider FP type in both exponent and mantissa.
It throws if we try to have it output a narrower type.

### Desired solution

Allow for conversion to both smaller exponents and/or mantissas by 
1) Adding a rounding capability, which can result in a bump in the exponent (and a check for overflow)
2) computing to a narrower exponent may require checking for overflow
3) potentially right-shifting the mantissa in the underflow case

### Alternatives considered

_No response_

### Additional details

_No response_

### LINKED ISSUE #199: Add basic operations directly on arithmetic types

### LINKED ISSUE #200: Add an easy way for arithmetic `Value`s to convert to `Const` versions of corresponding signals
### Motivation

It would be helpful to be able to create a `Value` and then create a `Const` version of it, corresponding to the `Logic` version of it.  For example, creating a `Const` version of `FloatingPoint` based on a value of `FloatingPointValue` type.

### Desired solution

Create some automation that accomplishes creating constant `FloatingPoint` (or similar) from their corresponding value type.

### Alternatives considered

_No response_

### Additional details

_No response_

### LINKED ISSUE #206: FloatingPointValue unit of least precision (ulp) is incorrect for subnormals
### Describe the bug

the ulp function is incomplete and does not properly compute the unit of least precision when the number
is subnormal.

### To Reproduce

_No response_

### Expected behavior

_No response_

### Actual behavior

_No response_

### Additional: Dart SDK info

_No response_

### Additional: pubspec.yaml

```yaml

```

### Additional: Context

_No response_

### LINKED ISSUE #241: FloatingPointConverter fails on narrowing mantissa with explicitJBit
### Describe the bug

Tests left a hole on changing the mantissaWidth when the output is explicitJBit.

### To Reproduce

Run explicitToExplicit exhaustive with narrowing mantissa and a case will fail.

Need better exhaustive tests, including different exponent and mantissa widths, but also with explicitJBit and now DAZ/FTZ to be tested.

### Expected behavior

_No response_

### Actual behavior

_No response_

### Additional: Dart SDK info

_No response_

### Additional: pubspec.yaml

```yaml

```

### Additional: Context

_No response_

### LINKED ISSUE #249: FixedPoint and FixedPointValue have different default values for signed
### Describe the bug

FixedPoint defaults to a signed version whereas FixedPointValue defaults to an unsigned version.

They should be the same, but this is a breaking change as there could be reliance on the default.

### To Reproduce

_No response_

### Expected behavior

_No response_

### Actual behavior

_No response_

### Additional: Dart SDK info

_No response_

### Additional: pubspec.yaml

```yaml

```

### Additional: Context

_No response_

### LINKED ISSUE #252: FP NaN are not comparable (e.g. nan != nan)
### Describe the bug

Floating-point values that are NaN are not comparable.  So all tests should check if computed.isNan equals expected.isNaN  before checking if computed == expected.

### To Reproduce

Create two NaN values and compare them.  They should not be equal even if their bitfields are.

### Expected behavior

_No response_

### Actual behavior

_No response_

### Additional: Dart SDK info

_No response_

### Additional: pubspec.yaml

```yaml

```

### Additional: Context

_No response_

desmonddak:
## Description & Motivation

This PR is focused on resolving issues reported as well as TODOs documented on the arithmetic library portion of ROHD-HCL.  The biggest fixes are around rounding and conversions.

### Features

- Added IEEE rounding modes across floating-point values and logic operations.
  Fixes #191.
- Unified fixed- and floating-point value rounding APIs. Fixes #173.
- Added exact `FloatingPointValue`/`FixedPointValue` conversions. Fixes #112.
- Added `toLogic()` conversion for fixed- and floating-point values. Fixes #200.
- Added typed fixed- and floating-point arithmetic signal operations. Fixes #199.
- Pushed runtime configuration control through multipliers based on PR #222.

### Fixes

- Corrected floating-point multiplier rounding, narrow-output behavior, and
  zero multiplication. Fixes #194.
- Corrected round-nearest-even behavior across mantissa widths and subnormal
  boundaries. Fixes #190.
- Corrected subnormal `FloatingPointConverter` conversions and added exhaustive
  implicit-j-bit subnormal coverage. Fixes #241.
- Corrected IEEE-754 NaN comparisons and encodings. Fixes #252 and Fixes #115.
- Corrected `FloatingPointValue.ulp()` for subnormals. Fixes #206.
- Made `FloatingPointValue.operator /` exact and correctly rounded. Fixes #113.
- Corrected `FloatToFixed` explicit-j-bit conversion, overflow handling, and
  discarded-bit sticky rounding.
- Added early validation for multiply-accumulate output widths and runtime
  boolean control widths.

### API Cleanup and Verification

- Added `StaticOrRuntimeControl<T>` as the general static-or-runtime control
  abstraction; deprecated `BooleanConfig` and `RuntimeConfig`.
- Expanded exhaustive conversion, rounding, arithmetic, and
  multiply-accumulate coverage.

## Testing

All existing tests pass. Several new tests added to cover bugs reported and fixed.

## Backwards-compatibility

- `FixedPointValue`, its factory, and its populator now default to
  `signed: true`, matching `FixedPoint`. Pass `signed: false` to retain the
  previous unsigned behavior. Fixes #249.

## Documentation

Yes.  Relevant documentation was updated to match the new functionality introduced (e.g., `StaticOrRuntimeControl`.


desmonddak:
@copilot resolve the merge conflicts in this pull request

